PR #3, opened 2017-01-29
Fix pam_xauth when using pam_mktemp.

klondi:
For some reason when using su, getgid won't return the calling user's gid but 0 (root).
pam_mktemp will cause the xauthority file to be created on a folder where either uid 0 or the user's gid are required to access it.
As a result of those two conditions, xauth will be run with the user's uid but gid 0 and therefore fail acquiring the magic cookies locking the call to su for 30 seconds.

In this commit we replace all calls to getuid and getgid to accesses to the cached structure from pam_modutil_getpwuid which is used at the beggining.
This will make xauth run using both the user's uid and gid and fix this issue.

As a side effect, having less syscalls may speed up loging slightly (but to a nearly unnoticeable level).

For more information, see https://bugs.gentoo.org/show_bug.cgi?id=585550

### DIFF
--- a/modules/pam_xauth/pam_xauth.c
+++ b/modules/pam_xauth/pam_xauth.c
@@ -478,7 +478,7 @@ pam_sm_open_session (pam_handle_t *pamh, int flags UNUSED,
 	   *                 if .xauth/export does not exist
 	   * import(user=*): if <ruser> is listed in .xauth/import, or
 	   *                 if .xauth/import does not exist */
-	  i = (getuid() != 0 || tpwd->pw_uid == 0) ? PAM_SUCCESS : PAM_PERM_DENIED;
+	  i = (rpwd->pw_uid != 0 || tpwd->pw_uid == 0) ? PAM_SUCCESS : PAM_PERM_DENIED;
 	  i = check_acl(pamh, "export", rpwd->pw_name, user, i, debug);
 	  if (i != PAM_SUCCESS) {
 	    retval = PAM_SESSION_ERR;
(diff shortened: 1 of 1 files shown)


ldv-alt:
I suppose your su is doing roughtly the following:
- setgroups/initgroups
- pam_setcred
- pam_open_session
- setgid
- setuid
If this is the case, then your su has a bug -- setgroups/initgroups should be called right before setgid,
and your proposal is not a fix of pam_xauth but a workaround for the bug in su.

ldv-alt:
It means that su provided by the shadow package is even more buggy than su provided by util-linux.

klondi:
So I have checked the documentation on this aspect:

From `man 3 pam_setcred`
> On a Linux system the user's UID and GID's are credentials too. However, it has been decided that these properties (along with the default supplementary groups of which the user is a member) are credentials that should be set directly by the application and not by PAM. Such credentials should be established, by the application, prior to a call to this function. For example, initgroups(2) (or equivalent) should have been performed.

Also:
> The pam_setcred function is used to establish, maintain and delete the credentials of a user. It should be called to set the credentials after a user has been authenticated and before a session is opened for the user (with pam_open_session(3)).

setgroups should be called before pam_setcred (which is called before pam_open_session),

Also since the getgid returns 0 instead of the user's gid it's also clear that setgid is being called before calling pam_open_session too.

I don't mind writing a patch for shadow's su but the documentation MUST be clear on that regard if I am to do so.

t8m:
Actually the su from shadow does exactly what @ldv-alt suggests to do - i.e. it calls setgid() before initgroups() and the rest of the sequence follows. But that is wrong!

When PAM is involved the right sequence is actually described in the @ldv-alt  comment above as wrong thing:

1. setgroups/initgroups
2. pam_setcred
3. pam_open_session
4. setgid
5. setuid

With this sequence (which is in util-linux login and su, and also in openssh sshd) everything will work correctly and conforms to the manpage text cited above. So yes, the bug is in shadow package su command but it is the other way around.

PR #4, opened 2017-02-09
strtoX() does not set endptr to NULL

jmoellers:
According to the manual page: "If there were no digits at all, strtol() stores the original value of nptr in *endptr (and returns 0)", so it will not be set to NULL.


t8m:
This would not handle the situation when the netmask_ptr points to non-empty but non-numeric string.

jmoellers:
It would be an optimization that would catch "172.16.0.0/".

t8m:
I've merged the commit manually.

PR #5, opened 2017-02-09
Resolve hostnames in access.conf file.

jmoellers:
Create addrinfo list of IP-addresses or hostnames in access.conf and use that to match with incoming IP address.


jmoellers:
Afetr some discussions, this patch needs some serious reworking.

PR #20, opened 2017-04-27
pam_timestamp: Correctly handle Unix98 ptys.

gnikandrov:
pam_timestamp doesn't work correctly on Unix98 pseudoterminals.

For example, if you configure it for su, it works fine on ttyS0 and tty1, but fails to work from remote ssh session or xterm (which use ttys like pts/1).

The proposed pullrequest should fix it.


Daniel-Abrecht:
1) Why doesn't check_tty get an absolute path? (/dev/pts/x instead of pts/x)
2) How can I reproduce this?

gnikandrov:
Sorry, my fault. Seems I haven't done enough testing. On Debian it actually gets the absolute path, only Fedora/RH is affected.

PR #23, opened 2017-06-28
pam_unix: Fix broken nullok handling

peterurbanec:
The ```nullok``` argument to ```pam_unix``` module was not working as expected. This was because the check for empty password erroneously used the ```UNIX__NONULL``` flag instead of ```UNIX__NULLOK``` flag.

Test case from an embedded system:

Configure openssh with the following options:
```
UsePAM yes
PermitRootLogin yes
PermitEmptyPasswords yes
```

Configure PAM for sshd service:
```
auth    [success=1 default=ignore]      pam_unix.so nullok
auth    requisite                       pam_deny.so
auth    required                        pam_permit.so
```

Clear root passwd:
```
passwd -d root
```

Connect with ssh:
```
$ ssh root@devt4
Password:
packet_write_wait: Connection to 192.168.42.34 port 22: Broken pipe
```

Log shows:
```
Jun 28 23:23:56 devt4 authpriv.notice sshd[15852]: pam_unix(sshd:auth): authentication failure; logname= uid=0 euid=0 tty=ssh ruser= rhost=192.168.42.16  user=root
Jun 28 23:23:59 devt4 auth.info sshd[15852]: Postponed keyboard-interactive for root from 192.168.42.16 port 35174 ssh2 [preauth]
Jun 28 23:24:00 devt4 auth.info sshd[15852]: Postponed keyboard-interactive/pam for root from 192.168.42.16 port 35174 ssh2 [preauth]
Jun 28 23:24:00 devt4 auth.info sshd[15852]: Accepted keyboard-interactive/pam for root from 192.168.42.16 port 35174 ssh2
Jun 28 23:24:00 devt4 auth.crit sshd[15852]: fatal: PAM: pam_setcred(): Failure setting user credentials
```

Signed-off-by: Peter Urbanec <git.user@urbanec.net>

### DIFF
--- a/modules/pam_tty_audit/pam_tty_audit.8.xml
+++ b/modules/pam_tty_audit/pam_tty_audit.8.xml
@@ -44,10 +44,10 @@
         </term>
         <listitem>
           <para>
-	    For each user matching one of comma-separated glob
-	    <option><replaceable>patterns</replaceable></option>, disable
-	    TTY auditing.  This overrides any previous <option>enable</option>
-	    option matching the same user name on the command line.
+	    For each user matching <option><replaceable>patterns</replaceable></option>,
+	    disable TTY auditing.  This overrides any previous <option>enable</option>
+	    option matching the same user name on the command line. See NOTES
+	    for further description of <option><replaceable>patterns</replaceable></option>.
           </para>
         </listitem>
       </varlistentry>
@@ -57,10 +57,10 @@
         </term>
         <listitem>
           <para>
-	    For each user matching one of comma-separated glob
-	    <option><replaceable>patterns</replaceable></option>, enable
-	    TTY auditing.  This overrides any previous <option>disable</option>
-	    option matching the same user name on the command line.
+	    For each user matching <option><replaceable>patterns</replaceable></option>,
+	    enable TTY auditing.  This overrides any previous <option>disable</option>
+	    option matching the same user name on the command line. See NOTES
+	    for further description of <option><replaceable>patterns</replaceable></option>.
           </para>
         </listitem>
       </varlistentry>
@@ -138,6 +138,16 @@
     <para>
       To view the data that was logged by the kernel to audit use
       the command <command>aureport --tty</command>.
+    </para>
+    <para>
+      The <option><replaceable>patterns</replaceable></option> are comma separated
+      lists of glob patterns or ranges of uids. A range is specified as
+      <replaceable>min_uid</replaceable>:<replaceable>max_uid</replaceable> where
+      one of these values can be empty. If <replaceable>min_uid</replaceable> is
+      empty only user with the uid <replaceable>max_uid</replaceable> will be
+      matched. If <replaceable>max_uid</replaceable> is empty users with the uid
+      greater than or equal to <replaceable>min_uid</replaceable> will be
+      matched.
     </para>
   </refsect1>
 
--- a/modules/pam_tty_audit/pam_tty_audit.c
+++ b/modules/pam_tty_audit/pam_tty_audit.c
@@ -199,6 +199,57 @@ cleanup_old_status (pam_handle_t *pamh, void *data, int error_status)
   free (data);
 }
 
+enum uid_range { UID_RANGE_NONE, UID_RANGE_MM, UID_RANGE_MIN,
+    UID_RANGE_ONE, UID_RANGE_ERR };
+
+static enum uid_range
+parse_uid_range(pam_handle_t *pamh, const char *s,
+                uid_t *min_uid, uid_t *max_uid)
+{
+    const char *range = s;
+    char *pmax;
+    char *endptr;
+    enum uid_range rv = UID_RANGE_MM;
+
+    if ((pmax=strchr(range, ':')) == NULL)
+        return UID_RANGE_NONE;
+    ++pmax;
+
+    if (range[0] == '@' || range[0] == '%')
+        ++range;
+
+    if (range[0] == ':')
+        rv = UID_RANGE_ONE;
+    else {
+            errno = 0;
+            *min_uid = strtoul (range, &endptr, 10);
+            if (errno != 0 || (range == endptr) || *endptr != ':') {
+                pam_syslog(pamh, LOG_DEBUG,
+                           "wrong min_uid value in '%s'", s);
+                return UID_RANGE_ERR;
+            }
+    }
+
+    if (*pmax == '\0') {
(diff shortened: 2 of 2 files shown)


t8m:
No, the UNIX__NONULL flag is used properly. The nullok option will clear the UNIX__NONULL flag besides setting the UNIX__NULLOK flag. There must be a different cause for the disallowing of empty password for root. What about regular users, are they allowed to log-in with empty password or not?

peterurbanec:
If this is the wrong approach, can you please suggest an alternative?

I added some simple debug to confirm that the two flags are in consistent state. They are not, so the assertion that "_The nullok option will clear the UNIX__NONULL flag besides setting the UNIX__NULLOK flag._" does not hold true.

Here's the simple debug code from line 761 in support.c:
```
D(("UNIX__NULLOK flag is %s", on(UNIX__NULLOK, ctrl) ? "on" : "off"));
D(("UNIX__NONULL flag is %s", on(UNIX__NONULL, ctrl) ? "on" : "off"));
retval = verify_pwd_hash(p, salt, on(UNIX__NULLOK, ctrl));
```
and the matching debug output:
```
[../../../Linux-PAM-1.3.0/modules/pam_unix/support.c:_unix_verify_password(761)] UNIX__NULLOK flag is off
[../../../Linux-PAM-1.3.0/modules/pam_unix/support.c:_unix_verify_password(762)] UNIX__NONULL flag is off
```

t8m:
Is the debugging statement in _set_ctrl() in support.c "D(("DISALLOW_NULL_AUTHTOK"));" printing something? If so, then the sshd is the culprit here, it explicitly disallows null authtok via setting PAM_DISALLOW_NULL_AUTHTOK flag and this flag has priority over the nullok option.
If not then there must be a bug in the _set_ctrl() code that makes the NULLOK and NONULL flags not synced. Please also verify whether the current code works with empty passwords for non-root users or not.

t8m:
Also your debugging output is very strange - it's saying that NONULL is off and NULLOK off as well? Is that with nullok option set for pam_unix? I don't see how that could happen.

peterurbanec:
I see what's happening with the ```UNIX__NULLOK``` flag. It's not a flag that can be set. It's a (one-way) method to clear the ```UNIX__NONULL``` flag.

Further debugging of one of the failure modes for sshd shows that when sshd enables empty passwords, it first tries to authenticate with an empty (zero length string) password. This fails for a user that has a password set. After the failure delay, ssh prompts for a password and tries again with the supplied password. This succeeds, but later I see this failure:
```
[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_setcred(203)] recovering return code from auth call
[../../Linux-PAM-1.3.0/libpam/pam_dispatch.c:_pam_dispatch_aux(114)] module returned: Success
[../../Linux-PAM-1.3.0/libpam/pam_dispatch.c:_pam_dispatch_aux(198)] use_cached_chain=1 action=0 cached_retval=7 retval=0
[../../Linux-PAM-1.3.0/libpam/pam_dispatch.c:_pam_dispatch_aux(106)] passing control to module...
[../../Linux-PAM-1.3.0/libpam/pam_dispatch.c:_pam_dispatch_aux(114)] module returned: Failure setting user credentials
[../../Linux-PAM-1.3.0/libpam/pam_dispatch.c:_pam_dispatch_aux(198)] use_cached_chain=1 action=-4 cached_retval=7 retval=17
[../../Linux-PAM-1.3.0/libpam/pam_auth.c:pam_setcred(70)] pam_setcred exit
```

I'll continue debugging this. It's starting to look like openssh may be doing something unexpected.

I'd welcome any suggestion you may have.

Thanks.

peterurbanec:
I made some progress on this. It appears that the root cause is that the caching of ```unix_setcred_return``` does not work across threads/processes and I suspect this will be due to privsep in openssh.

I added the result of ```pid_t tid = syscall(SYS_gettid)``` to debug output generated by ```_pam_output_debug_info``` so that it was obvious which thread/process is making calls to PAM. The thread id is at the start of the line, inside ```{}```. I also added some other debug info and I can see that the call sequence is something like this:

```
{12568}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_authenticate(108)] called.
{12568}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_authenticate(176)] user=root, password=[]
{12568}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_authenticate(182)] recording return code for next time [7]
{12568}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_authenticate(182)] done. [Authentication failure]
{12574}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_authenticate(108)] called.
{12574}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_authenticate(176)] user=root, password=[1]
{12574}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_authenticate(182)] recording return code for next time [0]
{12574}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_authenticate(182)] done. [Success]
{12574}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_acct.c:pam_sm_acct_mgmt(196)] called.
{12574}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_acct.c:pam_sm_acct_mgmt(202)] user = `root'
{12574}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_acct.c:pam_sm_acct_mgmt(307)] all done
{12568}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_setcred(203)] called.
{12568}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_setcred(209)] recovering return code from auth call
{12568}[../../../Linux-PAM-1.3.0/modules/pam_unix/pam_unix_auth.c:pam_sm_setcred(217)] recovered data indicates that old retval was 7
```

Am I correct in assuming that PAM does not support this cross-process/multi-threaded usage model and this should be considered a bug in openssh?

t8m:
Yes, the cross-proccess communication of pam data definitely won't work.

PR #24, opened 2017-06-30
pam_unix: Check return value of malloc used for setcred data

peterurbanec:
Check the return value of malloc and if it failed print debug info, send
a syslog message and return an error code.

Signed-off-by: Peter Urbanec <git.user@urbanec.net>


peterurbanec:
Pull request updated as requested, AUTH_RETURN no longer checks for ret_data being NULL.

PR #25, opened 2017-06-30
Update es.po: ES typo fix

aolle:
Typo fix.


t8m:
Please use fedora.zanata.org to submit translation updates.

t8m:
I've fixed the typo in Zanata.

PR #27, opened 2017-07-03
avoid overwriting quote

### LINKED ISSUE #14: pam_env: Potential issue with changing of a static variable
**pam_env.c**:74:
```
/* This is a flag used to designate an empty string */
static char quote='Z';
```

but in function **check_var** (row 511):
```
    if (&quote == var->defval) {
      /*
       * This means that the empty string was given for defval value
       * which indicates that a variable should be defined with no value
       */
      *var->defval = '\0';
```
as it can be seen, the last line changes the value of `quote`. This does not seem to have ill effects at the moment, but it is not nice because `quote` becomes = `'\0'`, which has also a special meaning different from `quote` itself (`quote` means the empty quoted string, `'\0'` means the empty unquoted string).

glensc:
i only made the changes according to description, not sure is the outcome valid :)

### DIFF
--- a/modules/pam_env/pam_env.c
+++ b/modules/pam_env/pam_env.c
@@ -378,6 +378,7 @@ _parse_line (const pam_handle_t *pamh, char *buffer, VAR *var)
 
   int length, quoteflg=0;
   char *ptr, **valptr, *tmpptr;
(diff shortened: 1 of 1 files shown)


t8m:
No, this is a wrong solution to the problem, this would break the module even more.

PR #28, opened 2017-07-03
setup travis

glensc:
setup travis for linux-pam project, contributors' PR's could be tested automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,9 @@ language: c
 matrix:
   fast_finish: true
   # list only supported gcc versions: https://gcc.gnu.org/
-  # for package names see:
-  # https://wiki.ubuntu.com/ToolChain#GCC
-  # https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test
   include:
+    # https://wiki.ubuntu.com/ToolChain#GCC
+    # https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test
     - compiler: gcc
       addons:
         apt:
@@ -33,39 +32,25 @@ matrix:
           packages:
             - gcc-7
       env: COMPILER=gcc-7
-    - compiler: clang
-      addons:
-        apt:
-          packages:
-            - clang-3.4
-      env: COMPILER=clang
+    # http://releases.llvm.org/
     - compiler: clang
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
-            - llvm-toolchain-precise-3.5
+            - llvm-toolchain-precise-3.8
           packages:
-            - clang-3.5
-      env: COMPILER=clang-3.5
+            - clang-3.8
+      env: COMPILER=clang-3.8
     - compiler: clang
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
-            - llvm-toolchain-precise-3.6
+            - llvm-toolchain-precise-3.9
           packages:
-            - clang-3.6
-      env: COMPILER=clang-3.6
-    - compiler: clang
-      addons:
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test
-            - llvm-toolchain-precise-3.7
-          packages:
-            - clang-3.7
-      env: COMPILER=clang-3.7
+            - clang-3.9
+      env: COMPILER=clang-3.9
 
 install:
   - sudo apt-get -qq update
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ matrix:
 
 install:
   - sudo apt-get -qq update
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,121 @@
+sudo: required
+dist: precise
```
ldv-alt: Why precise?

.travis.yml
```diff
@@ -0,0 +1,121 @@
...
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-4.9
+      env: COMPILER=gcc-4.9
+    - compiler: gcc
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
```
ldv-alt: The oldest supported gcc version is 5.4, why bother testing older versions?

.travis.yml
```diff
@@ -0,0 +1,121 @@
...
+            - gcc-5
+      env: COMPILER=gcc-5
+    - compiler: gcc
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-6
+      env: COMPILER=gcc-6
```
ldv-alt: gcc-7 is also worth testing

.travis.yml
```diff
@@ -0,0 +1,121 @@
...
+      env: COMPILER=clang-3.6
+    - compiler: clang
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-precise-3.7
+          packages:
+            - clang-3.7
+      env: COMPILER=clang-3.7
```
ldv-alt: i wouldn't bother testing clang versions older than 3.8.

.travis.yml
```diff
@@ -0,0 +1,121 @@
...
+
+install:
+  - sudo apt-get -qq update
+  - sudo apt-get -qq install automake autopoint gettext xsltproc docbook-xml docbook-xsl
```
ldv-alt: why not addons.apt.packages?
why listing already installed packages?

.travis.yml
```diff
@@ -0,0 +1,121 @@
...
+
+script:
+  - $COMPILER --version
+  - autopoint --version
+  - git fetch --tags
+  - git revert -n 58d2da7160a477a5f5654be08bb36cc475eb75d7
```
ldv-alt: this is no-no, gettext 0.18.3 is the latest oldstable
if the build environment does not provide a decent version of gettext, fix the environment

.travis.yml
glensc: there's no requirements, so i just took something that works (based on logrotate project setup)

.travis.yml
glensc: don't know what is the oldest supported gcc, just took something as base to get things going.

.travis.yml
glensc: can't mix matrix in sane way, i.e having per item packages and extending with general requirements. have to duplicate them.

but then again, this is example which can be improved if really needed

.travis.yml
glensc: as i wrote in PR body, i do not know how to fix this. so it's temporary measure to get travis working

> don't know how to pull newer gettext in travis
> so it just reverts 58d2da7 during the build.

glensc:
i can update compiler dependencies, but the rest of the changes can be done post merge and by other people.

i just wanted to help, but now it feels like i'm being attacked that i did things so wrong even there aren't any requirements specified anywhere.

.travis.yml
ldv-alt: precise is ubuntu 12.04, trusty is 14.04; the reason why you needed that gettext hack is that you've chosen the wrong repository

.travis.yml
ldv-alt: it's a good idea to visit https://gcc.gnu.org/ then :)

ldv-alt:
> i can update compiler dependencies, but the rest of the changes can be done post merge and by other people.
Thanks for the idea of setting up travis for linux-pam.
Unfortunately, your proposal lacks the quality required by linux-pam.
We usually don't apply changes that are known to need fixes.

.travis.yml
glensc: also, could you more precise? which packages are already installed packages?

glensc:
@ldv-alt changes made

steven-joruk:
I've made the remaining changes recommended by @ldv-alt. Please can you review and cherry-pick [my commit](https://github.com/steven-joruk/linux-pam/commit/db281b583823e6170c5780341cbf597c03ff8ce4) in to your pull request, @glensc?

Here's the Travis CI build from my branch - https://travis-ci.org/steven-joruk/linux-pam

glensc:
@steven-joruk sure, thanks!

Issue #33, opened 2017-09-12
Feature Request: System should reboot after N failed logins


zem:
I would like to have a feature that automatically reboots my system if a configurable number of auth failures occur in a row. 

I think it might best if that goes to the pam-tally2 module. 

The Idea is that if anyone starts playing around with the login on my mobile device (in case of theft for example) the system does a reboot so that the Disk encryption with a much longer, safer passphrase can kick in.

zem:
It is more than a year now and a patch is available for more than 10 Month. 
Did you have any chance to have a closer look at this?

broukema:
This was apparently merged in 2017, but deprecated in 2020: https://github.com/linux-pam/linux-pam/pull/37#pullrequestreview-438619065 .

This feature would be especially useful for mobile phones like the [PinePhone](https://en.wikipedia.org/wiki/PinePhone) running GNU/Linux systems such as [Mobian](https://mobian.org).

PR #37, opened 2017-11-18
pam_tally2: cmd_onerr=/sbin/halt

zem:
This one should hopefully solve feature request #33 please have a closer look because I had some difficulties compiling linux-pam from upstream. 

Once it is installed it  executes cmd_onerr as soon as an account is blocked which could be a simple shutdown so that the luks passphrase can kick in to protect your data.

### DIFF
--- a/modules/pam_tally2/pam_tally2.8.xml
+++ b/modules/pam_tally2/pam_tally2.8.xml
@@ -43,6 +43,9 @@
         root_unlock_time=<replaceable>n</replaceable>
       </arg>
       <arg choice="opt">
+        cmd_onerr=<replaceable>/path/to/halt</replaceable>
+      </arg>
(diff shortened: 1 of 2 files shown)


zem:
I also added the possibility to access the tallylog as a regular user. If you enable the user_access parameter, pam_tally expects /var/log/tallylog to be a directory, where it can place an user owned logfile to store its data.  

With this Patch it is possible to use tally from non suid screen lockers. 

Example: 

`auth     required       pam_tally2.so deny=4 user_access even_deny_root cmd_onerr=/sbin/poweroff onerr=fail unlock_time=1200`

ldv-alt:
pam_tally2 module was removed by commit 709e37b7e131d35b0ec30d31f858bc6917dd2b2e.

PR #41, opened 2018-02-15
Fix strftime options for date of last login for Catalan

rbuj:


ldv-alt:
The translations are maintained elsewhere, please use https://translate.fedoraproject.org/projects/linux-pam/master/ca/ instead.

PR #48, opened 2018-05-04
pam_motd: add support for a motd.d directory

### LINKED ISSUE #47: Support /etc/motd.d (cf. util-linux /etc/issue.d)
The cockpit project would like to add a brief bit of text to the motd to advertise the availability of the admin interface at a given URL.

We're able to do that for console (read:getty) logins via /etc/issue.d/ which is a recently added feature of util-linux, but the main beneficiary of such a message would be a user who is logging in via ssh.  We don't want to do an ssh banner, as that would be much too obtrusive, but since most interactive sessions already feature a motd, this seems like a good place to include it.

We also don't want to update the distribution's /etc/motd file.

Our approach to doing this for issue.d is that we have systemd tmpfiles create a "cockpit isn't running" text and write it to /run/cockpit/issue.  We change that text according to if the cockpit listener socket is started or stopped.  We symlink that into /etc/issue.d/.

If I were to design the feature, I'd:
 - add a new motd_directory= argument that would allow displaying other directories.
 - change the default arguments to motd=/etc/motd motd_directory=/etc/motd.d
 - if either of motd= or motd_directory= is given, don't apply the defaults

I'm willing to do a patch, if you like.

Thoughts?

allisonkarlitskaya:
Add a new feature to pam_motd to allow packages to install their own
message files in a "motd.d" directory, to be displayed after the primary
motd.

Add an option motd_d= to specify the location of this directory.

Modify the defaults, in the case where no options are given, to display
both /etc/motd and /etc/motd.d.

Fixes #47

 * modules/pam_motd/pam_motd.c: add support for motd.d
 * modules/pam_motd/pam_motd.8.xml: update the manpage

### DIFF
--- a/modules/pam_motd/pam_motd.c
+++ b/modules/pam_motd/pam_motd.c
@@ -17,6 +17,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <fcntl.h>
+#include <dirent.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <pwd.h>
@@ -33,6 +34,7 @@
 
 #define PAM_SM_SESSION
 #define DEFAULT_MOTD	"/etc/motd"
+#define DEFAULT_MOTD_D	"/etc/motd.d"
 
 #include <security/pam_modules.h>
 #include <security/pam_modutil.h>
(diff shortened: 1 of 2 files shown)


modules/pam_motd/pam_motd.c
```diff
@@ -72,41 +120,39 @@ int pam_sm_open_session(pam_handle_t *pamh, int flags,
 			   "motd= specification missing argument - ignored");
 	    }
 	}
+	else if (!strncmp(*argv,"motd_d=",5)) {
```
ldv-alt: Try to specify "motd_///etc" as an argument to this module. :)

allisonkarlitskaya:
eep.  copy/paste error, obviously.  I'll push again once I find the next place with wifi :)

allisonkarlitskaya:
I've pushed the fixed branch.  Thanks.

allisonkarlitskaya:
Can I please get another review?

t8m:
Can you please rename the new option to motd_dir ?

t8m:
Merged, thank you for the patch.

allisonkarlitskaya:
Thanks very much!

PR #52, opened 2018-06-16
Use utmpx where available

awilfox:
On the musl libc, utmp is unimplemented.  The [utmps library by skarnet](http://skarnet.org/software/utmps/) provides utmpx functionality on musl, but not utmp.  This pull request modifies Linux PAM to support user accounting on musl distros that link it to utmps, by using `<utmpx.h>` instead of `<utmp.h>` if `HAVE_UTMPX_H` is defined in `config.h`.

On glibc, utmpx and utmp are aliases, so there is no functional change there.

### DIFF
--- a/libpam/pam_modutil_getlogin.c
+++ b/libpam/pam_modutil_getlogin.c
@@ -10,7 +10,11 @@
 
 #include <stdlib.h>
 #include <unistd.h>
-#include <utmp.h>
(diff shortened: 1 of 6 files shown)


t8m:
I am concerned about making the utmpx calls as the default. They should be done only for the MUSL case.

Issue #31, opened 2017-07-19
pam_env: Variables read from environment file are truncated to 1023 characters


a12e:
If you define a long environment variable in `/etc/environment` it gets truncated to 1023 characters.

Example `/etc/environment`:
```
PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"
HTTP_PROXY=http://some.proxy:3128/
HTTPS_PROXY=http://some.proxy:3128/
[...]

PR #55, opened 2018-06-24
pam_env: ignore / skip lines beginning with '='

steved:
pam_putenv in libpam/pam_env.c makes a special exception for an '=' as
the first character, aborting with PAM_BAD_ITEM. When combined with the
truncation to 1023 characters, if the 1024th character is an '=', this
can cause unexpected system lockout. Note, this would also happen if the
line started with an '='.

Related to issue #31.


steved:
Appears to have been fixed in 563d21d6dbb6d64613919ccb1cc939bae546baab. Thanks!

PR #61, opened 2018-08-07
Add ability to specify alt shells and passwd files

hawartens:
This patch adds the ability to specify an alternate password file (pwfile=/path/to/pwfile; default /etc/passwd) and an alternate shells file (shellfile=/path/to/shellfile; default /etc/shells).

### DIFF
--- a/modules/pam_shells/pam_shells.c
+++ b/modules/pam_shells/pam_shells.c
@@ -1,7 +1,5 @@
 /* pam_shells module */
 
-#define SHELL_FILE "/etc/shells"
-
 /*
  * by Erik Troan <ewt@redhat.com>, Red Hat Software.
  * August 5, 1996.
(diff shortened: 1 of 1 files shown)


t8m:
I am sorry for ignoring this PR for so long. There are some issues with the patch - in particular could you please describe the use-case?
Also the call that replaces the querying of passwd file with reading another one from arbitrary path is problematic as the original code handled properly users stored in remote identity servers. The replacement will look only into the local /etc/passwd.

modules/pam_shells/pam_shells.c
```diff
@@ -45,83 +169,141 @@ static int perform_check(pam_handle_t *pamh)
...
+        {
+            pam_syslog(pamh, LOG_ERR, "Username is not set");
+            return PAM_SERVICE_ERR;
+        }
     }
 
-    pw = pam_modutil_getpwnam(pamh, userName);
-    if (!pw) {
-	return PAM_AUTH_ERR;		/* user doesn't exist */
+    pw = pam_shells_getpwnam(pamh, opt, userName);
```
t8m: This should not be called for the default case.

PR #62, opened 2018-08-19
Add group support to pam_namespace

UntoSten:
Add pam_limits style @groupname support to pam_namespace.


t8m:
I am sorry for taking so long time to look at the PR.
My first comment is that a documentation of the enable_groups option is missing.

Issue #32, opened 2017-08-18
Remove dependency on glibc extensions


PureTryOut:
Right now, it's impossible to build pam on musl-based systems, because of several problems. I'll try to list all I can find. Please somehow workaround these to make it buildable on musl.

## 1. undefined `strndupa`

Due to calls to the `strndupa` function. [As the manpage](http://man7.org/linux/man-pages/man3/strdup.3.html) shows, this is a GNU extension, and thus not available in other libc's. This means it can not be packaged for (as example) Alpine Linux.

Lines where this function is being used:

`modules/pam_exec/pam_exec.c:183:`
```
if (resp)
{
    pam_set_item (pamh, PAM_AUTHTOK, resp);
    authtok = strndupa (resp, PAM_MAX_RESP_SIZE);
    _pam_drop (resp);
}
```

`modules/pam_exec/pam_exec.:188:`
```
else
    authtok = strndupa (void_pass, PAM_MAX_RESP_SIZE);
```

These can potentially be fixed by replacing these calls with `strndup`, but this will require manual freeing memory.

## 2. undefined `_PATH_LASTLOG`
`modules/pam_lastlog/pam_lastlog.c` at line 30 doesn't check for general non-glibc systems and thus doesn't detect musl, thus resulting in a "_PATH_LASTLOG" error.

This is easily solved by appending `!defined(__GLIBC__)` to the end of line 30.

## 3. undefined reference to `logwtmp`
This actually seems to be lacking from musl. This function is nowhere to be found in the source, but [there is a patch available](http://www.openwall.com/lists/musl/2013/12/04/11). Either we have to patch musl, or a workaround in pam could be used (patching musl is really not an option, as it only follows the POSIX standard which `logwtmp` is not a part of).

ffontaine:
Issue can be closed thanks to https://github.com/linux-pam/linux-pam/commit/73bf6d25ddb7a2cb73bceda3d880174b1d1e4a26 and https://github.com/linux-pam/linux-pam/commit/91d4678388b2a7d768ee2ec8cc569e11fc223ffd

PR #66, opened 2018-09-21
Use strndupa only for glibc in pam_exec.c

DanySpin97:
strndupa is a glibc only function, replace it with a strcpy call.

See #32 

Based on [this patch](https://github.com/void-linux/void-packages/blob/2a51ba586a4dc0e01011fb4b6914f8edfef8e31b/srcpkgs/pam/patches/musl-fix-pam_exec.patch).


marv:
This was fixed by https://github.com/linux-pam/linux-pam/issues/32#issuecomment-578322156 and can be closed

t8m:
Closing. Thank you for the notice.

PR #69, opened 2018-10-16
pam_motd: Add support for specifying multiple motd directories

### LINKED ISSUE #68: pam_motd: Support multiple directory paths to motd files, with filename overrides
# pam_motd: Support multiple directory paths to motd files, with filename overrides

As part of Fedora CoreOS, we would like to show a `motd` snippet that has been generated at runtime upon SSH login, in order to show updated system information (https://github.com/coreos/fedora-coreos-tracker/issues/36).

This is similar to what the cockpit project currently does by having a systemd service unit generate and drop a file under `/run/cockpit/` (#47).

A current way to have PAM show a generated `motd` snippet is to symlink to a file from the user config directory `/etc/motd.d/`. Say `hello-pkg` is a package that displays a generated `motd` snippet as part of its functionality, `hello.motd`. The symlink may look like this:

```
/etc/motd.d/hello.motd -> /run/hello-pkg/motd.d/hello.motd
```

Where one runs into issues with this is needing SELinux configuration to be updated to allow PAM to read a file existing in `/run/hello-pkg/`. To have the change reflected in Fedora upstream, the `selinux-policy` package needs to be updated with this config.

For most use cases, one should be able to drop or install a generated `motd` snippet in a general (non-package-specific) directory and have it display, without needing to change the SELinux config. One should also not need to install symlinks or create package-specific directories to house `motd` snippets associated with the package.

**design**

A design of the feature we are discussing is:
- Have `pam_motd` by default search the following directories, highest priority first:

    ```
    /etc/motd.d/
    /run/motd.d/
    /usr/lib/motd.d/
    ```

    A default argument looks like `motd_dir=/etc/motd.d/:/run/motd.d/:/usr/lib/motd.d/`.

    The snippets existing in the above directories are sourced in lexicographic order by filename.

    Files in `/etc/motd.d/` override files of the same name in `/run/motd.d/` and `/usr/lib/motd.d/`. Files in `/run/motd.d/` override files of the same name in `/usr/lib/motd.d/`. This is similar to how systemd handles `tmpfiles.d` https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html.

    Users may also override snippets by creating a symlink to `/dev/null` in a higher priority directory with the same filename.

- Passing a `motd_dir` argument to `pam_motd` with alternative directories to follow will override the default behaviour (e.g. `session optional pam_motd.so motd_dir=/etc/motd.d/:/var/motd.d/:/usr/lib/motd.d/`) would search in `/etc/motd.d/` first, then `/var/motd.d/`, then `/usr/lib/motd.d/`.
- The same ideas apply for `/etc/motd`, `/run/motd`, `/usr/lib/motd`.

Upstream changes involved:
- Modify the `selinux-policy` to allow PAM to read `motd` snippets in the default directories `/run/motd.d/` and `/usr/lib/motd.d/`. Packages can then install static `motd`-related to files to `/usr/lib/motd.d/`, and drop generated files to `/run/motd.d/`.

**additional information**

We currently have a [Rust prototype](https://github.com/lucab/rpam_motd) which implements the default behaviour, with a [demo](https://asciinema.org/a/204664).

The change proposed in this issue would modify the `pam_motd` module of this repository to implement the change upstream.

I can work on this.

How does this sound? Appreciate feedback and suggestions. Thanks!

rfairley:
This adds to `pam_motd.c` functions to parse a string of colon-separated directories as arguments for `motd` and `motd_dir`, and check for overriding filenames before displaying the motds. This also displays the motds in lexicographic order of the filenames, after merging the files contained in each directory.

Tests are added to verify the previous single-directory functionality works, as well as with multiple directories, and with symlinks.

If this is suitable to have in PAM, I will also update the documentation in this PR.

Appreciate your review and feedback!

Closes #68

### DIFF
--- a/modules/pam_motd/pam_motd.c
+++ b/modules/pam_motd/pam_motd.c
@@ -33,8 +33,8 @@
  */
 
 #define PAM_SM_SESSION
-#define DEFAULT_MOTD	"/etc/motd"
-#define DEFAULT_MOTD_D	"/etc/motd.d"
+#define DEFAULT_MOTD	"/etc/motd:/run/motd:/usr/lib/motd"
+#define DEFAULT_MOTD_D	"/etc/motd.d:/run/motd.d:/usr/lib/motd.d"
 
 #include <security/pam_modules.h>
 #include <security/pam_modutil.h>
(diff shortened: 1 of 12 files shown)


rfairley:
Pushed another commit to fix the logic behind displaying the files in lexicographic order - was not merging filenames from directories before sorting before. Will squash the commit afterwards.

modules/pam_motd/pam_motd.c
```diff
@@ -97,9 +98,180 @@ static void try_to_display_directory(pam_handle_t *pamh, const char *dirname)
...
+    namelist_final = (char **)malloc(sizeof(char *) * total_size);
+    for (int i = 0; i < total_size; i++) {
+	namelist_final[i] = NULL;
+    }
+
+    int num_files = 0;
+    for (int i = 0; i < num_motd_dirs; i++) {
+	for (int j = 0; j < namelist_sizes[i]; j++) {
+	   namelist_final[num_files] = namelist[i][j]->d_name;
+	num_files++;
```
lucab: Misindented line.

modules/pam_motd/pam_motd.c
```diff
@@ -97,9 +98,180 @@ static void try_to_display_directory(pam_handle_t *pamh, const char *dirname)
...
+    if (num_strs < 1) {
+	return NULL;
+    }
+
+    char **arg_split = NULL;
+    char *arg_extracted = NULL;
+    int i = 0;
+
+    arg_split = (char **)malloc(sizeof(char *) * num_strs);
+    if (!arg_split) {
```
lucab: Here and everywhere else, I think it would be better to have null-pointer checks being explicit (i.e. `!= NULL`).

modules/pam_motd/pam_motd.c
```diff
@@ -97,9 +98,180 @@ static void try_to_display_directory(pam_handle_t *pamh, const char *dirname)
...
+	return NULL;
+    }
+
+    arg_extracted = (char *)strtok(arg, delim);
```
lucab: I think the returned pointer cast is not needed here, right?

xtests/tst-pam_motd3.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Verify motd is still displayed when not overridden
+echo "motd: test-show in run - show" > ${TST_DIR}/run/motd.d/test-show.motd
+
+# Test hidden by a symlink to another file (which should not be shown)
+echo "motd: hidden-by-symlink in usr/lib - not show" > ${TST_DIR}/usr/lib/motd.d/hidden-by-symlink.motd
+echo "motd: test-from-symlink - not show" > ${TST_DIR}/test-from-symlink.motd
+ln -s ${TST_DIR}/test-from-symlink.motd ${TST_DIR}/run/motd.d/hidden-by-symlink.motd
+
+# Test hidden by a null symlink
```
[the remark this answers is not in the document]
rfairley: Right, previously this was done implicitly, but by considering *any* symlink to override a file and following the symlink. It probably is better to check the symlink target for `/dev/null` and not try to print it - working this in

modules/pam_motd/pam_motd.c
```diff
@@ -97,9 +98,180 @@ static void try_to_display_directory(pam_handle_t *pamh, const char *dirname)
...
+    while (arg_extracted && i < num_strs) {
+	arg_split[i++] = arg_extracted;
+	arg_extracted = (char *)strtok(NULL, delim);
```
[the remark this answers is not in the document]
rfairley: :+1: addressed this

modules/pam_motd/pam_motd.c
rfairley: Addressed

modules/pam_motd/pam_motd.c
rfairley: Agreed, making them explicit

modules/pam_motd/pam_motd.c
rfairley: :+1: thanks for catching, removing the cast

rfairley:
Pushed changes from review comments aside from symlink checking `/dev/null` - tst-pam_motd3.sh fails as the symlink logic needs to be fixed. Working on this

xtests/tst-pam_motd3.sh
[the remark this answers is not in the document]
rfairley: Actually - leaving this done implicitly - think it is fine. This is to avoid an extra check like `readlink(2)` being necessary before displaying the motd files

rfairley:
Pushed changes to fix the problems with symlinks. One problem was forgetting to `strcmp` when checking for duplicates in `namelist_final`, and the other in the test where the symlink should be relative within the testing directory.

modules/pam_motd/pam_motd.c
```diff
@@ -97,9 +97,192 @@ static void try_to_display_directory(pam_handle_t *pamh, const char *dirname)
...
+static void try_to_display_directories_with_overrides(pam_handle_t *pamh, char **motd_dir_path_split, int num_motd_dirs)
+{
+    if (pamh == NULL || motd_dir_path_split == NULL) {
+	return;
+    }
+    if (num_motd_dirs < 1) {
+	return;
+    }
+
+    struct dirent ***namelist = NULL;
```
t8m: Call namelist better - something like dirlists and dirlists_sizes could be more clear.

modules/pam_motd/pam_motd.c
```diff
@@ -97,9 +97,192 @@ static void try_to_display_directory(pam_handle_t *pamh, const char *dirname)
...
+		    break;
+		}
+	    }
+	    _pam_drop(abs_path);
+	}
+    }
+
+    _pam_drop(namelist_final);
+    _pam_drop(namelist);
+}
```
t8m: You're leaking here members of namelist array, the individual dirent entries, and also the namelist_sizes array.

xtests/tst-pam_motd2.c
```diff
@@ -0,0 +1,65 @@
+/*
```
t8m: To avoid code duplication, please just have a single tst-pam_motd.c which would take the service name as parameter.

xtests/tst-pam_motd3.sh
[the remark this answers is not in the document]
t8m: Having it done implicitly is OK.

dustymabe:
wow. Thanks so much @t8m for the review! It sounds like most of your comments are for code style or implementation detail related. Does this mean if we resolve the minor issues then you'd be +1 for accepting this change ?

t8m:
Yes, functionality-wise I think this is OK.

modules/pam_motd/pam_motd.c
rfairley: Sorry I missed that - addressed now.

modules/pam_motd/pam_motd.c
rfairley: Addressed - sorry `namelist` was hastily chosen from reading the `dirscan` manpage example, should have thought a bit more about it. Changed to `dirscans` and `dirscans_sizes`, holding the output of a `dirscan`. How does this sound? `dirlists` is also good with me

rfairley:
Alright, addressed the changes, added the documentation, and think it is ready for another look! Whenever you have a moment of course. Thanks!

xtests/tst-pam_motd2.c
rfairley: I did this and created a test file `tst-pam_motd.sh`, which then calls the individual `tst-pam_motd{1,2,3}.sh` which each specify the service name. Is this alright to do from the perspective of the testing structure? I can have all the tests done in the one `tst-pam_motd.sh` file as well

t8m: Do as you prefer.

modules/pam_motd/pam_motd.c
```diff
@@ -323,43 +362,43 @@ int pam_sm_open_session(pam_handle_t *pamh, int flags,
...
+	if (pam_split_string(pamh, motd_dir_path_copy, ':',
+		&motd_dir_path_split, &num_motd_dir_paths) == 0) {
+	    goto out;
+	}
     }
 
-
     if (motd_path_split != NULL) {
-	for (int i = 0; i < num_motd_paths; i++) {
+	int i = 0;
```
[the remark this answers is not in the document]
rfairley: Addressed - initialization removed from here and done only at the for loop.

modules/pam_motd/pam_motd.c
```diff
@@ -97,12 +97,234 @@ static void try_to_display_directory(pam_handle_t *pamh, const char *dirname)
     }
 }
 
+/*
+ * Split a DELIM-separated string ARG into an array.
+ * Outputs a newly allocated array of strings OUT_ARG_SPLIT
+ * and the number of strings OUT_NUM_STRS.
+ * Returns 0 in case of error, 1 in case of success.
+ */
+int pam_split_string(const pam_handle_t *pamh, char *arg, char delim,
```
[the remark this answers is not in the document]
rfairley: Addressed

rfairley:
Addressed the two changes, with the patch in response to those changes here https://github.com/rfairley/linux-pam/commit/042eb63910a4abc3bdf0e0e29006225c54d8710b. Rebased on current master.

Good to merge from my end. Thanks!

t8m:
There is no release scheduled yet so yes, it would be fine to open PR against Fedora package.

asavah:
Was this runtime tested?
I'm suspicious this is causing segfault in libpam.so when logging in via sshd making sshd unusable.
I have a pet OS which uses latest and greatest (when possible) packages.
```
Nov 21 04:14:26 e45m1 systemd-coredump[386]: Process 383 (sshd) of user 0 dumped core.

                                             Stack trace of thread 383:
                                             #0  0x00007f54c112b106 __GI___strlen_sse2 (libc.so.6)
                                             #1  0x00007f54c111b15e __GI___strdup (libc.so.6)
                                             #2  0x00007f54c0ff55ba pam_sm_open_session (pam_motd.so)
                                             #3  0x00007f54c15af2e5 n/a (libpam.so.0)
                                             #4  0x000000000040a35b n/a (sshd)
                                             #5  0x00007f54c10bb02e __libc_start_main (libc.so.6)
                                             #6  0x000000000040a50a n/a (sshd)
```
linux-pam built  at https://github.com/linux-pam/linux-pam/commit/f7abb8c1ef3aa31e6c2564a8aaf69683a77c2016 behaves just fine.
x86_64, gcc 8.2, glibc 2.28 (git 2.28 branch)

rfairley:
Sorry this broke on your end. Thank you for reporting it. Looks like this is caused by a case where the line specifying the module in `/etc/pam.d/sshd` looks as follows:

```
session optional    pam_motd.so    motd=...
```

(with no `motd_dir` argument given), which makes the `motd_dir_path` pointer equal to `NULL` which is then passed into `strdup`, causing the segfault.

This is something I should have caught while writing, or I should have had a test with these conditions. The tests I added did not cover this. Pushed a patch which fixes this and adds a test case ( #76).

@t8m I opened a new PR for the fix, but can squash the fix commit into this PR if preferred.

dustymabe:
@asavah thanks for reporting that. Does the fix in #76 (already merged) work for you?

asavah:
@dustymabe Yes, thanks, ssh login works now.

@rfairley No worries, bugs happen ) , good thing that it's fixed now, thanks.
Just FYI this is my system-login which triggered the the crash:
```
cat /etc/pam.d/system-login
#%PAM-1.0

auth       required   pam_tally.so         onerr=succeed file=/var/log/faillog
auth       required   pam_shells.so
auth       requisite  pam_nologin.so
auth       include    system-auth

account    required   pam_access.so
account    required   pam_nologin.so
account    include    system-auth

password   include    system-auth

session    optional   pam_loginuid.so
session    include    system-auth
session    optional   pam_motd.so          motd=/etc/motd
session    optional   pam_mail.so          dir=/var/spool/mail standard quiet
-session   optional   pam_systemd.so
session    required   pam_env.so
```
/etc/motd is a file with one line in it. It works fine now.

PR #72, opened 2018-11-15
pam_unix: Use bcrypt b-variant for computing new hashes.

besser82:
Bcrypt hashes used the "$2a$" prefix since 1997.
However, in 2011 an implementation bug was discovered in bcrypt affecting the handling of characters in passphrases with the 8th bit set.

Besides fixing the bug, OpenBSD 5.5 introduced the "$2b$" prefix for a behavior that exactly matches crypt_blowfish's "$2y$", and the crypt_blowfish implementation supports it as well since v1.1.

That said new computed bcrypt hashes should use the "$2b$" prefix.

* modules/pam_unix/passverify.c: Use bcrypt b-variant.


t8m:
Merging.

PR #76, opened 2018-11-21
pam_motd: Fix segmentation fault when no motd_dir specified

rfairley:
This fixes a regression introduced by #69, where motd_path was set
to NULL and passed into strdup() if the motd_dir argument was
not specified in the configuration file. This caused a segmentation
fault.

* modules/pam_motd/pam_motd.c: fix checks for NULL in arguments
* xtests/Makefile.am: add test scripts and config file
* xtests/tst-pam_motd.sh: add running tst-pam_motd4.sh
* xtests/tst-pam_motd4.pamd: create
* xtests/tst-pam_motd4.sh: create


t8m:
LGTM

PR #78, opened 2018-11-23
pam_unix: Add crypt_default method, if supported.

besser82:
libxcrypt since v4.4.0 supports a default method for its gensalt function on most system configurations.  As the default method is to be considered the strongest available hash method, it should be preferred over all other hash methods supported by pam.

* modules/pam_unix/pam_unix.8.xml: Documentation for crypt_default.
* modules/pam_unix/passverify.c: Add crypt_default method.
* modules/pam_unix/support.h: Likewise.

***

@t8m  After this one, I'll open a PR for the no-MD5-fallback patch.

### DIFF
--- a/modules/pam_unix/pam_unix.8.xml
+++ b/modules/pam_unix/pam_unix.8.xml
@@ -363,6 +363,23 @@
       </varlistentry>
       <varlistentry>
         <term>
+          <option>crypt_default</option>
+        </term>
+        <listitem>
+          <para>
+            When a user changes their password next,
+            encrypt it with the default algorithm and the default
+            amount of rounds provided by the system configuration
+            of libcrypt. If this default algorithm is not known to
+            the <citerefentry>
+	    <refentrytitle>crypt</refentrytitle><manvolnum>3</manvolnum>
+            </citerefentry> function,
+            fall back to MD5.
+          </para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>
           <option>rounds=<replaceable>n</replaceable></option>
         </term>
         <listitem>
@@ -371,6 +388,9 @@
             blowfish, gost-yescrypt, and yescrypt password hashing
             algorithms to
             <replaceable>n</replaceable>.
+            This option will be ignored when the crypt_default option
+            is used, as the default algorithm always uses the value
+            from the system configuration of libcrypt.
           </para>
         </listitem>
       </varlistentry>
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -395,6 +395,14 @@ PAMH_ARG_DECL(char * create_password_hash,
 	if (on(UNIX_MD5_PASS, ctrl)) {
 		/* algoid = "$1" */
 		return crypt_md5_wrapper(password);
+#if (defined(CRYPT_PREFERRED_METHOD_AVAILABLE) && CRYPT_PREFERRED_METHOD_AVAILABLE)
+	} else if (on(UNIX_CRYPT_DEFAULT_PASS, ctrl)
+		  && crypt_preferred_method() != NULL) {
+		algoid = crypt_preferred_method();
+		rounds = 0; /* always use the system preset */
+#endif
+	} else if (on(UNIX_CRYPT_DEFAULT_PASS, ctrl)) {
+		algoid = "*0"; /* never ever a valid method */
 	} else if (on(UNIX_YESCRYPT_PASS, ctrl)) {
 		algoid = "$y$";
 	} else if (on(UNIX_GOST_YESCRYPT_PASS, ctrl)) {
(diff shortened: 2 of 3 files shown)


besser82:
@t8m This should be ready for getting merged.  The implemented changes are trivial and pretty straight forward.

modules/pam_unix/pam_unix.8.xml
```diff
@@ -371,6 +388,9 @@
             blowfish, gost-yescrypt, and yescrypt password hashing
             algorithms to
             <replaceable>n</replaceable>.
+            Will be ignored for the crypt_default algorithm, which
```
t8m: The value will be...

modules/pam_unix/passverify.c
```diff
@@ -395,6 +395,14 @@ PAMH_ARG_DECL(char * create_password_hash,
 	if (on(UNIX_MD5_PASS, ctrl)) {
 		/* algoid = "$1" */
 		return crypt_md5_wrapper(password);
+#if (defined(CRYPT_PREFERRED_METHOD_AVAILABLE) && CRYPT_PREFERRED_METHOD_AVAILABLE)
+	} else if (on(UNIX_CRYPT_DEFAULT_PASS, ctrl)
+		  && crypt_preferred_method() != NULL) {
```
t8m: In which case crypt_preferred_method could return NULL? I am not against checking it here but hopefully this is an "impossible condition" in libxcrypt.

modules/pam_unix/pam_unix.8.xml
besser82: The actual value may vary depending on the actual default method used by libxcrypt:

* sha512crypt: 5k rounds, real iterations of the underlying algorithm
* bcrypt: 5 rounds, real iterations of the underlying algorithm
* yescrypt: 5 "rounds" (16 MiBytes of RAM in 4096 blocks, 4 KiBytes each): yescrypt doesn't have a single parameter for iterations, but uses several tunable parameters for amount of RAM, CPU cost of computation, etc.  For the sake of interface compatibility with the gesalt family of functions, we "translate" the rounds parameter to a bouquet of internal parameters.  "rounds" < 3 use less than 2 MiBytes of RAM, but need more CPU time and cause more random reads of small (1 KiByte) memory blocks, thus more L{1,2}-cache misses; "rounds" > 3 use between 4 and 256 MiBytes of memory, but need less CPU time and fewer random reads of (4 KiBytes) memory blocks, thus likely more L{1,2} cache hits.  That way yescrypt compensates the time-memory trade-off (TMTO) most other hashing methods are suffering from.  The amount of real iterations is pseudo-randomized by the entropy of the passphrase and used salt.

modules/pam_unix/passverify.c
besser82: > In which case crypt_preferred_method could return NULL? I am not against checking it here but hopefully this is an "impossible condition" in libxcrypt.

libxcrypt needs one of those hashing methods enabled to _NOT_ return `NULL`:

* yescrypt ($y$, strongest)
* bcrypt ($2b$, still considered fairly strong), or
* sha512crypt ($6$, any Linux system *must* support this for compatibility with glibc).


In theory it can return `NULL`, when there is no strong enough hashing method selected during build-time configuration (or in near future by runtime preset), but that would lead to a generally br0k3n Linux system in any case.  For the almost impossible likeliness (or some other libcrypt is used by the system) reason, I've introduced a guaranteed to be invalid `algoid` right below.

modules/pam_unix/pam_unix.8.xml
t8m: Yes, that's fine. I just wanted to adjust the text: "This option will be ignored when the crypt_default option is used as the default algorithm always uses the value from the system configuration of libcrypt."

modules/pam_unix/passverify.c
besser82: Well, libxcrypt supports all (except for `crypt16`) of the [crypt() methods](https://en.wikipedia.org/wiki/Crypt_(C)#Support_in_operating_systems) ever used on any system and adds scrypt and (gost-)yescrypt on top of them.

If `glibc` (all distributions that already ship libxcrypt do that) is selected as available hashing group during build-time config, `sha256crypt` will be supported as well, but will not be a preferred method; if just the `glibc` group is selected, `sha512crypt` is preferred.

modules/pam_unix/pam_unix.8.xml
besser82: Fixed.

besser82:
Should be fine now.

t8m:
Merged, thank you.

t8m:
Hmm, actually this breaks the build. I am going to revert.

t8m:
Please open a new PR. Also after more thinking I believe that the missing support for the crypt_default during the build time needs to be handled in a better way. The manual page documentation should explicitly state that this requires a support in libcrypt and the missing support should be logged already in set_ctrl() function.

PR #79, opened 2018-11-23
pam_unix: Return NULL instead of calling crypt_md5_wrapper().

besser82:
If the call to the crypt(3) function failed for some reason during hashing a new login passphrase, the wrapper function for computing a hash with the md5crypt method was called internally by the pam_unix module in previous versions of linux-pam.

With CVE-2012-3287 in mind, the md5crypt method is not considered to be a safe nor recommended hashing method for a new login passphrase since at least 2012.  Thus pam_unix should error out in case of a failure in crypt(3) instead of silently computing a hashed passphrase using a potentially unsafe method.

* modules/pam_unix/pam_unix.8.xml: Update documentation.
* modules/pam_unix/passverify.c (create_password_hash): Return NULL
on error instead of silently invoke crypt_md5_wrapper().

***

As this is a long standing downstream patch from Fedora / Red Hat, this should be reviewed (or at least commented) by the Debian folks.

### DIFF
--- a/modules/pam_unix/pam_unix.8.xml
+++ b/modules/pam_unix/pam_unix.8.xml
@@ -293,11 +293,10 @@
         <listitem>
           <para>
             When a user changes their password next,
-            encrypt it with the SHA256 algorithm. If the
-            SHA256 algorithm is not known to the <citerefentry>
+            encrypt it with the SHA256 algorithm. The
+            SHA256 algorithm must be supported by the <citerefentry>
 	    <refentrytitle>crypt</refentrytitle><manvolnum>3</manvolnum>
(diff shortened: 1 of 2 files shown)


besser82:
@t8m Could you please ask the Debian folks for reviewing this?

t8m:
I asked them on the pam-developers mailing list. Let's see if they reply, but do not except the reply early.

besser82:
@vorlonofportland Any thoughts or comments?

besser82:
Mhh…  No response,yet…  :/

t8m:
I am going to merge it anyway. This is IMO the right thing to do.

PR #80, opened 2018-11-25
pam_unix: Do not fail passphraseless sudo on locked accounts.

besser82:
Commit 4da9febc39b9 introduced a regression that made passphraseless sudo fail when it was invoked from a user with a locked passphrase.  Thus we should check for such a scenario when evaluating the return value of crypt_checksalt(3).

* modules/pam_unix/passverify.c (check_shadow_expiry): Do not return
PAM_AUTHTOK_EXPIRED on locked user accounts.

***

@t8m This is the fix for [rhbz#1653023](https://bugzilla.redhat.com/show_bug.cgi?id=1653023) as shipped in pam-1.3.1-10.fc30.

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -266,8 +266,10 @@ PAMH_ARG_DECL(int check_shadow_expiry,
 	    && (curdays - spent->sp_lstchg > spent->sp_inact)
 	    && (curdays - spent->sp_lstchg > spent->sp_max + spent->sp_inact)
 	    && (spent->sp_max != -1) && (spent->sp_inact != -1))
-	    || (crypt_checksalt(spent->sp_pwdp) == CRYPT_SALT_METHOD_DISABLED)
-	    || (crypt_checksalt(spent->sp_pwdp) == CRYPT_SALT_INVALID)) {
+	    || (((crypt_checksalt(spent->sp_pwdp) == CRYPT_SALT_METHOD_DISABLED)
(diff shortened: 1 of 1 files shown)


modules/pam_unix/passverify.c
```diff
@@ -266,8 +266,10 @@ PAMH_ARG_DECL(int check_shadow_expiry,
 	    && (curdays - spent->sp_lstchg > spent->sp_inact)
 	    && (curdays - spent->sp_lstchg > spent->sp_max + spent->sp_inact)
 	    && (spent->sp_max != -1) && (spent->sp_inact != -1))
-	    || (crypt_checksalt(spent->sp_pwdp) == CRYPT_SALT_METHOD_DISABLED)
-	    || (crypt_checksalt(spent->sp_pwdp) == CRYPT_SALT_INVALID)) {
+	    || (((crypt_checksalt(spent->sp_pwdp) == CRYPT_SALT_METHOD_DISABLED)
+		 || (crypt_checksalt(spent->sp_pwdp) == CRYPT_SALT_INVALID))
+	        && ((spent->sp_pwdp != NULL)
```
t8m: Can you please factor this check for locked out passwords to a static function? And should there be also check for empty password hash? spent->sp_pwdp[0] == '\0' Another question is whether crypt_checksalt(spent->sp_pwdp) == CRYPT_SALT_INVALID should actually return the error. Most probably it should not.

t8m: The more I think about it, the more I am inclined to drop this part of the merged commit completely. It does not belong here.

t8m:
I've dropped this code altogether.

PR #81, opened 2018-11-26
pam_unix: Better implementation of checksalt for expired passphrases.

besser82:
libxcrypt can be build-time configured to support (or not support) various hashing methods.  Future versions will also have support for runtime configuration by the system's vendor and/or administrator.

For that reason users should be notified by pam if they cannot log into their account anymore because of such a change in the system's configuration of libxcrypt.

Also check for malformed des hashes like "$2...", which might have been generated by unsafe base64 encoding functions as in glibc prior to version 2.17 and get silently rejected by the crypt(3) function
of implemented in glibc 2.17 and later or libxcrypt version 4.1.0 and later.

The previous (and thus dropped) implementation of this feature had some flaws, which are fixed in this implementation.

* modules/pam_unix/passverify.c (checksalt_authtok_expired): New func.
* modules/pam_unix/passverify.c (check_shadow_expiry): Add checksalt for expired passphrases using the new function.

***

@t8m Do you agree with this implementation?  I mean, there should be some information for the user, why they cannot login anymore…


t8m:
No, I do not agree. This check does not belong here. It also will not give the user any information as the authentication process will not get to the point where this function would be called if the password hash is not accepted by the crypt lib.
You could add a check in the verify_pwd_hash function but do not modify the return value or the behavior of the function, just add syslog message if the hash is disabled i.e. CRYPT_SALT_METHOD_DISABLED is returned. Do not even try to check for invalid hash.

PR #82, opened 2018-11-26
pam_unix: Report unusable hashes found by checksalt to syslog.

besser82:
libxcrypt can be build-time configured to support (or not support) various hashing methods.  Future versions will also have support for runtime configuration by the system's vendor and/or administrator.

For that reason adminstrator should be notified by pam if users cannot log into their account anymore because of such a change in the system's configuration of libxcrypt.

Also check for malformed hashes, like descrypt hashes starting with "$2...", which might have been generated by unsafe base64 encoding functions as used in glibc <= 2.16.
Such hashes are likely to be rejected by many recent implementations of libcrypt.

* modules/pam_unix/passverify.c (verify_pwd_hash): Report unusable hashes found by checksalt to syslog.


modules/pam_unix/passverify.c
```diff
@@ -103,6 +103,40 @@ verify_pwd_hash(const char *p, char *hash, unsigned int nullok)
...
+			 * Check for malformed hashes, like descrypt hashes
+			 * starting with "$2...", which might have been
+			 * generated by unsafe base64 encoding functions
+			 * as used in glibc <= 2.16.
+			 * Such hashes are likely to be rejected by many
+			 * recent implementations of libcrypt.
+			 */
+			if (crypt_checksalt(hash) == CRYPT_SALT_INVALID) {
+				helper_log_err(LOG_ERR,
+				  "pam_unix(verify_pwd_hash): The setting "
```
t8m: s/The setting used in the hash does not name/The hash does not use/

modules/pam_unix/passverify.c
```diff
@@ -103,6 +103,40 @@ verify_pwd_hash(const char *p, char *hash, unsigned int nullok)
 			 * Ok, we don't know the crypt algorithm, but maybe
 			 * libcrypt knows about it? We should try it.
 			 */
+#if defined(CRYPT_CHECKSALT_AVAILABLE) && CRYPT_CHECKSALT_AVAILABLE
+			/*
+			 * Check for hashing methods that are disabled by
+			 * libcrypt configuration and/or system preset.
+			 */
+			if (crypt_checksalt(hash) == CRYPT_SALT_METHOD_DISABLED) {
```
t8m: Please call the crypt_checksalt just once and assign the result to a variable and then test it. Also do not write the whole hash to the system log - something like hash \"%.3s...\" should be fine.

modules/pam_unix/passverify.c
besser82: Fixed.

modules/pam_unix/passverify.c
besser82: Fixed.

besser82:
Should be fine now.

modules/pam_unix/passverify.c
```diff
@@ -103,6 +103,42 @@ verify_pwd_hash(const char *p, char *hash, unsigned int nullok)
...
+			 * libcrypt configuration and/or system preset.
+			 */
+			if (retval_checksalt == CRYPT_SALT_METHOD_DISABLED) {
+				/*
+				 * pam_syslog() needs a pam handle,
+				 * but that's not available here.
+				 */
+				helper_log_err(LOG_ERR,
+				  "pam_unix(verify_pwd_hash): The method "
+				  "for computing the hash "\"%.6s\" has been "
```
t8m: There seems to be extra " character. Did you try to build it?

besser82: Fixed now.  Yes, I tried to build it, but that was before I made some changes to the log strings, which likely have introduced the extra `"`.

t8m:
Merged, thank you.

PR #83, opened 2018-11-27
pam_unix: helper_log_err is not guaranteed to be generally available.

besser82:
We need a helper function for writing to syslog, that is guaranteed to be available for any build configuration.

* modules/pam_unix/passverify.c (verify_pwd_hash_log_err): New func.

***

@t8m This is additionally needed for #82 to not break the build in some configurations.


t8m:
You're right, but this is incorrect way to fix it. I'll fix it myself.

PR #84, opened 2018-11-27
pam_unix: Better approach to add crypt_default method, if supported.

besser82:
libxcrypt since v4.4.0 supports a default method for its gensalt function on most system configurations. 
 As the default method is to be considered the strongest available hash method, it should be preferred over all other hash methods supported by pam.

* modules/pam_unix/pam_unix.8.xml: Documentation for crypt_default.
* modules/pam_unix/passverify.c: Add crypt_default method.
* modules/pam_unix/support.c (_set_ctrl): Issue an error to syslog, if
the crypt_default method is not supported by the system's libcrypt
* modules/pam_unix/support.h: Likewise.

***

@t8m This is the new version of #78.

### DIFF
--- a/modules/pam_unix/pam_unix.8.xml
+++ b/modules/pam_unix/pam_unix.8.xml
@@ -363,6 +363,24 @@
       </varlistentry>
       <varlistentry>
         <term>
+          <option>crypt_default</option>
+        </term>
+        <listitem>
+          <para>
+            When a user changes their password next,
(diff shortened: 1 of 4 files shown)


t8m:
@besser82 If we could do something with this one before the 1.4.0 release it would be nice. But I'd like the release to be completed during March.

besser82:
@t8m working on it.  =)

PR #86, opened 2018-12-13
Fix a typo in the documentation

sbraz:
Hi, just a simple typo fix (with a preposition changed too).


ldv-alt:
There is one more typo in that sentence that needs to be fixed. :)

sbraz:
@ldv-alt you mean "does"? I kept it because I read it as "the way that Solaris' PAM does" without the parentheses.

t8m:
No, it is the 'derivitives'.

sbraz:
Thanks, I couldn't even see it :)

PR #90, opened 2018-12-31
Fix build on OpenWrt

neheb:
These are a few patches that are kept in OpenWrt in order to fix compilation with musl. These should be the non-OpenWrt specific ones.

### DIFF
--- a/modules/pam_exec/pam_exec.c
+++ b/modules/pam_exec/pam_exec.c
@@ -102,7 +102,7 @@ call_exec (const char *pam_type, pam_handle_t *pamh,
   int use_stdout = 0;
   int optargc;
   const char *logfile = NULL;
-  const char *authtok = NULL;
+  char authtok[PAM_MAX_RESP_SIZE] = {};
   pid_t pid;
   int fds[2];
   int stdout_fds[2];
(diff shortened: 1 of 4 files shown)


modules/pam_exec/pam_exec.c
```diff
@@ -180,12 +180,12 @@ call_exec (const char *pam_type, pam_handle_t *pamh,
...
 		{
 		  pam_set_item (pamh, PAM_AUTHTOK, resp);
-		  authtok = strndupa (resp, PAM_MAX_RESP_SIZE);
+		  strcpy (authtok, resp);
 		  _pam_drop (resp);
 		}
 	    }
 	  else
-	    authtok = strndupa (void_pass, PAM_MAX_RESP_SIZE);
+	    strcpy (authtok, void_pass);
```
t8m: This is unsafe. Although the password if obtained from conversation function cannot be larger than PAM_MAX_RESP_SIZE, in theory some module can set the PAM_AUTHTOK item to any arbitrarily large string. So to be on the safe side, please use strncpy correctly and do it in the above call as well.

neheb:
I think this is what you meant?

modules/pam_exec/pam_exec.c
```diff
@@ -180,12 +180,12 @@ call_exec (const char *pam_type, pam_handle_t *pamh,
 	      if (resp)
 		{
 		  pam_set_item (pamh, PAM_AUTHTOK, resp);
-		  authtok = strndupa (resp, PAM_MAX_RESP_SIZE);
+		  strncpy (authtok, resp, PAM_MAX_RESP_SIZE - 1);
```
t8m: It would be better to use sizeof(authtok) - 1 instead.

neheb:
done.

t8m:
There were bugs in both Makefile.am and configure.ac patches. I've fixed them.

Issue #92, opened 2019-01-09
Adding pam_setquota


shartge:
I've "rescued" and "maintained" a [pam_setquota](https://github.com/shartge/pam_setquota) module for several years and use it as locally compiled PAM module on several hosts in production.

But since I only know enough C to be dangerous and adapting and fixing bugs for this module is more than I can handle, it would be nice if it could eventually be included in the main PAM code.

t8m:
This looks like nice idea. Supplying a PR integrating it would help.

shartge:
I'll see what I can do to automakify the code before submitting it.

PR #94, opened 2019-01-10
pam_setquota: new module to set or modify disk quotas on session start

shartge:
This module makes disk quotas usable with central user databases, such as MySQL or
LDAP.

Resolves: #92

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -576,9 +576,10 @@ AC_FUNC_MEMCMP
 AC_FUNC_VPRINTF
 AC_CHECK_FUNCS(fseeko getdomainname gethostname gettimeofday lckpwdf mkdir select)
 AC_CHECK_FUNCS(strcspn strdup strspn strstr strtol uname)
-AC_CHECK_FUNCS(getutent_r getpwnam_r getpwuid_r getgrnam_r getgrgid_r getspnam_r)
+AC_CHECK_FUNCS(getutent_r getpwnam_r getpwuid_r getgrnam_r getgrgid_r getspnam_r getmntent_r)
 AC_CHECK_FUNCS(getgrouplist getline getdelim)
 AC_CHECK_FUNCS(inet_ntop inet_pton innetgr)
+AC_CHECK_FUNCS(quotactl)
 AC_CHECK_FUNCS([ruserok_af ruserok], [break])
 BACKUP_LIBS=$LIBS
 LIBS="$LIBS -lutil"
@@ -587,6 +588,7 @@ LIBS=$BACKUP_LIBS
 
 AM_CONDITIONAL([COND_BUILD_PAM_RHOSTS], [test "$ac_cv_func_ruserok_af" = yes -o "$ac_cv_func_ruserok" = yes])
 AM_CONDITIONAL([COND_BUILD_PAM_LASTLOG], [test "$ac_cv_func_logwtmp" = yes])
+AM_CONDITIONAL([COND_BUILD_PAM_SETQUOTA], [test "$ac_cv_func_quotactl" = yes])
 
 AC_CHECK_FUNCS(unshare, [UNSHARE=yes], [UNSHARE=no])
 AM_CONDITIONAL([HAVE_UNSHARE], [test "$UNSHARE" = yes])
@@ -703,7 +705,7 @@ AC_CONFIG_FILES([Makefile libpam/Makefile libpamc/Makefile libpamc/test/Makefile
 	modules/pam_pwhistory/Makefile modules/pam_rhosts/Makefile \
 	modules/pam_rootok/Makefile modules/pam_exec/Makefile \
 	modules/pam_securetty/Makefile modules/pam_selinux/Makefile \
-	modules/pam_sepermit/Makefile \
+	modules/pam_sepermit/Makefile modules/pam_setquota/Makefile \
 	modules/pam_shells/Makefile modules/pam_stress/Makefile \
 	modules/pam_succeed_if/Makefile modules/pam_tally/Makefile \
 	modules/pam_tally2/Makefile modules/pam_time/Makefile \
--- a/modules/pam_setquota/pam_setquota.8.xml
+++ b/modules/pam_setquota/pam_setquota.8.xml
@@ -1,0 +1,301 @@
+﻿<?xml version="1.0" encoding='UTF-8'?>
+<!DOCTYPE refentry PUBLIC "-//OASIS//DTD DocBook XML V4.3//EN" "http://www.oasis-open.org/docbook/xml/4.3/docbookx.dtd">
+
+<refentry id="pam_setquota">
+
+  <refmeta>
+    <refentrytitle>pam_setquota</refentrytitle>
+    <manvolnum>8</manvolnum>
+    <refmiscinfo class="sectdesc">Linux-PAM Manual</refmiscinfo>
+  </refmeta>
+
+  <refnamediv id="pam_setquota-name">
+    <refname>pam_setquota</refname>
+    <refpurpose>PAM module to set or modify disk quotas on session start</refpurpose>
+  </refnamediv>
+
+  <refsynopsisdiv>
+    <cmdsynopsis id="pam_setquota-cmdsynopsis">
+      <command>pam_setquota.so</command>
+       <arg choice="opt">
+        fs=<replaceable>/home</replaceable>
+      </arg>
+      <arg choice="opt">
+        overwrite=<replaceable>0</replaceable>
+      </arg>
+      <arg choice="opt">
+        debug=<replaceable>0</replaceable>
+      </arg>
+      <arg choice="opt">
+        startuid=<replaceable>1000</replaceable>
+      </arg>
+       <arg choice="opt">
+        enduid=<replaceable>0</replaceable>
+      </arg>
+       <arg choice="opt">
+        bsoftlimit=<replaceable>19000</replaceable>
+      </arg>
+       <arg choice="opt">
+        bhardlimit=<replaceable>20000</replaceable>
+      </arg>
+       <arg choice="opt">
+        isoftlimit=<replaceable>3000</replaceable>
+      </arg>
+       <arg choice="opt">
+        ihardlimit=<replaceable>4000</replaceable>
+      </arg>
+    </cmdsynopsis>
+  </refsynopsisdiv>
+
+  <refsect1 id="pam_setquota-description">
+
+    <title>DESCRIPTION</title>
+
+    <para>
+      pam_setquota is a PAM module to set or modify a disk quota at session start
+    </para>
+    <para>
+      This makes quotas usable with central user databases, such as MySQL or LDAP.
+    </para>
+
+  </refsect1>
+
+  <refsect1 id="pam_setquota-options">
+
+    <title>OPTIONS</title>
+    <para>
+      <variablelist>
+        <varlistentry>
+          <term>
+            <option>fs=<replaceable>/home</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+             The device file or mountpoint the policy applies to. Defaults to the
+             filesystem containing the users home directory.
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>overwrite=<replaceable>0</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+             Overwrite an existing quota. Note: Enabling this will remove the ability
+             for the admin to manually configure different quotas for users for a
+             filesystem with <citerefentry><refentrytitle>edquota</refentrytitle>
+             <manvolnum>8</manvolnum></citerefentry>. (Defaults to 0)
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>debug=<replaceable>0</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+             Enable debugging. A value of 1 outputs the old and new quota on a device.
+             A value of 2 also prints out the matched and found filesystems should
+             <option>fs</option> be unset. (Defaults to 0)
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>startuid=<replaceable>1000</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+             Describe the start of the UID range the policy is applied to.
+             (Defaults to UID_MIN from login.defs or the uidmin value defined
+             at compile-time if UID_MIN is undefined)
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>enduid=<replaceable>0</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+             Describe the end of the UID range the policy is applied to. Setting
+             <emphasis>enduid=0</emphasis> results in an open-ended UID
+             range (i.e. all uids greater than <option>startuid</option> are
+             included). (Defaults to 0)
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>bsoftlimit=<replaceable>19000</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+              Soft limit for disk quota blocks, as defined by <citerefentry>
+              <refentrytitle>quotactl</refentrytitle><manvolnum>2</manvolnum>
+              </citerefentry>.
+              Note: <option>bsoftlimit</option> and <option>bhardlimit</option>
+              <emphasis>must</emphasis> be set at the same time!
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>bhardlimit=<replaceable>20000</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+              Hard limit for disk quota blocks, as defined by <citerefentry>
+              <refentrytitle>quotactl</refentrytitle><manvolnum>2</manvolnum>
+              </citerefentry>.
+              Note: <option>bsoftlimit</option> and <option>bhardlimit</option>
+              <emphasis>must</emphasis> be set at the same time!
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>isoftlimit=<replaceable>3000</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+              Soft limit for inodes, as defined by <citerefentry><refentrytitle>
+              quotactl</refentrytitle><manvolnum>2</manvolnum></citerefentry>.
+              Note: <option>isoftlimit</option> and <option>ihardlimit</option>
+              <emphasis>must</emphasis> be set at the same time!
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term>
+            <option>ihardlimit=<replaceable>4000</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+              Hard limit for inodes, as defined by <citerefentry><refentrytitle>
+              quotactl</refentrytitle><manvolnum>2</manvolnum></citerefentry>.
+              Note: <option>isoftlimit</option> and <option>ihardlimit</option>
+              <emphasis>must</emphasis> be set at the same time!
+            </para>
+          </listitem>
+        </varlistentry>
+      </variablelist>
+    </para>
+  </refsect1>
+
+  <refsect1 id="pam_setquota-types">
+    <title>MODULE TYPES PROVIDED</title>
+    <para>
+       Only the <option>session</option> module type is provided.
+    </para>
+  </refsect1>
+
+  <refsect1 id='pam_setquota-return_values'>
+    <title>RETURN VALUES</title>
+    <para>
+      <variablelist>
+
+        <varlistentry>
+          <term>PAM_SUCCESS</term>
+          <listitem>
+            <para>
+              The quota was set successfully.
+            </para>
+          </listitem>
+        </varlistentry>
+
+	<varlistentry>
+          <term>PAM_IGNORE</term>
+          <listitem>
+            <para>
+              No action was taken because either the UID of the user was outside
+              of the specifed range, a quota already existed and
+              <option>overwrite=1</option> was not configured or no limits were
+              configured at all.
+            </para>
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term>PAM_CRED_INSUFFICIENT</term>
+          <listitem>
+            <para>
+              The user was not found.
+            </para>
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term>PAM_PERM_DENIED</term>
+          <listitem>
+            <para>
+               <filename>/proc/mounts</filename> could not be opened.
+            </para>
+            <para>
+               The filesystem or device specified was not found.
+            </para>
+            <para>
+               The limits for the user could not be retrieved. See syslog for
+               more information.
+            </para>
+            <para>
+               The limits for the user could not be set. See syslog for
+               more information.
+            </para>
+            <para>
+               Either <option>isoftlimit</option>/<option>ihardlimit</option>
+               or <option>bsoftlimit</option>/<option>bhardlimit</option> were
+               not set at the same time.
+            </para>
+          </listitem>
+        </varlistentry>
+
+      </variablelist>
+    </para>
+  </refsect1>
+
+  <refsect1 id='pam_setquota-examples'>
+    <title>EXAMPLES</title>
+    <para>
+    A single invocation of `pam_setquota` applies a specific policy to a UID
+    range. Applying different policies to specific UID ranges is done by
+    invoking pam_setquota more than once. The last matching entry
+    defines the resulting quota.
+    <programlisting>
+      session  required   pam_setquota.so bsoftlimit=1000 bhardlimit=2000 isoftlimit=1000 ihardlimit=2000 startuid=1000 enduid=0 fs=/home
+      session  required   pam_setquota.so bsoftlimit=19000 bhardlimit=20000 isoftlimit=3000 ihardlimit=4000 startuid=2001 enduid=3000 fs=/dev/sda1
+      session  required   pam_setquota.so bsoftlimit=19000 bhardlimit=20000 isoftlimit=3000 ihardlimit=4000 startuid=3001 enduid=4000 fs=/dev/sda1 overwrite=1
+    </programlisting>
+    </para>
+  </refsect1>
+
+  <refsect1 id='pam_setquota-see_also'>
+    <title>SEE ALSO</title>
+    <para>
+      <citerefentry>
+        <refentrytitle>pam.conf</refentrytitle><manvolnum>5</manvolnum>
+      </citerefentry>,
+      <citerefentry>
+        <refentrytitle>pam.d</refentrytitle><manvolnum>5</manvolnum>
+      </citerefentry>,
+      <citerefentry>
+        <refentrytitle>pam</refentrytitle><manvolnum>8</manvolnum>
+      </citerefentry>
+    </para>
+  </refsect1>
+
+  <refsect1 id='pam_setquota-author'>
+    <title>AUTHOR</title>
+      <para>
+       pam_setquota was originally written by
+       Ruslan Savchenko &lt;savrus@mexmat.net&gt;.
+      </para>
+      <para>
+       Further modifications were made by Shane Tzen&lt;shane@ict.usc.edu&gt;,
+       Sven Hartge &lt;sven@svenharte.de&gt;
+       and Keller Fuchs &lt;kellerfuchs@hashbang.sh&gt;
+      </para>
+  </refsect1>
+
+</refentry>
--- a/modules/pam_setquota/pam_setquota.c
+++ b/modules/pam_setquota/pam_setquota.c
@@ -1,0 +1,396 @@
+/* PAM setquota module
+
+   This PAM module sets disk quota when a session begins.
+
+   Copyright © 2006 Ruslan Savchenko <savrus@mexmat.net>
+   Copyright © 2010 Shane Tzen <shane@ict.usc.edu>
+   Copyright © 2012-2020 Sven Hartge <sven@svenhartge.de>
+   Copyright © 2016 Keller Fuchs <kellerfuchs@hashbang.sh>
+
+*/
+
+#include <sys/types.h>
+#include <sys/quota.h>
+#include <linux/quota.h>
+#include <pwd.h>
+#include <syslog.h>
+#include <errno.h>
+#include <mntent.h>
+#include <stdio.h>
+#include <stdbool.h>
+
+#define PAM_SM_SESSION
+
+#include <security/pam_modules.h>
+#include <security/_pam_macros.h>
+#include <security/pam_ext.h>
+#include <security/pam_modutil.h>
+#include "pam_inline.h"
+
+#ifndef PATH_LOGIN_DEFS
+# define PATH_LOGIN_DEFS "/etc/login.defs"
+#endif
+
+#define MAX_UID_VALUE 0xFFFFFFFFUL
+
+struct pam_params {
+  uid_t start_uid;
+  uid_t end_uid;
+  const char *fs;
+  size_t fs_len;
+  int overwrite;
+  int debug;
+};
+
+static inline void
+debug(pam_handle_t *pamh, const struct if_dqblk *p,
+      const char *device, const char *dbgprefix) {
+  pam_syslog(pamh, LOG_DEBUG, "%s device=%s bsoftlimit=%llu bhardlimit=%llu "
+                              "isoftlimit=%llu ihardlimit=%llu btime=%llu itime=%llu",
+             dbgprefix, device,
+             p->dqb_bsoftlimit, p->dqb_bhardlimit,
+             p->dqb_isoftlimit, p->dqb_ihardlimit,
+             p->dqb_btime, p->dqb_itime);
+}
+
+static unsigned long long
+str_to_dqb_num(pam_handle_t *pamh, const char *str, const char *param) {
+  char *ep = NULL;
+
+  errno = 0;
+  long long temp = strtoll(str, &ep, 10);
+  if (temp < 0 || str == ep || *ep != '\0' || errno !=0) {
+    pam_syslog(pamh, LOG_ERR, "Parameter \"%s=%s\" invalid, setting to 0", param, str);
+    return 0;
+  }
+  else {
+    return temp;
+  }
+}
+
+static bool
+parse_dqblk(pam_handle_t *pamh, int argc, const char **argv, struct if_dqblk *p) {
+  bool bhard = false, bsoft = false, ihard = false, isoft = false;
+
+  /* step through arguments */
+  for (; argc-- > 0; ++argv) {
+    const char *str;
+    if ((str = pam_str_skip_prefix(*argv, "bhardlimit=")) != NULL) {
+      p->dqb_bhardlimit = str_to_dqb_num(pamh, str, "bhardlimit");
+      p->dqb_valid |= QIF_BLIMITS;
+      bhard = true;
+    } else if ((str = pam_str_skip_prefix(*argv, "bsoftlimit=")) != NULL) {
+      p->dqb_bsoftlimit = str_to_dqb_num(pamh, str, "bsoftlimit");
+      p->dqb_valid |= QIF_BLIMITS;
+      bsoft = true;
+    } else if ((str = pam_str_skip_prefix(*argv, "ihardlimit=")) != NULL) {
+      p->dqb_ihardlimit = str_to_dqb_num(pamh, str, "ihardlimit");
+      p->dqb_valid |= QIF_ILIMITS;
+      ihard = true;
+    } else if ((str = pam_str_skip_prefix(*argv, "isoftlimit=")) != NULL) {
+      p->dqb_isoftlimit = str_to_dqb_num(pamh, str, "isoftlimit");
+      p->dqb_valid |= QIF_ILIMITS;
+      isoft = true;
+    } else if ((str = pam_str_skip_prefix(*argv, "btime=")) != NULL) {
+      p->dqb_btime = str_to_dqb_num(pamh, str, "btime");
+      p->dqb_valid |= QIF_BTIME;
+    } else if ((str = pam_str_skip_prefix(*argv, "itime=")) != NULL) {
+      p->dqb_itime = str_to_dqb_num(pamh, str, "itime");
+      p->dqb_valid |= QIF_ITIME;
+    }
+  }
+
+  /* return FALSE if a softlimit or hardlimit has been set
+   * independently of its counterpart.
+   */
+  return !(bhard ^ bsoft) && !(ihard ^ isoft);
+}
+
+/* inspired by pam_usertype_get_id */
+static uid_t
+str_to_uid(pam_handle_t *pamh, const char *value, uid_t default_value, const char *param) {
+    unsigned long ul;
+    char *ep;
+    uid_t uid;
+
+    errno = 0;
+    ul = strtoul(value, &ep, 10);
+    if (!(ul >= MAX_UID_VALUE
+        || (uid_t)ul >= MAX_UID_VALUE
+        || (errno != 0 && ul == 0)
+        || value == ep
+        || *ep != '\0')) {
+        uid = (uid_t)ul;
+    } else {
+        pam_syslog(pamh, LOG_ERR, "Parameter \"%s=%s\" invalid, "
+                   "setting to %u", param, value, default_value);
+        uid = default_value;
+    }
+
+    return uid;
+}
+
+static void
+parse_params(pam_handle_t *pamh, int argc, const char **argv, struct pam_params *p) {
+  /* step through arguments */
+  for (; argc-- > 0; ++argv) {
+    const char *str;
+    char *ep = NULL;
+    if ((str = pam_str_skip_prefix(*argv, "startuid=")) != NULL) {
+      p->start_uid = str_to_uid(pamh, str, p->start_uid, "startuid");
+    } else if ((str = pam_str_skip_prefix(*argv, "enduid=")) != NULL) {
+      p->end_uid = str_to_uid(pamh, str, p->end_uid, "enduid");
+    } else if ((str = pam_str_skip_prefix(*argv, "fs=")) != NULL) {
+      p->fs = str;
+      p->fs_len = strlen(str);
+      /* Mask the unnecessary '/' from the end of fs parameter */
+      if (p->fs_len > 1 && p->fs[p->fs_len - 1] == '/')
+        --p->fs_len;
+    } else if ((str = pam_str_skip_prefix(*argv, "overwrite=")) != NULL) {
+      errno = 0;
+      p->overwrite = strtol(str, &ep, 10);
+      if (*ep != '\0' || str == ep || errno !=0 || (p->overwrite < 0)) {
+        pam_syslog(pamh, LOG_ERR, "Parameter \"overwrite=%s\" invalid, "
+                        "setting to 0", str);
+        p->overwrite = 0;
+      }
+    } else if ((str = pam_str_skip_prefix(*argv, "debug=")) != NULL) {
+      errno = 0;
+      p->debug = strtol(str, &ep, 10);
+      if (*ep != '\0' || str == ep || errno != 0 || (p->debug < 0)) {
+        pam_syslog(pamh, LOG_ERR, "Parameter \"debug=%s\" invalid, "
+                        "setting to 0", str);
+        p->debug = 0;
+      }
+    }
+  }
+}
+
+PAM_EXTERN int pam_sm_open_session(pam_handle_t *pamh, int flags, int argc,
+                                   const char **argv) {
+  int retval;
+  unsigned long ul;
+  char *ep, *val, *mntdevice = NULL;
+  const void *user;
+  const struct passwd *pwd;
+  struct pam_params param = {
+          .start_uid = PAM_USERTYPE_UIDMIN,
+          .end_uid = 0,
+          .fs = NULL };
+  struct if_dqblk ndqblk;
+  FILE *fp;
+  size_t mnt_len = 0, match_size = 0;
+#ifdef HAVE_GETMNTENT_R
+  char buf[BUFSIZ];
+  struct mntent ent;
+#endif
+  const struct mntent *mnt;
+  const char *service;
+
+  if (pam_get_item(pamh, PAM_SERVICE, (const void **)&service) != PAM_SUCCESS)
+    service = "";
+
+  /* Get UID_MIN for default start_uid from login.defs */
+  val = pam_modutil_search_key(pamh, PATH_LOGIN_DEFS, "UID_MIN");
+
+  /* Should UID_MIN be undefined, use current value of param.start_uid
+   * pre-defined as PAM_USERTYPE_UIDMIN set by configure as safe
+   * starting UID to avoid setting a quota for root and system
+   * users if startuid= parameter is absent.
+   */
+  if (val) {
+    param.start_uid = str_to_uid(pamh, val, param.start_uid, PATH_LOGIN_DEFS":UID_MIN");
+  }
+
+  /* Parse parameter values
+   * Must come after pam_modutil_search_key so that the current system
+   * default for UID_MIN is already in p.start_uid to serve as default
+   * for str_to_uid in case of a parse error.
+   * */
+  parse_params(pamh, argc, argv, &param);
+
+  if (param.debug >= 1)
+    pam_syslog(pamh, LOG_DEBUG, "Config: startuid=%u enduid=%u fs=%s "
+                    "debug=%d overwrite=%d",
+                    param.start_uid, param.end_uid,
+                    param.fs ? param.fs : "(none)",
+                    param.debug, param.overwrite);
+
+  /* Determine the user name so we can get the home directory */
+  retval = pam_get_item(pamh, PAM_USER, &user);
+  if (retval != PAM_SUCCESS || user == NULL || *(const char *)user == '\0') {
+    pam_syslog(pamh, LOG_NOTICE, "user unknown");
+    return PAM_USER_UNKNOWN;
+  }
+
+  /* Get the password entry */
+  pwd = pam_modutil_getpwnam(pamh, user);
+  if (pwd == NULL) {
+    return PAM_CRED_INSUFFICIENT;
+  }
+
+  /* Check if we should not set quotas for user */
+  if ((pwd->pw_uid < param.start_uid) ||
+      ((param.end_uid >= param.start_uid) && (param.start_uid != 0) &&
+       (pwd->pw_uid > param.end_uid)))
+    return PAM_SUCCESS;
+
+  /* Find out what device the filesystem is hosted on */
+  if ((fp = setmntent("/proc/mounts", "r")) == NULL) {
+    pam_syslog(pamh, LOG_ERR, "Unable to open /proc/mounts");
+    return PAM_PERM_DENIED;
+  }
+
+  while (
+#ifdef HAVE_GETMNTENT_R
+           (mnt = getmntent_r(fp, &ent, buf, sizeof(buf))) != NULL
+#else
+           (mnt = getmntent(fp)) != NULL
+#endif
+        ) {
+    /* If param.fs is not specified use filesystem with users homedir
+     * as default.
+     */
+    if (param.fs == NULL) {
+      /* Mask trailing / from mnt->mnt_dir, to get a leading / on the
+       * remaining suffix returned by pam_str_skip_prefix_len()
+       */
+      for (mnt_len = strlen(mnt->mnt_dir); mnt_len > 0; --mnt_len)
+        if (mnt->mnt_dir[mnt_len - 1] != '/')
+          break;
+      const char *s;
+      if (param.debug >= 2)
+        pam_syslog(pamh, LOG_DEBUG, "Trying to match pw_dir=\"%s\" "
+                        "with mnt_dir=\"%s\"", pwd->pw_dir, mnt->mnt_dir);
+      /*
+       * (mnt_len > match_size) Only try matching the mnt_dir if its length
+       * is longer than the last matched length, trying to find the longest
+       * mnt_dir for a given pwd_dir.
+       *
+       * (mnt_len == 0 && mnt->mnt_dir[0] == '/') special-cases the
+       * root-dir /, which is the only mnt_dir with a trailing '/', which
+       * got masked earlier.
+       */
+      if ((mnt_len > match_size || (mnt_len == 0 && mnt->mnt_dir[0] == '/')) &&
+         (s = pam_str_skip_prefix_len(pwd->pw_dir, mnt->mnt_dir, mnt_len)) != NULL &&
+         (s[0] == '\0' || s[0] == '/')) {
+        free(mntdevice);
+        if ((mntdevice = strdup(mnt->mnt_fsname)) == NULL) {
+          pam_syslog(pamh, LOG_CRIT, "Memory allocation error");
+          endmntent(fp);
+          return PAM_PERM_DENIED;
+        }
+        match_size = mnt_len;
+        if (param.debug >= 2)
+          pam_syslog(pamh, LOG_DEBUG, "Found pw_dir=\"%s\" in mnt_dir=\"%s\" "
+                     "with suffix=\"%s\" on device=\"%s\"", pwd->pw_dir,
+                     mnt->mnt_dir, s, mntdevice);
+      }
+    /* param.fs has been specified, find exactly matching fileystem */
+    } else if ((strncmp(param.fs, mnt->mnt_dir, param.fs_len) == 0
+                && mnt->mnt_dir[param.fs_len] == '\0') ||
+               (strncmp(param.fs, mnt->mnt_fsname, param.fs_len) == 0
+                && mnt->mnt_fsname[param.fs_len] == '\0' )) {
+        free(mntdevice);
+        if ((mntdevice = strdup(mnt->mnt_fsname)) == NULL) {
+          pam_syslog(pamh, LOG_CRIT, "Memory allocation error");
+          endmntent(fp);
+          return PAM_PERM_DENIED;
+        }
+        if (param.debug >= 2)
+          pam_syslog(pamh, LOG_DEBUG, "Found fs=\"%s\" in mnt_dir=\"%s\" "
+                     "on device=\"%s\"", param.fs, mnt->mnt_dir, mntdevice);
+    }
+  }
+
+  endmntent(fp);
+
+  if (mntdevice == NULL) {
+    pam_syslog(pamh, LOG_ERR, "Filesystem or device not found: %s", param.fs ? param.fs : pwd->pw_dir);
+    return PAM_PERM_DENIED;
+  }
+
+  /* Get limits */
+  if (quotactl(QCMD(Q_GETQUOTA, USRQUOTA), mntdevice, pwd->pw_uid,
+               (void *)&ndqblk) == -1) {
+    pam_syslog(pamh, LOG_ERR, "fail to get limits for user %s : %m",
+               pwd->pw_name);
+    free(mntdevice);
+    return PAM_PERM_DENIED;
+  }
+
+  if (param.debug >= 1)
+    debug(pamh, &ndqblk, mntdevice, "Quota read:");
+
+  /* Only overwrite if quotas aren't already set or if overwrite is set */
+  if ((ndqblk.dqb_bsoftlimit == 0 && ndqblk.dqb_bhardlimit == 0 &&
+       ndqblk.dqb_isoftlimit == 0 && ndqblk.dqb_ihardlimit == 0) ||
+      param.overwrite == 1) {
+
+    /* Parse new limits
+     * Exit with an error should only the hard- or softlimit be
+     * configured but not both.
+     * This avoids errors, inconsistencies and possible race conditons
+     * during setquota.
+     */
+    ndqblk.dqb_valid = 0;
+    if (!parse_dqblk(pamh, argc, argv, &ndqblk)) {
+      pam_syslog(pamh, LOG_ERR,
+                 "Both soft- and hardlimits for %s need to be configured "
+                 "at the same time!", mntdevice);
+      free(mntdevice);
+      return PAM_PERM_DENIED;
+    }
+
+    /* Nothing changed? Are no limits defined at all in configuration? */
+    if (ndqblk.dqb_valid == 0) {
+      pam_syslog(pamh, LOG_AUTH | LOG_WARNING, "no limits defined in "
+                 "configuration for user %s on %s", pwd->pw_name, mntdevice);
+      free(mntdevice);
+      return PAM_IGNORE;
+    }
+
+    /* Set limits */
+    if (quotactl(QCMD(Q_SETQUOTA, USRQUOTA), mntdevice, pwd->pw_uid,
+                 (void *)&ndqblk) == -1) {
+      pam_syslog(pamh, LOG_ERR, "failed to set limits for user %s on %s: %m",
+                 pwd->pw_name, mntdevice);
+      free(mntdevice);
+      return PAM_PERM_DENIED;
+    }
+    if (param.debug >= 1)
+      debug(pamh, &ndqblk, mntdevice, "Quota set:");
+
+    /* End module */
+    free(mntdevice);
+    return PAM_SUCCESS;
+
+  } else {
+    /* Quota exists and overwrite!=1 */
+    if (param.debug >= 1) {
+      pam_syslog(pamh, LOG_DEBUG, "Quota already exists for user %s "
+                 "on %s, not overwriting it without \"overwrite=1\"",
+                 pwd->pw_name, mntdevice);
+    }
+    /* End module */
+    free(mntdevice);
+    return PAM_IGNORE;
+  }
+
+}
+
+PAM_EXTERN
+int pam_sm_close_session(pam_handle_t *pamh, int flags, int argc,
+                         const char **argv) {
+  return PAM_SUCCESS;
+}
+
+#ifdef PAM_STATIC
+
+/* static module data */
+struct pam_module _pam_setquota_modstruct = {
+  "pam_setquota",      NULL,                 NULL, NULL,
+  pam_sm_open_session, pam_sm_close_session, NULL
+};
+
+#endif
--- a/doc/sag/pam_setquota.xml
+++ b/doc/sag/pam_setquota.xml
@@ -1,0 +1,34 @@
+﻿<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE section PUBLIC "-//OASIS//DTD DocBook XML V4.4//EN"
+        "http://www.oasis-open.org/docbook/xml/4.4/docbookx.dtd">
+<section id='sag-pam_setquota'>
+  <title>pam_setquota - set or modify disk quotas on session start</title>
+  <cmdsynopsis>
+    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude"
+     href="../../modules/pam_setquota/pam_setquota.8.xml" xpointer='xpointer(//cmdsynopsis[@id = "pam_setquota-cmdsynopsis"]/*)'/>
+  </cmdsynopsis>
+  <section id='sag-pam_setquota-description'>
+    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude"
+     href="../../modules/pam_setquota/pam_setquota.8.xml" xpointer='xpointer(//refsect1[@id = "pam_setquota-description"]/*)'/>
+  </section>
+  <section id='sag-pam_setquota-options'>
+    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude"
+     href="../../modules/pam_setquota/pam_setquota.8.xml" xpointer='xpointer(//refsect1[@id = "pam_setquota-options"]/*)'/>
+  </section>
+  <section id='sag-pam_setquota-types'>
+    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude"
+     href="../../modules/pam_setquota/pam_setquota.8.xml" xpointer='xpointer(//refsect1[@id = "pam_setquota-types"]/*)'/>
+  </section>
+  <section id='sag-pam_setquota-return_values'>
+    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude"
+     href="../../modules/pam_setquota/pam_setquota.8.xml" xpointer='xpointer(//refsect1[@id = "pam_setquota-return_values"]/*)'/>
+  </section>
+  <section id='sag-pam_setquota-examples'>
+    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude"
+     href="../../modules/pam_setquota/pam_setquota.8.xml" xpointer='xpointer(//refsect1[@id = "pam_setquota-examples"]/*)'/>
+  </section>
+  <section id='sag-pam_setquota-author'>
+    <xi:include xmlns:xi="http://www.w3.org/2001/XInclude"
+     href="../../modules/pam_setquota/pam_setquota.8.xml" xpointer='xpointer(//refsect1[@id = "pam_setquota-author"]/*)'/>
+  </section>
+</section>
--- a/modules/Makefile.am
+++ b/modules/Makefile.am
@@ -10,6 +10,10 @@ if COND_BUILD_PAM_LASTLOG
  MAYBE_PAM_LASTLOG = pam_lastlog
 endif
 
+if COND_BUILD_PAM_SETQUOTA
+ MAYBE_PAM_SETQUOTA = pam_setquota
+endif
+
 SUBDIRS := pam_access pam_cracklib pam_debug pam_deny pam_echo \
 	pam_env pam_exec pam_faildelay pam_filter pam_ftp \
 	pam_group pam_issue pam_keyinit pam_limits \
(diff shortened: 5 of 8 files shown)


t8m:
@ldv-alt what do you think about it?

modules/pam_setquota/pam_setquota.8.xml
```diff
@@ -0,0 +1,239 @@
...
+      </variablelist>
+    </para>
+  </refsect1>
+
+  <refsect1 id='pam_setquota-examples'>
+    <title>EXAMPLES</title>
+    <para>
+    A single invocation of `pam_setquota` applies a specific policy to a UID range. Applying different policies to specific UID ranges is done by invoking pam_setquota more than once. The last matching entry defines the resulting quota.
+      <programlisting>
+	session    required     /lib/security/pam_setquota.so bsoftlimit=1000 bhardlimit=2000 isoftlimit=1000 ihardlimit=2000 startuid=1000 enduid=0 fs=/home
```
thkukuk: The examples are wrong, never use an absolute path, especially not one which is wrong on most systems.

shartge: Right. I will fix this to be more in line with the documentation of the other modules.

modules/pam_setquota/pam_setquota.8.xml
```diff
@@ -0,0 +1,239 @@
...
+    <para>
+      <variablelist>
+
+        <varlistentry>
+          <term>
+            <option>overwrite</option>
+          </term>
+          <listitem>
+            <para>
+	      Override an existing quote. Note: Enabling this will remove the ability for the admin to manually configure different quotas for users for a filesystem with <citerefentry><refentrytitle>edquota</refentrytitle><manvolnum>8</manvolnum></citerefentry>.
```
t8m: typo: quote. Can you also please wrap the long lines?

modules/pam_setquota/pam_setquota.8.xml
```diff
@@ -0,0 +1,239 @@
...
+          </listitem>
+        </varlistentry>
+
+        <varlistentry>
+          <term>
+            <option>fse<replaceable>/home</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+              The device file or mountpoint the policy applies to. Defaults to the filesystem containing the users home directory.
```
t8m: Wrap here please.

modules/pam_setquota/pam_setquota.8.xml
```diff
@@ -0,0 +1,239 @@
...
+            </para>
+          </listitem>
+	</varlistentry>
+      <varlistentry>
+          <term>
+            <option>startuid=<replaceable>1000</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+              Describe the start of the UID range the policy is applied to. (Defaults to 1000)
```
t8m: Could you please take the default from login.defs UID_MIN? We have helper function for that.

modules/pam_setquota/pam_setquota.8.xml
```diff
@@ -0,0 +1,239 @@
...
+            </para>
+          </listitem>
+	</varlistentry>
+	      <varlistentry>
+          <term>
+            <option>enduid=<replaceable>0</replaceable></option>
+          </term>
+          <listitem>
+            <para>
+	      Describe the end of the UID range the policy is applied to. Setting  <emphasis>enduid=0</emphasis> results in an open-ended UID range (i.e. all uids greater than <option>startuid</option> are included). (Defaults to 0)
```
t8m: Please wrap.

modules/pam_setquota/pam_setquota.8.xml
```diff
@@ -0,0 +1,239 @@
...
+        </varlistentry>
+
+      </variablelist>
+    </para>
+  </refsect1>
+
+  <refsect1 id='pam_setquota-examples'>
+    <title>EXAMPLES</title>
+    <para>
+    A single invocation of `pam_setquota` applies a specific policy to a UID range. Applying different policies to specific UID ranges is done by invoking pam_setquota more than once. The last matching entry defines the resulting quota.
```
t8m: Please wrap.

modules/pam_setquota/pam_setquota.8.xml
```diff
@@ -0,0 +1,239 @@
...
+    </para>
+  </refsect1>
+
+  <refsect1 id='pam_setquota-author'>
+    <title>AUTHOR</title>
+      <para>
+        pam_setquota was originally written by Ruslan Savchenko &lt;savrus@mexmat.net&gt;.
+      </para>
+      <para>
+      Further modifications were made by Shane Tzen&lt;shane@ict.usc.edu&gt;, Sven Hartge &lt;sven@svenharte.de&gt;  and Keller Fuchs &lt;kellerfuchs@hashbang.sh&gt;
```
t8m: Please wrap.

modules/pam_setquota/pam_setquota.8.xml
shartge: You mean something like
`param->startuid = pam_modutil_search_key (pamh, LOGIN_DEFS, "UID_MIN");`

shartge: https://github.com/linux-pam/linux-pam/pull/94/commits/292deb0ebcb19edf80369f0e24439e81ffbb890f should address this, unless I misunderstood you.

t8m: Yes, I meant that.

shartge:
> Still one small nit. Also please could you avoid the merge commit? We do not want merge commits in the git.

I have to admit, my git-foo has lost me on that one.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,238 @@
...
+      p->overwrite = strtol(*argv + 10, NULL, 10);
+    else if (strncmp(*argv, "debug=", 6) == 0)
+      p->debug = strtol(*argv + 6, NULL, 10);
+  }
+}
+
+PAM_EXTERN int pam_sm_open_session(pam_handle_t *pamh, int flags, int argc,
+                                   const char **argv) {
+  int retval;
+	char *val;
```
t8m: Indentation seems wrong - tabs/vs spaces?

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  /* Find out what device the filesystem is hosted on */
+  if ((fd = setmntent("/etc/mtab", "r")) == NULL) {
+    pam_syslog(pamh, LOG_ERR, "Unable to open /etc/mtab");
+    return PAM_PERM_DENIED;
+  }
+
+  *mntpoint = *mntdevice = '\0';
+  size_t match_size = 0;
```
t8m: Here is another variable declaration in the middle of a block.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,238 @@
...
+
+static inline void debug(pam_handle_t *pamh, struct if_dqblk *p) {
+  pam_syslog(pamh, LOG_DEBUG, "bsoftlimit=%llu bhardlimit=%llu"
+                              "isoftlimit=%llu ihardlimit=%llu",
+             p->dqb_bsoftlimit, p->dqb_bhardlimit, p->dqb_isoftlimit,
+             p->dqb_ihardlimit);
+}
+
+static bool _pam_parse_dqblk(int argc, const char **argv, struct if_dqblk *p) {
+  bool bhard = false, bsoft = false, ihard = false, isoft = false;
```
t8m: Another modern C feature that we do not use anywhere else - bool type. I am not sure we want to introduce it here. Could you please replace with int and true/false with 1/0?

t8m:
Ping to @ldv-alt and @thkukuk for a second review.

modules/pam_setquota/pam_setquota.c
shartge: https://github.com/linux-pam/linux-pam/pull/94/commits/2b2a3521abdad2ab148ae84ecedef154ca906b7c should fix this.

modules/pam_setquota/pam_setquota.c
shartge: 89187f1 fixes this.

modules/pam_setquota/pam_setquota.c
shartge: When you once use an editor on a different machine... Fixed in 3960007

t8m:
Ping to @ldv-alt and @thkukuk for a second review. I'd like to merge this one.

ldv-alt:
On Tue, Feb 18, 2020 at 02:03:22AM -0800, Tomáš Mráz wrote:
> Ping to @ldv-alt and @thkukuk for a second review. I'd like to merge this one.

The code has several issues but I have no idea how to comment until
all these commits are squashed into a single commit.

t8m:
> On Tue, Feb 18, 2020 at 02:03:22AM -0800, Tomáš Mráz wrote: Ping to @ldv-alt and @thkukuk for a second review. I'd like to merge this one.
> The code has several issues but I have no idea how to comment until all these commits are squashed into a single commit.

@ldv-alt If you do not mind to use the GitHub web interface, in the 'Files changed' tab you can review and comment on all the changes squashed as single patch.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,244 @@
...
+
+PAM_EXTERN int pam_sm_open_session(pam_handle_t *pamh, int flags, int argc,
+                                   const char **argv) {
+  int retval;
+  char *val;
+  const void *user;
+  const struct passwd *pwd;
+  struct pam_params param = { .start_uid = 1000, .end_uid = 0, .fs = NULL };
+  struct if_dqblk ndqblk;
+  FILE *fd;
```
ldv-alt, line 108: Please rename ```fd``` to ```fp```, fd stands for File Descriptor.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,244 @@
...
+  /* Remove the unnecessary '/' from the end of fs parameter */
+  if (param.fs != NULL) {
+    size_t len = strlen(param.fs);
+    if (len > 1)
+      if (param.fs[len - 1] == '/')
+        param.fs[len - 1] = '\0';
+  }
+
+  /* Find out what device the filesystem is hosted on */
+  if ((fd = setmntent("/etc/mtab", "r")) == NULL) {
```
ldv-alt, line 162: The stream returned by ```setmntent``` has to be closed afterwards (using ```endmntent```).

ldv-alt:
On Thu, Mar 05, 2020 at 01:26:04AM -0800, Tomáš Mráz wrote:
> > On Tue, Feb 18, 2020 at 02:03:22AM -0800, Tomáš Mráz wrote: Ping to @ldv-alt and @thkukuk for a second review. I'd like to merge this one.
> > The code has several issues but I have no idea how to comment until all these commits are squashed into a single commit.
> 
> @ldv-alt If you do not mind to use the GitHub web interface, in the 'Files changed' tab you can review and comment on all the changes squashed as single patch.

OK, I've managed to do this, but it was a PITA.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,244 @@
...
+#include <security/pam_modutil.h>
+
+#define LOGIN_DEFS "/etc/login.defs"
+
+#ifndef TRUE
+# define TRUE 1
+#endif
+#ifndef FALSE
+# define FALSE 0
+#endif
```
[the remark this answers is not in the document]
shartge: I was told by @t8m in https://github.com/linux-pam/linux-pam/pull/94#discussion_r367847175 to not use stdbool.h.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,244 @@
...
+      p->dqb_valid |= QIF_ILIMITS;
+      isoft = TRUE;
+    } else if (strncmp(*argv, "btime=", 6) == 0) {
+      p->dqb_btime = strtol(*argv + 6, NULL, 10);
+      p->dqb_valid |= QIF_BTIME;
+    } else if (strncmp(*argv, "itime=", 6) == 0) {
+      p->dqb_itime = strtol(*argv + 6, NULL, 10);
+      p->dqb_valid |= QIF_ITIME;
+    }
+  }
```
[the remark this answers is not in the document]
shartge: Yes, that looks much cleaner using that function. What module do you suggest to look at for a good example?

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,244 @@
...
+
+  /* Find out what device the filesystem is hosted on */
+  if ((fd = setmntent("/etc/mtab", "r")) == NULL) {
+    pam_syslog(pamh, LOG_ERR, "Unable to open /etc/mtab");
+    return PAM_PERM_DENIED;
+  }
+
+  *mntpoint = *mntdevice = '\0';
+  while ((mnt = getmntent(fd)) != NULL) {
```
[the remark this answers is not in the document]
shartge: Can I just use getmntent_r and assume it is available or should this be guarded by `#ifdef HAVE_GETMNTENT_R`?

ldv-alt:
On Wed, Apr 08, 2020 at 03:20:30AM -0700, Sven Hartge wrote:
> Yes, that looks much cleaner using that function. What module do you suggest to look at for a good example?

Since I converted all modules, you can have a look at any module
in the master branch that uses pam_str_skip_prefix.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,244 @@
...
+      /* If fs is not specified use filesystem with homedir as default
+       * Checking the mnt_len-th character in pwd->pw_dir is safe because of the
+       * strncmp(2) check before (whose success implies strlen(pwd->pw_dir) >=
+       * mntlen)
+       */
+      if ((strncmp(pwd->pw_dir, mnt->mnt_dir, mnt_len) == 0) &&
+          (mnt_len > match_size) &&
+          (pwd->pw_dir[mnt_len] == '\0' || pwd->pw_dir[mnt_len] == '/')) {
+        strncpy(mntpoint, mnt->mnt_dir, sizeof(mntpoint));
+        strncpy(mntdevice, mnt->mnt_fsname, sizeof(mntdevice));
```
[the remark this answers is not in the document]
shartge: While it is quite improbable for sizeof(mntdevice) be longer than BUFSIZ, I agree a check would be in order.

What would you suggest? A simple check like
```
if (sizeof(mntdevice) > BUFSIZ) {
    pam_syslog(pamh, LOG_ERR, "Device name %s longer than %lu bytes", mnt->mnt_fsname, BUFSIZ);
    endmntent(fp);
    return PAM_PERM_DENIED;
}
```
should cover that, right?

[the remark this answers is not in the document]
ldv-alt: I suppose a simple ```strdup``` would be a more straightforward way of handling this.

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
ldv-alt: Strictly speaking, ```getmntent_r``` is a GNU extension, although quite an old and popular one.  For example, it was added to musl back in 2011.
@t8m, @thkukuk: can we rely on ```getmntent_r``` being universally available for Linux-PAM?

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,244 @@
...
+  while ((mnt = getmntent(fd)) != NULL) {
+    if (param.fs == NULL) {
+      size_t mnt_len = strlen(mnt->mnt_dir);
+      /* If fs is not specified use filesystem with homedir as default
+       * Checking the mnt_len-th character in pwd->pw_dir is safe because of the
+       * strncmp(2) check before (whose success implies strlen(pwd->pw_dir) >=
+       * mntlen)
+       */
+      if ((strncmp(pwd->pw_dir, mnt->mnt_dir, mnt_len) == 0) &&
+          (mnt_len > match_size) &&
```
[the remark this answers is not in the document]
shartge: I can see why comparing the sizes first is useful, but please explain the problem with strncmp in this case here?

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
shartge: 20bb3b2ba3e3f3d9668f422d2dda249d8036e44c changes the code to use strdup + error checking.

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
t8m: I suppose adding check for getmntent_r to configure a doing the ifdef here couldn't hurt. However there is one more thing to make conditional to build - and that's the whole module. What if quotactl is not supported at all on the target system? Please look at pam_lastlog on how to make the build optional.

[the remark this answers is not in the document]
shartge: > I suppose adding check for getmntent_r to configure a doing the ifdef here couldn't hurt. 

How would the code in pam_setquota look, then? Use `getmntent_r()` via `#ifdef HAVE_GETMNTENT_R` if available and just fall back to `getmntent()` if not and hope for the best?

Or wait until you implemented a linux-pam specific macro?

> However there is one more thing to make conditional to build - and that's the whole module. What if quotactl is not supported at all on the target system? Please look at pam_lastlog on how to make the build optional.

I might need some help with this. autoconf and I never got along well I have to admin, but I will see what I can do.

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
ldv-alt: With ```strncmp```, you have to use ```pwd->pw_dir[mnt_len]```, and you need this lengthy comment explaining why the use of ```pwd->pw_dir[mnt_len]``` is correct.

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
shartge: RIght. I have made the compilation of pam_setquota conditional and it seems to work as it should. "seems" because autoconf is all black magic to me, sorry.

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
shartge: OK, I see.

That code was not originally written by me, so it will take a moment for me to adapt it. Any help is appreciated here.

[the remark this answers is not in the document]
ldv-alt:
```
size_t mnt_len = strlen(mnt->mnt_dir);
const char *s;
if (mnt_len > match_size &&
    (s = pam_str_skip_prefix_len(pwd->pw_dir, mnt->mnt_dir, mnt_len)) != NULL &&
    (s[0] == '\0' || s[0] == '/')) {
```

configure.ac
```diff
@@ -559,6 +559,9 @@ AC_CHECK_HEADERS(fcntl.h limits.h malloc.h sys/file.h sys/ioctl.h sys/time.h sys
 dnl For module/pam_lastlog
 AC_CHECK_HEADERS(lastlog.h utmp.h utmpx.h)
 
+dnl For module/pam_setquota
+AC_CHECK_HEADERS(sys/quota.h linux/quota.h)
+
```
ldv-alt, line 564: You don't need to check these headers if you don't use the result of the check.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+
+static void _pam_parse_params(int argc, const char **argv,
+                              struct pam_params *p) {
+  for (; argc-- > 0; ++argv) {
+    if (strncmp(*argv, "startuid=", 9) == 0)
+      p->start_uid = strtol(*argv + 9, NULL, 10);
+    else if (strncmp(*argv, "enduid=", 7) == 0)
+      p->end_uid = strtol(*argv + 7, NULL, 10);
+    else if (strncmp(*argv, "fs=", 3) == 0)
+      p->fs = (char *)*argv + 3;
```
ldv-alt, line 92: The cast from (const char *) to (char *) looks really bad.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+# define TRUE 1
+#endif
+#ifndef FALSE
+# define FALSE 0
+#endif
+
+struct pam_params {
+  uid_t start_uid;
+  uid_t end_uid;
+  char *fs;
```
ldv-alt, line 40: This should rather be ```const char *fs```.

modules/pam_setquota/pam_setquota.c
shartge: That code is going to be replaced with the new argv parsing using pam_str_skip_prefix anyways, isn't it?

configure.ac
shartge: I see, pam_lastlog has `#ifdef HAVE_UTMP_H` because it can use either utmp.h or lastlog.h.

No such thing is possible for pam_setquota, the queck for quotactl() should be enough then.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+       (pwd->pw_uid > param.end_uid)))
+    return PAM_SUCCESS;
+
+  /* Remove the unnecessary '/' from the end of fs parameter */
+  if (param.fs != NULL) {
+    size_t len = strlen(param.fs);
+    if (len > 1)
+      if (param.fs[len - 1] == '/')
+        param.fs[len - 1] = '\0';
+  }
```
ldv-alt, line 157: It's too late to fix ```param.fs``` at this point.
Please fix it in ```_pam_parse_params```

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+      p->overwrite = strtol(*argv + 10, NULL, 10);
+    else if (strncmp(*argv, "debug=", 6) == 0)
+      p->debug = strtol(*argv + 6, NULL, 10);
+  }
+}
+
+PAM_EXTERN int pam_sm_open_session(pam_handle_t *pamh, int flags, int argc,
+                                   const char **argv) {
+  int retval;
+  char *val, *mntdevice;
```
ldv-alt, line 104: Please initialize ```mntdevice``` with NULL.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+        param.fs[len - 1] = '\0';
+  }
+
+  /* Find out what device the filesystem is hosted on */
+  if ((fp = setmntent("/etc/mtab", "r")) == NULL) {
+    pam_syslog(pamh, LOG_ERR, "Unable to open /etc/mtab");
+    return PAM_PERM_DENIED;
+  }
+
+  *mntdevice = '\0';
```
ldv-alt, line 165: This must be removed now.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+      size_t mnt_len = strlen(mnt->mnt_dir);
+      /* If fs is not specified use filesystem with homedir as default
+       * Checking the mnt_len-th character in pwd->pw_dir is safe because of the
+       * strncmp(2) check before (whose success implies strlen(pwd->pw_dir) >=
+       * mntlen)
+       */
+      if ((strncmp(pwd->pw_dir, mnt->mnt_dir, mnt_len) == 0) &&
+          (mnt_len > match_size) &&
+          (pwd->pw_dir[mnt_len] == '\0' || pwd->pw_dir[mnt_len] == '/')) {
+        if ((mntdevice = strdup(mnt->mnt_fsname)) == NULL) {
```
ldv-alt, line 174: Please ```free(mntdevice)``` first.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+          pam_syslog(pamh, LOG_CRIT,  "Memory allocation error");
+          endmntent(fp);
+          return PAM_PERM_DENIED;
+        }
+        match_size = mnt_len;
+      }
+    } else if ((strcmp(param.fs, mnt->mnt_dir) == 0) ||
+               (strcmp(param.fs, mnt->mnt_fsname) == 0)) {
+        if ((mntdevice = strdup(mnt->mnt_fsname)) == NULL) {
```
ldv-alt, line 183: Please ```free(mntdevice)``` first.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+          pam_syslog(pamh, LOG_CRIT,  "Memory allocation error");
+          endmntent(fp);
+          return PAM_PERM_DENIED;
+        }
+    }
+  }
+
+  if (*mntdevice == '\0') {
```
ldv-alt, line 191: There must be a check of ```mntdevice``` against NULL instead.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  /* Get limits */
+  if (quotactl(QCMD(Q_GETQUOTA, USRQUOTA), mntdevice, pwd->pw_uid,
+               (void *)&ndqblk) == -1) {
+    pam_syslog(pamh, LOG_ERR, "fail to get limits for user %s : %m",
+               pwd->pw_name);
+    endmntent(fp);
+    return PAM_PERM_DENIED;
+  }
```
ldv-alt, line 204: Please don't leak ```mntdevice```.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /* Set limits */
+    if (quotactl(QCMD(Q_SETQUOTA, USRQUOTA), mntdevice, pwd->pw_uid,
+                 (void *)&ndqblk) == -1) {
+      pam_syslog(pamh, LOG_ERR, "fail to set limits for user %s : %m",
+                 pwd->pw_name);
+      endmntent(fp);
+      return PAM_PERM_DENIED;
+    }
```
ldv-alt, line 228: Please don't leak ```mntdevice```.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,252 @@
...
+    if (param.debug == 1)
+      debug(pamh, &ndqblk);
+  }
+  endmntent(fp);
+  return PAM_SUCCESS;
```
ldv-alt, line 233: Please don't leak ```mntdevice```.

modules/pam_setquota/pam_setquota.c
ldv-alt: Yes, but the problem with the trailing slash will have to be dealt with anyway.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,244 @@
...
+
+  /* Get UID_MIN for default start_uid from login.defs */
+  val = pam_modutil_search_key(pamh, LOGIN_DEFS, "UID_MIN");
+
+  /* Should UID_MIN be undefined, use 1000 as safe starting UID
+   * to avoid setting a quota for root and system users if startuid= parameter
+   * is absent.
+   */
+  if (val) {
+    param.start_uid = strtol(val, NULL, 10);
```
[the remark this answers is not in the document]
shartge: Neat. I would have expected such a function to be available via pam_modutil, to be honest.

I adapted the code from pam_usertype.c to pam_setquota.c.

modules/pam_setquota/pam_setquota.c
shartge: Pardon my ignorance, but if mntdevice has been initialized to NULL and not used before, why would I need to free() it before using strdup();

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
shartge: Thank you, I updated the code to your example.

modules/pam_setquota/pam_setquota.c
shartge: Right. Which is quite possibly the reason the original author made it char instead of const char. To be able to manipulate the string later on.

modules/pam_setquota/pam_setquota.c
ldv-alt: What if it was assigned before?
```free(NULL)``` is a no-op.

Comment (shartge) [reply modules/pam_setquota/pam_setquota.c:194]: I see my comprehension problem here. I totally blanked out that we are in a loop here, so of course mntdevice could have been strdup'ed before.
Again, thank you for being patient with me. My normal coding environment is Perl and Python, my C-foo is a bit rusty after 15 years not using it.

File: modules/pam_setquota/pam_setquota.c [hunk 31]
```diff
@@ -0,0 +1,273 @@
...
+        param.fs[len - 1] = '\0';
+  }
+
+  /* Find out what device the filesystem is hosted on */
+  if ((fp = setmntent("/etc/mtab", "r")) == NULL) {
+    pam_syslog(pamh, LOG_ERR, "Unable to open /etc/mtab");
+    return PAM_PERM_DENIED;
+  }
+
+  while ((mnt = getmntent(fp)) != NULL) {
```
ldv-alt, line 186: I suppose you can call ```endmntent``` right after this getmntent loop.

shartge: I suppose so, too. Done.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,273 @@
...
+  if ((ndqblk.dqb_bsoftlimit == 0 && ndqblk.dqb_bhardlimit == 0 &&
+       ndqblk.dqb_isoftlimit == 0 && ndqblk.dqb_ihardlimit == 0) ||
+      param.overwrite == 1) {
+
+    /* Parse new limits */
+    ndqblk.dqb_valid = 0;
+    if (!_pam_parse_dqblk(argc, argv, &ndqblk)) {
+      pam_syslog(pamh, LOG_AUTH | LOG_WARNING, "setting soft and hard limits"
+                                               "separately");
+    }
```
ldv-alt, line 242: I'm not sure what does this warning actually mean.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,273 @@
...
+
+    /* Set limits */
+    if (quotactl(QCMD(Q_SETQUOTA, USRQUOTA), mntdevice, pwd->pw_uid,
+                 (void *)&ndqblk) == -1) {
```
ldv-alt, line 246: Is there any reason to call setquota if ndqblk.dqb_valid == 0?

shartge: You mean if `ndqblk.dqb_valid` remains at 0 after  `_pam_parse_dqblk()`?

ldv-alt: Yes.

shartge: Yes, one could check for `ndqblk.dqb_valid != 0` but for this to happen you would need to configure pam_setquota without any limits and why would anyone do this?

ldv-alt: By mistake?  This would justify issuing a warning.

shartge: Fair enough. Something like:

```
    /* No limits defined at all in configuration?*/
    if (ndqblk.dqb_valid == 0) {
      pam_syslog(pamh, LOG_AUTH | LOG_WARNING, "no limits defined in configuration");
    } 
```  
I am unsure if the module should exit in that `if` and if, then with what exit code, or just let the code run along the normal path and exit at the end.

modules/pam_setquota/pam_setquota.c
shartge: I added a comment lining out the possible problem that could occur.

shartge: Nothing in the documentation hints at being able to do the following:

```
session required        pam_setquota.so bhardlimit=568000 fs=/home 
session required        pam_setquota.so bsoftlimit=512000 fs=/home 
```
It will work but only if you do it in that exact order, because the softlimit *has* to be lower than the hardlimit. And because it will work (at first) but you are setting yourself up for trouble later, the code warns about it.

Also imagine having the above setup in that way and now you decide to change things:

```
session required        pam_setquota.so bhardlimit=268000 fs=/home overwrite=1
session required        pam_setquota.so bsoftlimit=212000 fs=/home overwrite=1
```
Now it will fail for every user with an existing quota, because now the new hardlimit again is lower than the old softlimit.

Normally no user would do this, because the documentation never really shows this possibility.

shartge: I have reworded and hopefully improved the warning logged when this configuration is encountered.
Because such a configuration technically works, I don't think it is necessary to exit with an error.

modules/pam_setquota/pam_setquota.c
shartge: I have moved the /-removal function into _pam_parse_params, but because I have to change the fs parameter if it has a trailing /, I need to cast something somewhere. My solution so far is at https://github.com/linux-pam/linux-pam/pull/94/files#diff-a68500e7c99fcd33f0d387c65f0257f7R100 but I am open to anything more sophisticated.

shartge: Nix that, casting to const char* is not necessary, only the other way around.

modules/pam_setquota/pam_setquota.c
```diff
@@ -164,15 +165,12 @@ PAM_EXTERN int pam_sm_open_session(pam_handle_t *pamh, int flags, int argc,
...
-       * strncmp(2) check before (whose success implies strlen(pwd->pw_dir) >=
-       * mntlen)
-       */
-      if ((strncmp(pwd->pw_dir, mnt->mnt_dir, mnt_len) == 0) &&
-          (mnt_len > match_size) &&
-          (pwd->pw_dir[mnt_len] == '\0' || pwd->pw_dir[mnt_len] == '/')) {
+      const char *s;
+      if (mnt_len > match_size &&
+         (s = pam_str_skip_prefix_len(pwd->pw_dir, mnt->mnt_dir, mnt_len)) != NULL &&
+         (s[0] == '\0' || s[0] == '/')) {
```
shartge, line 225: I need to go back on parts of that change, because, as it turns out, the code was incorrect from the start.
I need to look at `pwd->pw_dir[mnt_len-1]` to check for `\0` or `/` and this seems not possible with `pam_str_skip_prefix_len()`, because it would technically be `s[-1]`.

ldv-alt: Why? You need to check the symbol behind the matched prefix, don't you?

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,317 @@
...
+      char *temp = NULL;
+      size_t len = strlen(str);
+      temp = strdup(str);
+      if (len > 1) {
+        if (temp[len - 1] == '/') {
+          temp[len - 1] = '\0';
+        }
+      }
+      p->fs = strdup(temp);
+      free(temp);
```
ldv-alt, line 133: I don't see why you need so many copies of the string.
I'm not sure whether you really need to make a copy, but if you do, please do not forget to check return code and free the copy afterwards.

modules/pam_setquota/pam_setquota.c
shartge: No. I've been debugging this for half of the day.
It needs to find the longest matching mountpoint containing pwd->pw_dir.
The above code never triggers, even if /home/hartge is on /.

shartge: Or maybe, and I just considered this now, *s needs to be initialized to NULL?

modules/pam_setquota/pam_setquota.c
shartge: This is me fumbling with strings in C, assistance is appreciated.
str is const char here, so I can't remove the / from the end, right? So I need a temp variable to do the changes and then strdup it into p->fs, because just p->fs = temp will cause p->fs to contain garbage after free(temp).

modules/pam_setquota/pam_setquota.c
shartge: > No. I've been debugging this for half of the day.
> It needs to find the longest matching mountpoint containing pwd->pw_dir.
> The above code never triggers, even if /home/hartge is on /.

Inspecting s for "/home/hartge and "/" and "1" as input results in s being "home/hartge".
But I need to check the last byte that matched and not the one after, so s contains the wrong stuff here.

shartge: To add to my shame: That part of the code was broken for a long time and I never noticed it, because I always used it with explicit fs=... parameters and never fell back on the automatic home-directory search.

ldv-alt: You just have to decrease mnt_len for every trailing "/" in mnt->mnt_dir, that's all.

shartge: Wouldn't that break in mnt_dir='/' and mnt_len then being 0?

ldv-alt: something like this:
```size_t mnt_len;
for (mnt_len = strlen(mnt->mnt_dir); mnt_len > 0; --mnt_len)
  if (mnt->mnt_dir[mnt_len - 1] != '/')
    break;```
This way ```mnt_len``` might become 0 so ```mnt_len > match_size``` condition has to be dealt with somehow.

Comment (ldv-alt) [reply modules/pam_setquota/pam_setquota.c:225]: > > No. I've been debugging this for half of the day.
> > It needs to find the longest matching mountpoint containing pwd->pw_dir.
> > The above code never triggers, even if /home/hartge is on /.
> 
> Inspecting s for "/home/hartge and "/" and "1" as input results in s being "home/hartge".
> But I need to check the last byte that matched and not the one after, so s contains the wrong stuff here.

Please repeat the same thought experiment with "/home/hartge" and "/home".

Comment (shartge) [reply modules/pam_setquota/pam_setquota.c:225]: Yes, I am adding a bunch of logging code to the function to throw onto my test VM.
I just noticed there is something else strange here. I need to meditate over this for a night.

File: modules/pam_setquota/pam_setquota.c [hunk 36]
```diff
@@ -0,0 +1,317 @@
...
+     * when the quota for a user is reduced and the old soft limit is
+     * higher than the new hard limit.
+     * Setting both softlimit and hardlimit at the same time avoids that
+     * problem.
+     */
+    ndqblk.dqb_valid = 0;
+    if (!_pam_parse_dqblk(argc, argv, &ndqblk)) {
+      pam_syslog(pamh, LOG_AUTH | LOG_WARNING,
+        "Do not set soft and hard limits separately, undefined behavior might occur!");
+    }
```
ldv-alt, line 281: Why do we want to allow setting a hard limit without a soft limit, and visa versa?
Is there any potential usefulness in adding support for that?
There must be a reason for enabling this kind of user interface fragility in a new module.

modules/pam_setquota/pam_setquota.c
shartge: From looking at mnt_dir, pw_dir and the resulting s I think I only need to special case mnt_dir='/', because it is the only mounted filesystem with a trailing /.
In every other case, the remaining s will always either start with a / or be empty, in case of a complete match.

ldv-alt: Yes.

modules/pam_setquota/pam_setquota.c
ldv-alt: The only place where you use param.fs as a string is
```    } else if ((strcmp(param.fs, mnt->mnt_dir) == 0) ||
               (strcmp(param.fs, mnt->mnt_fsname) == 0)) {
```

If you could rewrite it without strcmp, then you could save e.g. ```param.fs``` and ```param.fs_len``` and use them instead.

modules/pam_setquota/pam_setquota.c
shartge: While I could think of some possible edge cases to just set the hardlimit and leave the softlimit alone, I have to agree that I don't really see any useful application here.
I will change the code the ouput an error should only one of the limits exit in the configuration.

modules/pam_setquota/pam_setquota.c
shartge: Sorry, either it is too late or I have reached the limit of my understanding of C.

shartge: Hmm, something like this?
```
struct pam_params {
   uid_t start_uid;
   uid_t end_uid;
   const char *fs;
  size_t fs_len;
   int overwrite;
   int debug;
 };
```
and then

```
} else if ((str = pam_str_skip_prefix(*argv, "fs=")) != NULL) {
      p->fs = str;
      p->fs_len = strlen(str);
      /* Mask the unnecessary '/' from the end of fs parameter */
      if (p->fs_len > 1 && p->fs[len - 1] == '/')
        --p->fs_len;
}
```
and finally

```
    } else if ((strncmp(param.fs, mnt->mnt_dir, param.fs_len) == 0) ||
               (strncmp(param.fs, mnt->mnt_fsname, param.fs_len) == 0)) {
```
This is all I can come up right now at this time.

ldv-alt: > ```
>     } else if ((strncmp(param.fs, mnt->mnt_dir, param.fs_len) == 0) ||
>                (strncmp(param.fs, mnt->mnt_fsname, param.fs_len) == 0)) {
> ```

```
} else if ((strncmp(param.fs, mnt->mnt_dir, param.fs_len) == 0
                    && mnt->mnt_dir[param.fs_len] == '\0') ||
           (strncmp(param.fs, mnt->mnt_fsname, param.fs_len) == 0
                    && mnt->mnt_fsname[param.fs_len] == '\0')) {
```

shartge: Interesting, yes.
Checking for the string terminator to check if the match would be exact on the 2nd paramater to strncmp was something not obvious to me.

modules/pam_setquota/pam_setquota.c
shartge: Ok, I got it working as it should now, the if() just needed the special case for mnt_len == 0 and that solved the problem of not finding pw_dir if it was contained in /.
Now it looks like this (I've snipped some uninteresting lines for brevity):
1) User "oweh" with $HOME in /home/oweh contained directly in /:
```
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/sys"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/proc"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/dev"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/dev/pts"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/run"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Found pw_dir="/home/oweh" in mnt_dir="/" with suffix="/home/oweh" on device="/dev/sda1"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/sys/kernel/security"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/dev/shm"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/run/lock"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/sys/fs/cgroup"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/sys/fs/cgroup/unified"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/home2"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/home3"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): Trying to match pw_dir="/home/oweh" with mnt_dir="/run/user/1000"
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): device=/dev/sda1 bsoftlimit=512000 bhardlimit=563200 isoftlimit=0 ihardlimit=0
Apr 10 15:11:04 debian sshd[14343]: pam_setquota(sshd:session): device=/dev/sda1 bsoftlimit=512000 bhardlimit=563200 isoftlimit=0 ihardlimit=0
```
2) User "tester" with $HOME in "/home2" on the fs "/home2":
```
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/sys"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/proc"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/dev"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/dev/pts"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/run"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Found pw_dir="/home2" in mnt_dir="/" with suffix="/home2" on device="/dev/sda1"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/sys/kernel/security"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/dev/shm"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/run/lock"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/sys/kernel/debug"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/run/user/117"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/run/user/0"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/home2"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Found pw_dir="/home2" in mnt_dir="/home2" with suffix="" on device="/dev/loop0"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/home3"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): Trying to match pw_dir="/home2" with mnt_dir="/run/user/1001"
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): device=/dev/loop0 bsoftlimit=512000 bhardlimit=563200 isoftlimit=0 ihardlimit=0
Apr 10 15:11:43 debian sshd[14377]: pam_setquota(sshd:session): device=/dev/loop0 bsoftlimit=512000 bhardlimit=563200 isoftlimit=0 ihardlimit=0
```
3) And finally user "test3" with $HOME as "/home3/test3" in fs "/home3":
```
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/sys"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/proc"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/dev"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/dev/pts"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/run"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Found pw_dir="/home3/test3" in mnt_dir="/" with suffix="/home3/test3" on device="/dev/sda1"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/sys/kernel/security"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/dev/shm"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/run/lock"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/sys/fs/cgroup"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/sys/kernel/debug"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/run/user/117"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/run/user/0"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/home2"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/home3"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Found pw_dir="/home3/test3" in mnt_dir="/home3" with suffix="/test3" on device="/dev/loop1"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/run/user/1001"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): Trying to match pw_dir="/home3/test3" with mnt_dir="/run/user/1002"
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): device=/dev/loop1 bsoftlimit=512000 bhardlimit=563200 isoftlimit=0 ihardlimit=0
Apr 10 15:12:18 debian sshd[14407]: pam_setquota(sshd:session): device=/dev/loop1 bsoftlimit=512000 bhardlimit=563200 isoftlimit=0 ihardlimit=0
```

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
shartge: I don't mind using stdbool.h again, just tell me which one you prefer.

modules/pam_setquota/pam_setquota.c
shartge: After sleeping over this: I don't think the module should throw an error here. Sure, this is unusual, but since no harm is done and the module then is just a NO-OP, it should just warn and then exit normally.

modules/pam_setquota/pam_setquota.c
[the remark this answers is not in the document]
shartge: > Strictly speaking, `getmntent_r` is a GNU extension, although quite an old and popular one. For example, it was added to musl back in 2011.
> @t8m, @thkukuk: can we rely on `getmntent_r` being universally available for Linux-PAM?

I added conditional support for `getment_r` to the code and modified `configure.ac` to check for it. The resulting #ifdef'ery is a bit ugly, but there is only so much you can do in this case.

ldv-alt:
On Fri, Apr 10, 2020 at 06:54:03AM -0700, Sven Hartge wrote:
> @shartge commented on this pull request.
> 
> > +    /* Set limits */
> +    if (quotactl(QCMD(Q_SETQUOTA, USRQUOTA), mntdevice, pwd->pw_uid,
> +                 (void *)&ndqblk) == -1) {
> 
> After sleeping over this: I don't think the module should throw an error here. Sure, this is unusual, but since no harm is done and the module then is just a NO-OP, it should just warn and then exit normally.

If Q_SETQUOTA failed, it means that configured restrictions haven't been applied.
If the module failed to do the only thing that was requested from it,
it should return an error code.
Those who want to ignore the error can always include the module as optional.

shartge:
> If Q_SETQUOTA failed, it means that configured restrictions haven't been applied. If the module failed to do the only thing that was requested from it, it should return an error code. Those who want to ignore the error can always include the module as optional.

But Q_SETQUOTA won't fail here, because the bitmask ndqblk.dqb_valid == 0. In that case no limits will have been attempted to be set, if no limits have been configured by the admin (for whatever reasons).

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,346 @@
...
+      isoft = TRUE;
+    } else if ((str = pam_str_skip_prefix(*argv, "btime=")) != NULL) {
+      p->dqb_btime = strtol(str, &ep, 10);
+      if (!ep || (p->dqb_btime < 0))
+        p->dqb_btime = 0;
+      p->dqb_valid |= QIF_BTIME;
+    } else if ((str = pam_str_skip_prefix(*argv, "itime=")) != NULL) {
+      p->dqb_itime = strtol(str, &ep, 10);
+      if (!ep || (p->dqb_itime < 0))
+        p->dqb_itime = 0;
```
ldv-alt, line 98: All these members of ```struct if_dqblk``` have an unsigned 64-bit type, so they always >= 0.
You probably want to use ```strtoll``` instead of ```strtol``` and assign its result to a temporary variable of type ```long long```.
I also suggest to create a helper function taking a ```const char *``` argument and returning an ```unsigned long long``` result, this would make the parser shorter and simpler.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,346 @@
...
+        pam_syslog(pamh, LOG_DEBUG, "Trying to match pw_dir=\"%s\" with mnt_dir=\"%s\"", pwd->pw_dir, mnt->mnt_dir);
+      /*
+       * (mnt_len > match_size) Only try matching the mnt_dir if its length
+       * is longer than the last matched length, trying to find the longest
+       * mnt_dir for a given pwd_dir.
+       *
+       * (mnt_len == 0) special-cases the root-dir /, which is the only
+       * mnt_dir with a trailing '/', which got masked earlier.
+       */
+      if ((mnt_len > match_size || mnt_len == 0) &&
```
ldv-alt, line 245: I'd rather be on the safe side:
```
if ((mnt_len > match_size || (mnt_len == 0 && mnt->mnt_dir[0] == '/')) &&
```

modules/pam_setquota/pam_setquota.c
shartge: Module internal helper functions names should start with "_", correct?

ldv-alt: > Module internal helper functions names should start with "_", correct?

No, but they have to be static.

ldv-alt:
On Fri, Apr 10, 2020 at 03:37:15PM -0700, Sven Hartge wrote:
> > If Q_SETQUOTA failed, it means that configured restrictions haven't been applied. If the module failed to do the only thing that was requested from it, it should return an error code. Those who want to ignore the error can always include the module as optional.
> 
> But Q_SETQUOTA won't fail here, because the bitmask ndqblk.dqb_valid == 0. In that case no limits will have been attempted to be set, if no limits have been configured by the admin (for whatever reasons).

What's the reason of invoking Q_SETQUOTA with dqb_valid == 0?  I'm confused.

shartge:
> What's the reason of invoking Q_SETQUOTA with dqb_valid == 0? I'm confused.

That is the philosophical question here.
One could just exit directly when `ndqblk.dqb_valid == 0` with `PAM_SUCCESS` (because the module succeeded in doing nothing, as it was asked to do) in that `if` or just run along and hit the `return PAM_SUCCESS` at the end.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,346 @@
...
+      pam_syslog(pamh, LOG_ERR,
+        "Both soft- and hardlimits for %s need to be configured at the same time!", mntdevice);
+      free(mntdevice);
+      return PAM_PERM_DENIED;
+    }
+
+    /* Nothing changed? Are no limits defined at all in configuration? */
+    if (ndqblk.dqb_valid == 0) {
+      pam_syslog(pamh, LOG_AUTH | LOG_WARNING, "no limits defined in configuration");
+    }
```
ldv-alt, line 315: Should we rather return PAM_IGNORE in this case?

modules/pam_setquota/pam_setquota.c
shartge: So:
```
static unsigned long long
_pam_setquota_convert_num(const char *str) {
  long long temp;
  char *ep = NULL;

  temp = strtoll(str, &ep, 10);
  if (!ep || (temp < 0)) 
    return 0
  else
    return temp;
}
```
`return` should to an automatic cast, if I am not wrong and since because of the preceding `if` `temp` has to be >= 0, this is safe.

modules/pam_setquota/pam_setquota.c
shartge: Yes, that would be a better matching value.

modules/pam_setquota/pam_setquota.c
shartge: Agreed.

modules/pam_setquota/pam_setquota.c
shartge: > No, but they have to be static.
Yes, of course. But since `_pam_parse_dqblk` already exists, I am keeping the naming somewhat consistent.

ldv-alt: No leading underscores, please, this is a reserved namespace.
You don't need pam_setquota prefix either.
convert_num looks too generic, maybe some mentioning of dqb won't harm.
I'd suggest something in lines of ```str_to_dqb_num```.
You can just write
```
long long num = strtoll(str, &ep, 10);
```
P.S. there has to be a semicolon after ```return 0```.

shartge: I will rename the other functions then as well.

ldv-alt: > > No, but they have to be static.
> 
> Yes, of course. But since `_pam_parse_dqblk` and `_pam_parse_params` already exist, I am keeping the naming somewhat consistent.

They shouldn't exist either, please rename them, simple ```parse_dqblk``` and ```parse_params``` would do.

shartge: Done.

shartge:
Huh?
That latest CI failure is an CI internal error, right? Because the code compiled and ran fine locally.

shartge:
I think I have solved all outstanding issues so far with the only one left being the stdbool one which I can't resolve on my own because this depends on input  from @t8m and @ldv-alt.

t8m:
I am not blocking the approval on stdbool. So I am fine with it either way.

shartge:
I have changed to code to use stdboo.h again and also updated the man-page to reflect the currently used return codes.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,387 @@
...
+   Copyright © 2016 Keller Fuchs <kellerfuchs@hashbang.sh>
+
+*/
+
+#include <sys/types.h>
+#include <sys/quota.h>
+#include <linux/quota.h>
+#include <pwd.h>
+#include <syslog.h>
+#include <ctype.h>
```
ldv-alt, line 17: ctype.h used to be needed for isspace, but it'ss currently unused, see below.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,387 @@
...
+  uid_t start_uid;
+  uid_t end_uid;
+  const char *fs;
+  size_t fs_len;
+  int overwrite;
+  int debug;
+};
+
+static inline void
+debug(pam_handle_t *pamh, struct if_dqblk *p,
```
ldv-alt, line 47: ```struct if_dqblk *``` is accessed read-only, please add ```const``` qualifier.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,387 @@
...
+}
+
+static unsigned long long
+str_to_dqb_num(const char *str) {
+  char *ep = NULL;
+
+  errno = 0;
+  long long temp = strtoll(str, &ep, 10);
+  if (!ep || errno != 0 || (temp < 0))
+    return 0;
```
ldv-alt, line 63: This implements roughly the following logic: if the given string is not a number from our PoV, ignore it by treating it as 0 (which means the corresponding quota limit is disabled).  Should we issue some diagnostics in this case?

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,387 @@
...
+        || (errno != 0 && ul == 0)
+        || value == ep
+        || *ep != '\0')) {
+        uid = (uid_t)ul;
+    } else {
+        uid = default_value;
+    }
+
+    return uid;
+}
```
ldv-alt, line 126: ```pam_usertype_get_id``` also strips trailing whitespace while this function doesn't.  Is this intended?
As this function is also used to parse command line arguments, should we issue some diagnostics in case the value is ignored because it doesn't represent a valid number?

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,387 @@
...
+        if (param.debug >= 2)
+          pam_syslog(pamh, LOG_DEBUG, "Found fs=\"%s\" in mnt_dir=\"%s\" "
+                     "on device=\"%s\"", param.fs, mnt->mnt_dir, mntdevice);
+    }
+  }
+
+  endmntent(fp);
+
+  if (mntdevice == NULL) {
+    pam_syslog(pamh, LOG_ERR, "Filesystem not found");
```
ldv-alt, line 297: I think it might be useful to log what exactly was not found.

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,387 @@
...
+                 pwd->pw_name, mntdevice);
+    }
+    /* End module */
+    free(mntdevice);
+    return PAM_IGNORE;
+  }
+
+  /* End module, should never be reached, just a safeguard */
+  free(mntdevice);
+  return PAM_SUCCESS;
```
ldv-alt, line 370: Sorry, I don't see how this safeguard could help.
Of course the compiler is perfectly capable of dead code elimination, but I'd rather see the compiler complaining that control reaches the end of non-void function if it ever happens in the future.

modules/pam_setquota/pam_setquota.c
shartge: I don't think `str` can contain any whitespace in this application here, but I will look into it.

modules/pam_setquota/pam_setquota.c
shartge: You mean differentiating between fs=/whatever and the $HOME autodetect case?

modules/pam_setquota/pam_setquota.c
shartge: OK, this is quite obviously my out-of-date compile knowledge here, because way way back (stupid) code like this was needed to shut the compiler up complaining about missing a `return` on a non-void function because it was too stupid to know that the preceding `if` or `case` could never not lead to a `return`.
If you say this is no longer the case then I am more than happy to drop that final piece of code.

modules/pam_setquota/pam_setquota.c
shartge: We could. But that would mean dragging `pamh` through both `parse_dqblk` and `str_to_dqb_num` to be able to use `pam_syslog()`, wouldn't it?

modules/pam_setquota/pam_setquota.c
shartge: The way I read the code for `pam_modutil_search_key()` it already removes tabs and spaces, same with the strings in `argv`. So if I am not wrong, no string presented to this function should contain leading or trailing whitespaces.

ldv-alt: Apparently, ```pam_modutil_search_key``` already performs trailing whitespace removal,
so no need to duplicate that work here.

modules/pam_setquota/pam_setquota.c
ldv-alt: We could just print the path we were looking for, i.e. either param.fs or pwd->pw_dir.

modules/pam_setquota/pam_setquota.c
ldv-alt: Yes, feel free to drop that.

modules/pam_setquota/pam_setquota.c
ldv-alt: Yes, it would.

modules/pam_setquota/pam_setquota.c
shartge: Something like `pam_syslog(pamh, LOG_ERR, "Filesystem not found: %s", (param.fs)?param.fs:pwd->pw_dir);`
Would that work? I always found the `?:` operator confusing.

modules/pam_setquota/pam_setquota.c
shartge: I will do the same for `str_to_uid()` then.

modules/pam_setquota/pam_setquota.c
ldv-alt: Yes, ```param.fs ? param.fs : pwd->pw_dir``` should work.

shartge: Done.

modules/pam_setquota/pam_setquota.c
shartge: Code happily removed.

modules/pam_setquota/pam_setquota.c
shartge: I have added some more output on parse errors, according to Linux-PAM_MWG at the LOG_ERR level.

shartge:
Testing the changes other than compiling them will have to wait for tomorrow/later this day.

shartge:
Right. Testing the debug output revealed the error condition checking was in part wrong. I have no idea what my idea was by having it the way it was, but it was clearly incorrect. It is fixed now and works the way it is supposed to be:
```
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Parameter "/etc/login.defs:UID_MIN=-1000" invalid, setting to 1000
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Parameter "startuid=test" invalid, setting to 1000
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Parameter "enduid=test" invalid, setting to 0
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Config: startuid=1000 enduid=0 fs=(null) debug=1 overwrite=1
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Quota read: device=/dev/sda1 bsoftlimit=0 bhardlimit=0 isoftlimit=0 ihardlimit=0 btime=0 itime=0
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Parameter "bsoftlimit=122test" invalid, setting to 0
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Parameter "bhardlimit=123test" invalid, setting to 0
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Parameter "isoftlimit=300test" invalid, setting to 0
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Parameter "ihardlimit=-400" invalid, setting to 0
Apr 15 14:38:02 debian sshd[24221]: pam_setquota(sshd:session): Quota set: device=/dev/sda1 bsoftlimit=0 bhardlimit=0 isoftlimit=0 ihardlimit=0 btime=0 itime=0
```

modules/pam_setquota/pam_setquota.c
```diff
@@ -0,0 +1,395 @@
...
+             p->dqb_btime, p->dqb_itime);
+}
+
+static unsigned long long
+str_to_dqb_num(pam_handle_t *pamh, const char *str, const char *param) {
+  char *ep = NULL;
+
+  errno = 0;
+  long long temp = strtoll(str, &ep, 10);
+  if (*ep != '\0' || str == ep || errno != 0 || (temp < 0)) {
```
ldv-alt, line 62: Could you reorder these checks a bit, e.g.
```temp < 0 || str == ep || *ep != '\0' || errno != 0```

File: modules/pam_setquota/pam_setquota.c [hunk 47]
```diff
@@ -0,0 +1,395 @@
...
+   * Must come after pam_modutil_search_key so that the current system
+   * default for UID_MIN is already in p.start_uid to serve as default
+   * for str_to_uid in case of a parse error.
+   * */
+  parse_params(pamh, argc, argv, &param);
+
+  if (param.debug >= 1)
+    pam_syslog(pamh, LOG_DEBUG, "Config: startuid=%u enduid=%u fs=%s "
+                    "debug=%d overwrite=%d",
+                    param.start_uid, param.end_uid, param.fs,
```
ldv-alt, line 215: Passing null pointer as a string is unsafe, please do something about it, e.g.
```param.fs ? param.fs : ""```

File: modules/pam_setquota/pam_setquota.c [hunk 48]
```diff
@@ -0,0 +1,395 @@
...
+       * (mnt_len == 0 && mnt->mnt_dir[0] == '/') special-cases the
+       * root-dir /, which is the only mnt_dir with a trailing '/', which
+       * got masked earlier.
+       */
+      if ((mnt_len > match_size || (mnt_len == 0 && mnt->mnt_dir[0] == '/')) &&
+         (s = pam_str_skip_prefix_len(pwd->pw_dir, mnt->mnt_dir, mnt_len)) != NULL &&
+         (s[0] == '\0' || s[0] == '/')) {
+        free(mntdevice);
+        if ((mntdevice = strdup(mnt->mnt_fsname)) == NULL) {
+          pam_syslog(pamh, LOG_CRIT,  "Memory allocation error");
```
ldv-alt, line 278: There are two consecutive spaces between 2nd and 3rd arguments.

modules/pam_setquota/pam_setquota.c
shartge: Sure.
Just for me to learn: Reasoning? Optimize checks from most to least probable?

modules/pam_setquota/pam_setquota.c
shartge: Will fix, there is another one of this because of copy'n'waste.

modules/pam_setquota/pam_setquota.c
shartge: Yes.

modules/pam_setquota/pam_setquota.c
ldv-alt: It's hard to say which are less probable, so let's do cheaper checks first.
I presume that pointer dereferences are slightly more expensive.

modules/pam_setquota/pam_setquota.c
shartge: I have also checked the other strings to see if they can be NULL and have not found any.
The others are all behind some kind of if() that checks if the string is NULL before.

modules/pam_setquota/pam_setquota.c
shartge: In the end everything is ORed, so all checks have to be done either way, right? So order shouldn't matter that much, compared to an ANDed if.

shartge: Well, no.
As soon as the first one is true, the if stops.
Just as when doing &&, the first false cause it to stop.

shartge:
As requested before, I have squashed all commit into one to make merging less messy.

ldv-alt:
Could you amend the commit message to start with something like
```pam_setquota: new module to set or modify disk quotas on session start```
and end with
```Resolves: https://github.com/linux-pam/linux-pam/issues/92```
?

shartge:
I have amended the commit message in git and also changed the PR message here.
The CI failures are because of some breakage in the 32bit gcc9 environments.

ldv-alt:
> The CI failures are because of some breakage in the 32bit gcc9 environments.

They must have fixed these gcc9 installation issues, I've re-run the tests, all are green now.

ldv-alt:
Does anybody else have any comments about this module?

shartge:
Side comment: I also did a test-build with master+setquota and gcc10 yesterday to test if adding it introduces any new errors and can report that master+setquota does indeed compile using gcc10 without errors.
(There are some new warnings in other unrelated modules though.)

t8m:
No comments from me. Dmitry, please merge, if you're OK with the module.

ldv-alt:
Merged, thanks!

shartge:
No, thank *you* for providing me the opportunity to contribute.
(It also helps me because I no longer have to maintain the module on my own.)

PR #95, opened 2019-01-24
Changed variable salt to hash

hulto:
helper_verify_password's variable salt is not just the salt but the whole hash. Renamed for clarity and conformity with the rest of the code.

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -1080,21 +1080,21 @@ int
 helper_verify_password(const char *name, const char *p, int nullok)
 {
 	struct passwd *pwd = NULL;
-	char *salt = NULL;
+	char *hash = NULL;
 	int retval;
 
-	retval = get_pwd_hash(name, &pwd, &salt);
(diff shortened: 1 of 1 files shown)


ldv-alt:
On Wed, Jan 15, 2020 at 09:18:01AM -0800, Tomáš Mráz wrote:
> t8m approved this pull request.

Sorry but I find the commit message unreadable.
I suggest we don't accept badly formatted commit messages into the repository.

ldv-alt:
On Mon, Jan 20, 2020 at 02:53:44AM -0800, Tomáš Mráz wrote:
> Unfortunately I did not find anything wrong with the commit message. Yes, I understand it could be a little bit more precise such as "Renamed a variable salt to hash for clarity" or so. However the additional commit text clarifies it. So in general, yes I agree that we should not accept badly formatted commit messages but this particular one did not seem to me to be such.

Well, it's so long that it doesn't fit into my screen,
so despite being clear, it's technically unreadable.

t8m:
Ah, so it is about the full log text not being wrapped properly. I am sorry for not noticing that, I'll make sure to check next time.

PR #96, opened 2019-01-25
Add missing va_end() call

jmoellers:
According to the man pages, "Each invocation of va_start() must be matched by a corresponding invocation of va_end() in the same function."


t8m:
Can you please clean up the pull request so it includes only the change that is in the summary. Especially no merge commits should be there.

PR #97, opened 2019-02-12
Deprecate pam_unix' explicit "usergroups" option in favor of /etc/login.defs

vorlonofportland:
/etc/login.defs has a USERGROUP_ENAB option with the same semantics as the
usergroup module option.  If our definition of the umask comes from
/etc/login.defs, we should read this setting as well.

This restores compatibility with the pre-PAM behaviour of login.

See https://blueprints.launchpad.net/ubuntu/+spec/umask-to-0002.

Signed-off-by: Martin Pitt <martin.pitt@ubuntu.com>
Signed-off-by: Steve Langasek <vorlon@debian.org>

Bug-Debian: http://bugs.debian.org/583958

This patch has been carried in Ubuntu for a number of years and belongs upstream.  I'm happy to discuss whether deprecation of the module option in favor of /etc/login.defs is the right thing to do (IMHO it is).

### DIFF
--- a/modules/pam_umask/pam_umask.8.xml
+++ b/modules/pam_umask/pam_umask.8.xml
@@ -58,7 +58,8 @@
         </listitem>
         <listitem>
           <para>
-            UMASK entry from /etc/login.defs
+            UMASK entry from /etc/login.defs (influenced by USERGROUPS_ENAB in
+	    /etc/login.defs)
           </para>
         </listitem>
         <listitem>
@@ -115,6 +116,11 @@
               If the user is not root and the username is the same as
               primary group name, the umask group bits are set to be the
               same as owner bits (examples: 022 -> 002, 077 -> 007).
+	      Note that using this option explicitly is discouraged. pam_umask
+	      enables this functionality by default if /etc/login.defs enables
+	      USERGROUPS_ENAB, and the umask is not set explicitly in other
+	      places than /etc/login.defs (this is compatible with login's
+	      behaviour without PAM).
             </para>
           </listitem>
         </varlistentry>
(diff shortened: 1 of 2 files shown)


ldv-alt:
The commit message says about pam_unix but the module being changed is pam_umask; I'm confused.

andhe:
> The commit message says about pam_unix but the module being changed is pam_umask; I'm confused.

The commit message is a typo. It should say pam_umask.

andhe:
I think usergroups is rather a distribution option, than a user config switch. I'm mostly interested in distributions being able to set sane defaults. While during my archeology of login.defs for another case I came to a personal conclusion that I'd rather see the /etc/login.defs file deprecated and/or not shipped by default (and thus is no big supporter of moving options to login.defs) I wouldn't care much where this is configured as long as the default is sane (for the distribution shipping pam binaries). Would adding a compile-time option be acceptable to set the usergroups default so that distributions building binaries can use that? (Maybe there are many more options in pam than this one that are distribution choices which needs considerations rather than just looking at this option alone though.)

I noticed commit 65d6735c59 (and looking forward to a release including it), which AIUI allows shipping distribution defaults separately (and leave /etc empty/overridable for user configuration). I however think built-in sane defaults are even better (which eliminates the need to ship vendor configs) thus my above question.....

t8m:
The setting from login.defs should not override the pam module command-line option if set (either usergroups or nousergroups as recently merged from #164).

modules/pam_umask/pam_umask.8.xml
```diff
@@ -115,6 +116,11 @@
               If the user is not root and the username is the same as
               primary group name, the umask group bits are set to be the
               same as owner bits (examples: 022 -> 002, 077 -> 007).
+	      Note that using this option explicitly is discouraged. pam_umask
```
t8m: Indentation is wrong.

PR #99, opened 2019-02-19
Adding check if user is member of any group from the list.(pam_succeed_if.c)

sanicheev:
There is no way to check if user is a member of any group from the list like with an attribute from passwd struct for example.

Examples:
	account    requisite    pam_succeed_if.so user ingroup group1:group2
OR
	account    requisite    pam_succeed_if.so user notingroup group1:group2

Can be very convenient to grant access based on complex group memberships(LDAP, etc)

### DIFF
--- a/modules/pam_succeed_if/pam_succeed_if.8.xml
+++ b/modules/pam_succeed_if/pam_succeed_if.8.xml
@@ -198,27 +198,15 @@
         </listitem>
       </varlistentry>
       <varlistentry>
-        <term><option>user ingroup group</option></term>
+        <term><option>user ingroup group:group:...</option></term>
         <listitem>
-          <para>User is in given group.</para>
+          <para>User is in given group(s).</para>
         </listitem>
       </varlistentry>
       <varlistentry>
-        <term><option>user ismemberof group:group:...</option></term>
+        <term><option>user notingroup group:group:...</option></term>
         <listitem>
-          <para>User is a member of any group in the list.</para>
-        </listitem>
-      </varlistentry>
-      <varlistentry>
-        <term><option>user notingroup group</option></term>
-        <listitem>
-          <para>User is not in given group.</para>
-        </listitem>
-      </varlistentry>
-      <varlistentry>
-        <term><option>user isnonmemberof group:group:...</option></term>
-        <listitem>
-          <para>User is not a member of any group in the list.</para>
+          <para>User is not in given group(s).</para>
         </listitem>
       </varlistentry>
       <varlistentry>
@@ -287,6 +275,7 @@
     </para>
     <programlisting>
 auth required pam_succeed_if.so quiet user ingroup wheel
+auth required pam_succeed_if.so quiet user ingroup group1:group2
     </programlisting>
 
     <para>
--- a/modules/pam_succeed_if/pam_succeed_if.8.xml
+++ b/modules/pam_succeed_if/pam_succeed_if.8.xml
@@ -198,13 +198,13 @@
         </listitem>
       </varlistentry>
       <varlistentry>
-        <term><option>user ingroup group:group:...</option></term>
+        <term><option>user ingroup group[:group:....]</option></term>
         <listitem>
           <para>User is in given group(s).</para>
         </listitem>
       </varlistentry>
       <varlistentry>
-        <term><option>user notingroup group:group:...</option></term>
+        <term><option>user notingroup group[:group:....]</option></term>
         <listitem>
           <para>User is not in given group(s).</para>
         </listitem>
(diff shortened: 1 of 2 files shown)


t8m:
This can be already achieved (although not as conveniently) by using `user notingroup group1 user notingroup group2` and then doing then reversing the action by using `[success=die default=ignore]` if you need the ismemberof condition. Also the patch would be more acceptable if it just added the list parsing to ingroup notingroup conditions as groups cannot have ':' character in name anyway.

sanicheev:
> This can be already achieved (although not as conveniently) by using `user notingroup group1 user notingroup group2` and then doing then reversing the action by using `[success=die default=ignore]` if you need the ismemberof condition. Also the patch would be more acceptable if it just added the list parsing to ingroup notingroup conditions as groups cannot have ':' character in name anyway.

Hi i just took and applied example which already exists(this is from man):
field in item:item:...
           Field is contained in the list of items separated by colons.

Not sure if i understood correctly. Can you please give an example how do you want this to be implemented and i will try to adjust my PR?

t8m:
You would not add new action options ismemberof and isnotmemberof but you would just add the list splitting to the ingroup and notingroup actions. You could even allow the actions to be ingroups and notingroups - i.e. allow added 's'.

sanicheev:
> options

ohh ok i think i got it. i will modify my PR so evaluate_ingroup function will detect if it is a list and will process it accordingly

sanicheev:
Updated. Please check if it looks good.

modules/pam_succeed_if/pam_succeed_if.8.xml
```diff
@@ -275,6 +275,7 @@
     </para>
     <programlisting>
 auth required pam_succeed_if.so quiet user ingroup wheel
+auth required pam_succeed_if.so quiet user ingroup group1:group2
```
t8m: This could be `ingroup wheel:root.` And the sentence above could be adjusted like `except the fallback to group 0 being only approximated by checking also the root group membership`.

modules/pam_succeed_if/pam_succeed_if.8.xml
```diff
@@ -198,15 +198,15 @@
         </listitem>
       </varlistentry>
       <varlistentry>
-        <term><option>user ingroup group</option></term>
+        <term><option>user ingroup group:group:...</option></term>
```
t8m: Please use `group1[:group2:....]`

modules/pam_succeed_if/pam_succeed_if.8.xml
```diff
@@ -198,15 +198,15 @@
...
         <listitem>
-          <para>User is in given group.</para>
+          <para>User is in given group(s).</para>
         </listitem>
       </varlistentry>
       <varlistentry>
-        <term><option>user notingroup group</option></term>
+        <term><option>user notingroup group:group:...</option></term>
```
t8m: Same as above.

modules/pam_succeed_if/pam_succeed_if.c
```diff
@@ -217,16 +217,32 @@ evaluate_notinlist(const char *left, const char *right)
...
 {
-	if (pam_modutil_user_in_group_nam_nam(pamh, user, group) == 1)
-		return PAM_SUCCESS;
+	char *ptr = NULL;
+	const char const *delim = ":";
+	char const *grp = NULL;
+
+	grp = strtok_r(group, delim, &ptr);
+	while(grp != NULL) {
+		if (pam_modutil_user_in_group_nam_nam(pamh, user, group) == 1)
```
t8m: Typo: group->grp.

modules/pam_succeed_if/pam_succeed_if.c
```diff
@@ -217,16 +217,32 @@ evaluate_notinlist(const char *left, const char *right)
...
 {
-	if (pam_modutil_user_in_group_nam_nam(pamh, user, group) == 0)
-		return PAM_SUCCESS;
+	char *ptr = NULL;
+	const char const *delim = ":";
+	char const *grp = NULL;
+
+	grp = strtok_r(group, delim, &ptr);
+	while(grp != NULL) {
+		if (pam_modutil_user_in_group_nam_nam(pamh, user, group) == 0)
```
t8m: Same as above.

modules/pam_succeed_if/pam_succeed_if.8.xml
```diff
@@ -271,10 +271,10 @@
     <title>EXAMPLES</title>
     <para>
       To emulate the behaviour of <emphasis>pam_wheel</emphasis>, except
-      there is no fallback to group 0:
+      there is no fallback to group 0 being only approximated by checking also the root group membership:
```
t8m: Please replace `there is no fallback to group 0` with `the fallback to group 0`

t8m:
And please rebase the PR against fresh master and resolve the conflict.

t8m:
You did a merge instead of rebase. We do not want merge commits. Please drop the merge commit and do git rebase.

PR #105, opened 2019-03-05
propagate PAM_USER_UNKNOWN from unix_chkpwd.c meaningfully

qralston:
Consider a host where sssd(8) is used to map network accounts, via the sss provider for the NSS passwd and group databases.

On such a host, we want to use pam_unix.so to authenticate local accounts, and pam_sss.so to authenticate network accounts.  If there is both a local account and a network account with the same name, the local account should take precedence.

We update sssd.conf(5) to tell sssd that it should map the password field returns by getpwnam() as "x", to indicate to pam_unix.so that the shadow(5) file should be used:

```
[nss]
pwfield = x
```

Then we create a PAM auth stack something like this:

```
auth [default=1 authinfo_unavail=ignore success=done] pam_unix.so try_first_pass
auth sufficient pam_sss.so use_first_pass
auth required   pam_deny.so
```

But this will always fail for network accounts, because pam_unix.so will never return PAM_AUTHINFO_UNAVAIL on systems that use the unix_chkpwd helper.

The reason is that in unix_chkpwd.c, towards the end of main(), if helper_verify_password() does not return PAM_SUCCESS, main() ignores the actual error that helper_verify_password() returned and instead returns PAM_AUTH_ERR.

The only work-around for this issue is to attempt network authentication regardless of what pam_unix.so returns.  E.g.:

```
auth sufficient pam_unix.so try_first_pass
auth sufficient pam_sss.so use_first_pass
auth required   pam_deny.so
```

But this is problematic, because it makes it unsafe to use a local-then-network authentication strategy, because the creation of a new network account with the same name as a local account could unintentionally grant access for network users.

Furthermore, this makes it impossible to use pam_faillock.so in the auth stack, which is required by various security guides (e.g., the DISA RHEL7 STIG): if pam_unix.so fails, and you call pam_faillock.so authfail, you will strike every _successful_ network login, which is clearly wrong.

The root problem here is that _it is an error_ for unix_chkpwd to always override the return value of helper_verify_password(). The difference between "we failed because there was no local entry for this user" and "we failed because the password was incorrect" _must_ be propagated.

This merge request corrects this behavior. Specifically, if helper_verify_password() returns PAM_USER_UNKNOWN, unix_chkpwd exits with PAM_AUTHINFO_UNAVAIL, which is arguably the best interpretation of "getpwnam() succeeds for this user but getspnam() fails." For any other error from helper_verify_password(), unix_chkpwd continues to exit with PAM_AUTH_ERR.

I've tested locally that with this patch applied, the first PAM snippet above works correctly.


t8m:
Unfortunately your configuration leaks information whether a user is local on the system

t8m:
Clicked close too fast before finishing a sentence - the information is leaked to possible remote attacker via a timing difference.

qralston:
I do not understand why this pull request was closed. Could you please explain?

As I explained in the comment, the difference between "we failed because there was no local entry for this user" and "we failed because the password was incorrect" _must_ be propagated, because otherwise, it is impossible to use pam_faillock.so correctly when a mix of local and network accounts is in play. And we are working with systems where we must comply with the RHEL7 STIG, which requires the use of pam_faillock.so.

I understand the desire to not make it easier for users to accidentally create a PAM configuration that exposes timing information to remote attackers. But the priority of fixing outright bugs in the code outweighs this concern. And it is pointless, anyway: the hoops we have to jump through in our PAM stack (the one we actually use, not the hypothetical example above) to avoid this bug reveals the same timing information you are attempting to prevent. (As well as making our PAM stack needlessly complex.)

This is a bug. The code is wrong. My pull request corrects the bug. If you have an issue with how I have corrected the bug, then please provide feedback, and I will resubmit the pull request accordingly. But otherwise: please accept this pull request and fix the bug. Thank you.

t8m:
There is probably no reasonable way how to create a configuration that would have the properties that you want and at the same time it did not leak the timing info to the attacker. And that is regardless of whether this patch is merged or not.

To avoid the leakage it would have to be handled completely on the sssd side. I.e. let it handle both local and remote accounts, which is possible with recent releases of sssd.

But you're right that there is no reason why not to propagate the return value and this way to leave it up to the user to decide on to which of the evils to choose.

t8m:
I've merged your commit (after amending the commit message). Thank you for your contribution.

PR #113, opened 2019-05-10
pam_namespace: secure tmp-inst directories

### LINKED ISSUE #111: pam_namespace (or systemd) should ship a tmpfiles.d drop-in file to secure tmp-inst directory
When using polyinstantiation for `/tmp` and/or `/var/tmp`, `pam_namespace` creates subdirectories with fixed name `tmp-inst`. These paths should be secured as early as possible to avoid that somehow these directories could created and controlled by for example a malicious user or service. One way to do this would be to ship a `tmpfiles.d` drop-in file, which would create the directories early in boot sequence with correct permissions and ownership, maybe also SELinux labels (they are created by namespace.init later in any case, but perhaps they also should be correct earlier).

This drop-in file could be provided by the package which provides pam_namespace, or maybe by systemd.

topimiettinen:
When using polyinstantiation for /tmp and/or /var/tmp, pam_namespace
creates subdirectories with fixed name tmp-inst. These paths should be
secured as early as possible to avoid that somehow these directories
could created and controlled by for example a malicious user or
service.

Ship a tmpfiles.d drop-in file, which creates the directories early in
boot sequence with correct permissions and ownership.

Closes #111.

Signed-off-by: Topi Miettinen <toiwoton@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -687,6 +687,7 @@ AC_CONFIG_FILES([Makefile libpam/Makefile libpamc/Makefile libpamc/test/Makefile
 	modules/pam_loginuid/Makefile modules/pam_mail/Makefile \
 	modules/pam_mkhomedir/Makefile modules/pam_motd/Makefile \
 	modules/pam_namespace/Makefile \
+	modules/pam_namespace/pam_namespace_helper modules/pam_namespace/pam_namespace.service \
 	modules/pam_nologin/Makefile modules/pam_permit/Makefile \
 	modules/pam_pwhistory/Makefile modules/pam_rhosts/Makefile \
 	modules/pam_rootok/Makefile modules/pam_exec/Makefile \
--- a/modules/pam_namespace/Makefile.am
+++ b/modules/pam_namespace/Makefile.am
@@ -8,7 +8,7 @@ MAINTAINERCLEANFILES = $(MAN5) $(MAN8) README
 
 if HAVE_DOC
 MAN5 = namespace.conf.5
-MAN8 = pam_namespace.8
+MAN8 = pam_namespace.8 pam_namespace_helper.8
 endif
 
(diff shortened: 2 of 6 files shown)


Maryse47:
`pam_namespace` doesn't do anything automatically when installed so why it should create those dirs?

topimiettinen:
You're right, a better solution would be to ship a startup service/script, which reads `/etc/security/namespace.conf` and only if enabled, create the configured directories during boot.

There's still the case where `pam_namespace` is not enabled during boot but then administrator enables it later. After enabling but before the next user logs in, there could be a window where the directories are not protected. I think this could be better solved with documentation, for example as a comment in `/etc/security/namespace.conf` which proposes to secure any directories before enabling.

 I'll make an update.

topimiettinen:
Updated. I also excluded lines with `$HOME` from processing, they probably always exist.

topimiettinen:
@thkukuk Could you please review this?

t8m:
I think this is fine. @thkukuk @ldv-alt would you please provide a second review?

t8m:
As the helper is in sbin, could you please add a documentation in form of a simple manual page describing its purpose?

topimiettinen:
@t8m makes sense, updated.

ldv-alt:
Note that this PR fails the CI test.

topimiettinen:
@ldv-alt Thanks. The CI failed due to uninstall rule, which didn't remove the systemd service file. Should be fixed in the force pushed version. I wonder why I didn't see that earlier.

t8m:
Could you please rebase and fix the conflict?

topimiettinen:
@t8m rebased & fixed.

t8m:
Merged, thank you for the PR.

PR #115, opened 2019-05-23
pam_succeed_if: Request user data only when needed

olafmandel:
Allow for conditions that just check the user field to also work for
users not known to the system. Before this caused a PAM_USER_UNKNOWN
even if no extra data for an existing user was needed. E.g.

    auth sufficient pam_succeed_if.so user = NotKnownToSystem

modules/pam_succeed_if/pam_succeed_if.c (evaluate): Change the pwd
parameter to an input/output parameter. Lazily request pwd with
pam_modutil_getpwnam() if needed and return PAM_USER_UNKNOWN on failure.

modules/pam_succeed_if/pam_succeed_if.c (pam_sm_authenticate): Don't
request the pwd if !use_uid anymore and shift the output from audit to
after the evaluate() call. Also make sure not to give the normal failure
message if the lazy pwd loading failed.


t8m:
LGTM.

t8m:
Merged. Thank you for the patch.

Issue #118, opened 2019-06-25
Option to specify maximum password length


spaceone:
Hello ;-)
If would be nice if pam-unix would provide a configuration option to specify a maximum allowed password length.
A client could potentially supply a password which is > 10000000 (or more) characters long. During the hashsum/crypt generation (sha512, sha256, etc.) of the supplied password the process is blocked and therefore allows Denial of Service. Especially in processes where PAM is called in a (e.g. python) thread with global interpreter lock, this blocks the whole process and deadlocks.

The limit should probably be configurable for `pam_authenticate()`, `pam_acct_mgmt()` (not sure) and `pam_chauthtok()`.
Nice would be, if the user gets a error message explaining that the maximum password length limit of %d characters was reached.

spaceone:
https://github.com/linux-pam/linux-pam/blob/master/modules/pam_unix/passverify.c#L149

spaceone:
gdb output of hanging process:
```
Thread 2 (Thread 0x7f9b6e1bd700 (LWP 21819)):
#0  sha512_process_block (buffer=<optimized out>, len=len@entry=49792, ctx=ctx@entry=0x7f9b6e1bb5b0) at ./sha512-block.c:70
#1  0x00007f9b8a4e3dbb in __sha512_process_bytes (buffer=<optimized out>, buffer@entry=0x7f9b680cd920, len=<optimized out>, len@entry=50010, ctx=ctx@entry=0x7f9b6e1bb5b0) at sha512.c:215
#2  0x00007f9b8a4e08c2 in __sha512_crypt_r (key=<optimized out>, key@entry=0x7f9b680cd920 "foobarblah", 'A' <repeats 190 times>..., salt=0x7f9b6e1bb328 "p3/yYoizYnBYiz1B\\u5\217\233\177",
    salt@entry=0x7f9b680c69a0 "$6$p3/yYoizYnBYiz1B$TnZ5ULjUxy8hU8AotZpXsavCY474omncOUCc7SIqN9jUZ9E64jLClRZTYkexzuoEdQRGyYZP2kTEQ/iHME9I41", buffer=<optimized out>, buflen=<optimized out>, buflen@entry=131232) at sha512-crypt.c:240
[...]

PR #120, opened 2019-06-26
Trim password at PAM_MAX_RESP_SIZE chars

spaceone:
Fixes Issue #118: Protect against Denial of Service attacks.
To prevent hashsum generation via crypt of very long passwords the
password is now stripped to 512 characters. This is equivalent behavior
to `unix_chkpwd`.

Please be careful in code-reviewing. I am not a usual C programmer.
Aspects I considered:
If `pw` (prior `p`) is a NULL pointer, the called sub-functions will still be called with a NULL pointer.
If `pw` is an empty string or any string the sub-functions are called with that string trimmed to 512 characters. A null byte is appended to the string implicit via `memset()`. I used `strncpy()` as this is standard C's `string.h` (not `glibc`).


spaceone:
Are there any Continues Integration tests e.g. run via TravisCI, etc.?

spaceone:
Do we need to adjust more places, e.g. when changing passwords via `pam_chauthtok`?

t8m:
Checking the length of new password on pam_chauthtok would make sense - not from DoS perspective but to avoid setting a password that cannot be authenticated against.

modules/pam_unix/support.c
```diff
@@ -641,11 +641,13 @@ _unix_blankpasswd (pam_handle_t *pamh, unsigned long long ctrl, const char *name
 }
 
 int _unix_verify_password(pam_handle_t * pamh, const char *name
-			  ,const char *p, unsigned long long ctrl)
+			  ,const char *pw, unsigned long long ctrl)
 {
 	struct passwd *pwd = NULL;
 	char *salt = NULL;
 	char *data_name;
+	char _p[MAX_PASS + 1] = "";
```
t8m: Please use some different name not started with underscore. Also the initializer is redundant as you clear it with memset anyway.
You could also avoid renaming p to pw and just reuse the parameter.

t8m:
> Are there any Continues Integration tests e.g. run via TravisCI, etc.?

Unfortunately no, that is something that is on my TODO list for a long time.

modules/pam_unix/support.c
spaceone: But `pw` it is a `const` so I cannot change it.

t8m: It is a pointer to const data not a constant pointer. So you can freely change the pointer to point it at the copied truncated buffer.

spaceone: Thanks, I updated the PR accordingly. Looks better?

spaceone: Maybe also:
```diff --git a/modules/pam_unix/support.c b/modules/pam_unix/support.c
index 1171f63..e5415f5 100644
--- a/modules/pam_unix/support.c
+++ b/modules/pam_unix/support.c
@@ -787,6 +787,7 @@ int _unix_verify_password(pam_handle_t * pamh, const char *name
        }

 cleanup:
+       memset(pw, 0, sizeof(pw)); /* clear memory of the password */
        if (data_name)
                _pam_delete(data_name);
        if (salt)
```

spaceone:
With pam_chauthtok the same denial of service is possible:

```
#0  0x00007fffef258a9f in sha512_process_block (buffer=<optimized out>, len=len@entry=99712, ctx=ctx@entry=0x7fffebfa1db0) at ./sha512-block.c:64
#1  0x00007fffef258dbb in __sha512_process_bytes (buffer=<optimized out>, buffer@entry=0x7fffe4037870, len=<optimized out>, len@entry=100010, ctx=ctx@entry=0x7fffebfa1db0) at sha512.c:215
#2  0x00007fffef2558c2 in __sha512_crypt_r (key=<optimized out>, salt=0x7fffebfa1b38 "/XFmw61l", buffer=<optimized out>, buflen=<optimized out>) at sha512-crypt.c:240
#3  0x00007fffef4920e1 in create_password_hash (pamh=pamh@entry=0x555555ea5840, password=0x7fffe4037870 "univention", 'A' <repeats 190 times>..., ctrl=ctrl@entry=142608660, rounds=<optimized out>) at passverify.c:426
#4  0x00007fffef48ee77 in pam_sm_chauthtok (pamh=0x555555ea5840, flags=<optimized out>, argc=<optimized out>, argv=<optimized out>) at pam_unix_passwd.c:822
#5  0x00007ffff48d50d7 in _pam_dispatch_aux (use_cached_chain=<optimized out>, resumed=<optimized out>, h=0x555555ec5c50, flags=8192, pamh=0x555555ea5840) at pam_dispatch.c:110
#6  _pam_dispatch (pamh=pamh@entry=0x555555ea5840, flags=flags@entry=8192, choice=choice@entry=6) at pam_dispatch.c:395
#7  0x00007ffff48d9483 in pam_chauthtok (pamh=0x555555ea5840, flags=0) at pam_password.c:46
#8  0x00007ffff4ae23d5 in PyPAM_chauthtok (self=<pam at remote 0x7fffeb72af70>, args=<optimized out>) at PAMmodule.c:248
#9  0x000055555564f84a in call_function (oparg=<optimized out>, pp_stack=0x7fffebfa23e8) at ../Python/ceval.c:4352
#10 PyEval_EvalFrameEx () at ../Python/ceval.c:2989
```
The user specifies any *current* password and `pam_chauthtok` tries to validate it in the same manner.

t8m:
This looks good, please just squash the WIP commit into the first commit and I will merge it.

The pam_chauthtok case at least does not apply to previously unauthenticated users so the severity there is much lower but this patch solves it for pam_chauthtok as well.

ldv-alt:
Please do not add English translation strings for non-English locales.

spaceone:
The branch has been squashed.
The translations have been adjusted, to keep empty messagestring for non-english locales.
I tested the changes with `libpam-modules 1.1.8-3.6` - works, except that there these macros doesn't exists yet.

t8m:
OK, nice, thank you for the work.

PR #125, opened 2019-07-16
Add autogen.sh in the build documentation

2xyo:


t8m:
autogen.sh needs to be run only if you check out the sources from git. If you download a release tarball, it is not needed. Please adjust the change to mention this.

t8m:
Merged manually with squash. Thanks for the patch.

PR #128, opened 2019-08-01
pam_mkhomedir: add "userprivategroup" option that uses the group

elric1:
named after the user if it exists rather than the user's
primary group.

### DIFF
--- a/modules/pam_mkhomedir/pam_mkhomedir.c
+++ b/modules/pam_mkhomedir/pam_mkhomedir.c
@@ -59,8 +59,9 @@
 #include <security/pam_ext.h>
 
 /* argument parsing */
-#define MKHOMEDIR_DEBUG      020	/* be verbose about things */
-#define MKHOMEDIR_QUIET      040	/* keep quiet about things */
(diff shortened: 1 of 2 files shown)


andhe:
This is similar/same-as/related-to the pam_umask "usergroups"  options? I think it would be good to use the same name everywhere please....

t8m:
I would say it is only remotely related so it does not have to be necessarily the same. What is missing for the enhancement to be merged is a documentation of the new option.
Can you please add it?

modules/pam_mkhomedir/pam_mkhomedir.c
```diff
@@ -88,6 +89,8 @@ _pam_parse (const pam_handle_t *pamh, int flags, int argc, const char **argv,
 	 opt->ctrl |= MKHOMEDIR_QUIET;
       } else if (!strcmp(*argv, "debug")) {
          opt->ctrl |= MKHOMEDIR_DEBUG;
+      } else if (!strcmp(*argv, "userprivategroup")) {
```
t8m: Please add the documentation to the man page.

PR #129, opened 2019-08-01
pam_mkhomedir: call umask(2) and let the OS do the masking

elric1:
rather than chmod(2)ing everything that we create.  This fixes a bug
where the sgid bit from the parent directory was not inherited.  We also
try to preserve the permissions of the sub-directories from the skeleton
directory.

### DIFF
--- a/modules/pam_mkhomedir/mkhomedir_helper.c
+++ b/modules/pam_mkhomedir/mkhomedir_helper.c
@@ -26,13 +26,12 @@
 #include <security/pam_ext.h>
 #include <security/pam_modutil.h>
 
-static unsigned long u_mask = 0022;
 static char skeldir[BUFSIZ] = "/etc/skel";
 
 /* Do the actual work of creating a home dir */
 static int
 create_homedir(const struct passwd *pwd,
-	       const char *source, const char *dest)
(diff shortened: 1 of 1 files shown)


modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -26,21 +26,20 @@
 #include <security/pam_ext.h>
 #include <security/pam_modutil.h>
 
-static unsigned long u_mask = 0022;
```
t8m: Please do not drop the 0022 default umask handling.

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -378,12 +375,16 @@ main(int argc, char *argv[])
...
+	mode_t u_mask;
+
 	errno = 0;
 	u_mask = strtoul(argv[2], &eptr, 0);
 	if (errno != 0 || *eptr != '\0') {
 		pam_syslog(NULL, LOG_ERR, "Bogus umask value %s", argv[2]);
 		return PAM_SESSION_ERR;
 	}
+
+	umask(u_mask);
```
t8m: Please call umask with the default 0022 value when the umask argument is not specified.

PR #138, opened 2019-09-06
Fix the man page for "pam_fail_delay()"

lifecrisis:
See the commit messages for details.

### DIFF
--- a/doc/man/pam_fail_delay.3.xml
+++ b/doc/man/pam_fail_delay.3.xml
@@ -98,10 +98,10 @@ void (*delay_fn)(int retval, unsigned usec_delay, void *appdata_ptr);
       Note that the PAM_FAIL_DELAY item is set to NULL by default. This
       indicates that PAM should perform a random delay as described
       above when authentication fails and a delay has been suggested.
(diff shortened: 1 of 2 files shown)


doc/man/pam_fail_delay.3.xml
```diff
@@ -93,8 +93,15 @@ void (*delay_fn)(int retval, unsigned usec_delay, void *appdata_ptr);
...
         <refentrytitle>pam_set_item</refentrytitle><manvolnum>3</manvolnum>
       </citerefentry>.
-      Note, if PAM_FAIL_DELAY item is unset (or set to NULL), then no delay
-      will be performed.
+    </para>
+    <para>
+      Note that the PAM_FAIL_DELAY item is set to NULL by default. This
+      indicates that PAM should perform a random delay as described
+      above when authentication fails and a delay has been suggested.
+      An application may perfer that a delay never be performed on
```
t8m: Typo `perfer`. I would also recommend rather something like: "If an application does not want the PAM library to perform any delay on authentication failure, the application should define....".

t8m:
Merged manually - squashed the last two commits.

PR #140, opened 2019-09-11
pam_lastlog: prevent crash due to reduced 'fsize' limit

casantos:
It a reduced fsize limit is set in /etc/security/limits.conf and
pam_limits is in use pam_lastlog may cause a crash, e.g.

    ----- begin /etc/pam.d/su ----
    auth        sufficient   pam_rootok.so
    auth        required     pam_wheel.so use_uid
    auth        required     pam_env.so
    auth        required     pam_unix.so nullok
    account     required     pam_unix.so
    password    required     pam_unix.so nullok
    session     required     pam_limits.so
    session     required     pam_env.so
    session     required     pam_unix.so
    session     optional     pam_lastlog.so
    ----- end /etc/pam.d/su -----

    ----- begin /etc/security/limits.d/fsize.conf -----
    * soft fsize 1710
    * hard fsize 1710
    ----- end /etc/security/limits.d/fsize.conf -----

    # id user1
    uid=1000(user1) gid=1000(user1) groups=1000(user1)
    # su - user1
    Last login: Wed Sep 11 01:52:44 UTC 2019 on console
    $ exit
    # id user2
    uid=60000(user2) gid=60000(user2) groups=60000(user2)
    # su - user2
    File size limit exceeded

This happens because pam_limits sets RLIMIT_FSIZE before pam_lastlog
attempts to write /var/log/lastlog, leading to a SIGXFSZ signal.

In order to fix this, save the 'fsize' limit and set it to unlimited
before writing lastlog. After that, restore the saved value.

Failing to set the 'fsize' limit is not a fatal error.  With luck the
configured limit will suffice, so we try to write lastlog anyway, even
under the risk of dying due to a SIGXFSZ.

Failing to restore the 'fsize' limit is a fatal error, since we don't
want to keep it unlimited.

Signed-off-by: Carlos Santos <casantos@redhat.com>


modules/pam_lastlog/pam_lastlog.c
```diff
@@ -427,13 +436,44 @@ last_login_write(pam_handle_t *pamh, int announce, int last_fd,
...
     }
 
+    /*
+     * Failing to set the 'fsize' limit is not a fatal error. We try to write
+     * lastlog anyway, under the risk of dying due to a SIGXFSZ.
+     */
+    D(("setting limit for 'fsize'"));
+    getrlimit_res = getrlimit(RLIMIT_FSIZE, &old_limit);
+    if (getrlimit_res == 0) {
+	setrlimit_res = setrlimit(RLIMIT_FSIZE, &no_limit);
```
t8m: Could you please also check whether the current limit is infinity and in that case avoid calling setrlimit()?

t8m:
This is something to consider @thkukuk  @ldv-alt 
What if the limit was set to avoid making the lastlog file large? Then by overriding it we are basically making an incompatible change.

The other option would be to catch/ignore SIGXFSZ signal and not to crash - but in that case the module should still fail and it should be made optional in the PAM stack configuration if the user does not want to block on its failure.

thkukuk:
The size of the lastlog file depends on the highest UID used on that system. If you want to avoid that the lastlog file is getting that large, you need to lower the UIDs on your system. Setting a limit will only break the lastlog functionality, which means pam_lastlog is no long functional. In this case, it would be better to disable lastlog functionality on that system.

So if people want to use lastlog, we either have to ignore the limit for that file, or print an error, that the limit is too small. Currently I would go for the first option, ignore the limit for that file.

t8m:
> So if people want to use lastlog, we either have to ignore the limit for that file, or print an error, that the limit is too small. Currently I would go for the first option, ignore the limit for that file.

Which is what this patch basically does. OK then, I will merge it after the change requested above is done.

casantos:
> > So if people want to use lastlog, we either have to ignore the limit for that file, or print an error, that the limit is too small. Currently I would go for the first option, ignore the limit for that file.
> 
> Which is what this patch basically does. OK then, I will merge it after the change requested above is done.

Perhaps we should add an "unlimited" option. This would preserve backward compatibility by default but allow the user to choose ignoring the limit.

t8m:
I was thinking about that but it seems to me to be a little bit overstretching it.

casantos:
> I was thinking about that but it seems to me to be a little bit overstretching it.

It's preferable to maintain the current behaviour by default, since it is what users would expect or desire. There are even Red Hat solutions documenting it and making the change behind the scenes may cause surprises. So I sent a new patch set adding the "unlimited" option.

PR #144, opened 2019-10-02
pam_mkhomedir: Add debug option to pam_mkhomedir(8) man page

miztake:
pam_mkhomedir has 'debug' option, but it is not listed in the man page.
Add 'debug' option to the man page so that the user can specify 'debug' option.


t8m:
Looks fine. Thank you.

PR #146, opened 2019-10-14
doc: fix module type written in MODULE TYPES PROVIDED

miztake:
Fixed the provided module type in pam_nologin and pam_rootok (acct -> account).


t8m:
Merged, thank you for the PR.

PR #150, opened 2019-11-03
Adds an auth module to pam_keyinit

### LINKED ISSUE #149: pam_keyinit lacks pam_sm_setcred
Since many PAM applications execute _pam_setcred_ before _pam_open_session_, the session keyring is not available when it is needed.

For example: if we use pam_krb5 with credential cache of the form KEYRING:session:some_name and login using gdm3, the credential cache will not be available to the user session (pam_krb5 fills in the credential cache during in whichever of the previously mentioned functions gets called first).

I propose to add an auth module to pam_keyinit, with a _pam_sm_setcred_ function identical to _pam_sm_open_session_.

ppkarwasz:
Adds an auth module to pam_keyinit, whose implementation of

pam_sm_setcred

is identical to the implementation of pam_sm_open_session.

It is useful with PAM applications, which call pam_setcred,
before calling pam_open_session. It closes #149.

### DIFF
--- a/modules/pam_keyinit/pam_keyinit.c
+++ b/modules/pam_keyinit/pam_keyinit.c
@@ -30,9 +30,9 @@
 #define KEYCTL_REVOKE			3 /* revoke a key */
 #define KEYCTL_LINK			8 /* link a key into a keyring */
 
-static int my_session_keyring;
-static int session_counter;
-static int do_revoke;
+static int my_session_keyring = 0;
+static int session_counter = 0;
+static int do_revoke = 0;
 static int revoke_as_uid;
 static int revoke_as_gid;
 static int xdebug = 0;
@@ -68,7 +68,7 @@ static int error(pam_handle_t *pamh, const char *fmt, ...)
 /*
  * initialise the session keyring for this process
  */
-static int init_keyrings(pam_handle_t *pamh, int force, int ERROR)
+static int init_keyrings(pam_handle_t *pamh, int force, int error_ret)
 {
 	int session, usession, ret;
 
@@ -85,7 +85,7 @@ static int init_keyrings(pam_handle_t *pamh, int force, int ERROR)
 			 * installed */
 			if (errno == ENOSYS)
 				return PAM_SUCCESS;
-			return ERROR;
+			return error_ret;
 		}
 
 		usession = syscall(__NR_keyctl,
@@ -94,7 +94,7 @@ static int init_keyrings(pam_handle_t *pamh, int force, int ERROR)
 				   0);
 		debug(pamh, "GET SESSION = %d", usession);
 		if (usession < 0)
-			return ERROR;
+			return error_ret;
 
 		/* if the user session keyring is our keyring, then we don't
 		 * need to do anything if we're not forcing */
@@ -108,7 +108,7 @@ static int init_keyrings(pam_handle_t *pamh, int force, int ERROR)
 		      NULL);
 	debug(pamh, "JOIN = %d", ret);
 	if (ret < 0)
-		return ERROR;
+		return error_ret;
 
 	my_session_keyring = ret;
 
@@ -118,7 +118,7 @@ static int init_keyrings(pam_handle_t *pamh, int force, int ERROR)
 		      KEY_SPEC_USER_KEYRING,
 		      KEY_SPEC_SESSION_KEYRING);
 
-	return ret < 0 ? ERROR : PAM_SUCCESS;
+	return ret < 0 ? error_ret : PAM_SUCCESS;
 }
 
 /*
@@ -162,7 +162,7 @@ static void kill_keyrings(pam_handle_t *pamh)
 	}
 }
 
-static int do_open_session(pam_handle_t *pamh, int argc, const char **argv, int ERROR)
+static int do_keyinit(pam_handle_t *pamh, int argc, const char **argv, int error_ret)
 {
 	struct passwd *pw;
 	const char *username;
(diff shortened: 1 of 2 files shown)


modules/pam_keyinit/pam_keyinit.c
```diff
@@ -162,11 +162,7 @@ static void kill_keyrings(pam_handle_t *pamh)
 	}
 }
 
-/*
- * open a PAM session by making sure there's a session keyring
- */
-int pam_sm_open_session(pam_handle_t *pamh, int flags UNUSED,
-			int argc, const char **argv)
+static int do_open_session(pam_handle_t *pamh, int argc, const char **argv, int ERROR)
```
t8m: Name the function differently - I'd suggest do_keyinit or something similar. Also please note that the session_counter should not be incremented when calling from pam_sm_setcred.

modules/pam_keyinit/pam_keyinit.c
```diff
@@ -234,6 +230,31 @@ int pam_sm_open_session(pam_handle_t *pamh, int flags UNUSED,
...
+{
+  return do_open_session(pamh, argc, argv, PAM_CRED_ERR);
+}
+
+int pam_sm_open_session(pam_handle_t *pamh, int flags UNUSED,
+			int argc, const char **argv)
+{
+  return do_open_session(pamh, argc, argv, PAM_SESSION_ERR);
+}
+
```
t8m: To be on the safe side in pam_sm_close_session I'd recommend changing `if (session_counter == 0 ... `to `if (session_counter <= 0 ...`

modules/pam_keyinit/pam_keyinit.c
```diff
@@ -234,6 +230,31 @@ int pam_sm_open_session(pam_handle_t *pamh, int flags UNUSED,
 	return ret;
 }
 
+/*
+ * Dummy
+ */
+int pam_sm_authenticate(pam_handle_t *pamh UNUSED, int flags UNUSED,
+                   int argc UNUSED, const char **argv UNUSED)
+{
+  return PAM_SUCCESS;
```
t8m: Here PAM_IGNORE should be returned to make it similar to the other setcred-only modules.

ppkarwasz:
@t8m, thank you for the suggestions.

By reading some other PAM modules (notably [pam-krb5](https://github.com/rra/pam-krb5)), I have some doubts about the use of `setreuid()` in the module. On the one hand it is required to the correct linking of the user keyring, but on the other hand it messes up the PAM application if it is multi-threaded and the `pam_setcred` man page states:
> On a Linux system the user's UID and GID's are credentials too.
> However, it has been decided that these properties (along with the default supplementary
> groups of which the user is a member) are credentials that should
> be set directly by the application and not by PAM. Such credentials should be established,
> by the application, prior to a call to this function.

t8m:
I do not really think pam_setcred() is supposed to be called from multithreaded applications. Other PAM modules with pam_sm_setcred() would break multithreaded applications as well, although in possibly more subtle ways.

t8m:
@ldv-alt @thkukuk What is your opinon on calling setreuid() from pam_sm_setcred()? BTW pam_env module already does it.

thkukuk:
if pam_krb5 depends on pam_keyinit in the session part, it should be called after pam_keyinit and not before in the stack. That's is at least how we solved this with other modules.

I don't think that this change is correct and I doubt it will not break existing configurations and has no bad sideeffects.
If called with pam_setcred, the arguments are ignored, as nobody will confiure  pam_keyinit for auth.
What happens if we do the init_keyring twice? I would expect, that we get a second keyring and the first is no longer accessible. So what's with the modules called in between?

ppkarwasz:
The reason behind this patch is the confusion that exists on the calling order of `pam_setcred` and `pam_open_session`.

I have a standard Debian configuration: **pam_keyinit** is one of the first modules in every application's stack, **pam_krb5** is in `common-*` files which are included after **pam_keyinit**.

For some reasons @rra [explained shortly](https://github.com/rra/pam-krb5/issues/14#issuecomment-549179249), **pam_krb5** permanently stores its ticket cache during `pam_setcred`. Depending on the PAM application I use for logging in I end up with or without a ticket cache:
1. **login** calls `pam_open_session` before `pam_setcred` and I have a ticket cache in my session ring,
2. **gdm3** calls `pam_setcred` before `pam_open_session` and I end up with my ticket cache in gdm's session ring (which probably gets unlinked and vanishes when **pam_keyinit** is called).

The patched **pam_keyinit** should create a new session keyring in the first of `pam_setcred` and `pam_open_session` that gets called, so nothing should change if **pam_keyinit** is not in the _auth_ stack. On the other hand I can put it in **gdm3**'s stack and get a functional ticket cache.

Since I am not a PAM expert, I do not fully understand the difference between `pam_setcred` and `pam_open_session`, notably:
* which one should be called first (I assume `pam_authenticate` is the first),
* whether a session keyring and Kerberos ticket cache should be considered credentials to set in `pam_setcred` or part of a session to set in `pam_open_session` (I have got one opinion in the [discussion linked before](https://github.com/rra/pam-krb5/issues/14#issuecomment-549179249))

Hence I am not sure if you should consider this patch as a solution to **pam_keyinit**'s bug or a workaround to other application's bugs.

t8m:
@thkukuk As in the current configurations pam_keyinit could not be present in the auth section it would not be called from pam_setcred(). So there is no problem with existing configurations.
Also there should be no doubling of init_keyring call as there is a guard in the pam_keyinit code against that.

t8m:
@ppkarwasz - there should be one more thing in the patch the do_keyinit should be called from pam_sm_setcred only when PAM_ESTABLISH_CRED flag is set.
`if (flags & PAM_ESTABLISH_CRED)`

ppkarwasz:
Thanks, I added support for `PAM_ESTABLISH_CRED` and `PAM_DELETE_CRED`. Other flags are ignored.

t8m:
I've forgot about one more thing that is missing - documenting properly the new functionality in the pam_keyinit manpage. After that is added I will merge unless @thkukuk or @ldv-alt is against that.

PR #154, opened 2019-11-22
Lower "bad username" log priority

sophie-h:
This is more in conformance with how other message priorities are chosen.


t8m:
Merged, thank you for the patch.

PR #157, opened 2019-12-14
Keep returning only PAM_IGNORE from pam_motd

rbalint:
Follow-up for c81280b16e1831ab0bdd0383486c7e2d1eaf1b5e.

### DIFF
--- a/modules/pam_motd/pam_motd.8.xml
+++ b/modules/pam_motd/pam_motd.8.xml
@@ -133,10 +133,26 @@
     <title>RETURN VALUES</title>
     <variablelist>
       <varlistentry>
+        <term>PAM_ABORT</term>
+        <listitem>
+           <para>
+             Not all relevant data or options could be obtained.
+          </para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>PAM_BUF_ERR</term>
+        <listitem>
+           <para>
+              Memory buffer error.
+          </para>
+        </listitem>
+      </varlistentry>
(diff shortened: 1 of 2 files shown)


ldv-alt:
On Sat, Dec 14, 2019 at 02:22:21AM -0800, Balint Reczey wrote:
> Follow-up for c81280b16e1831ab0bdd0383486c7e2d1eaf1b5e.
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/linux-pam/linux-pam/pull/157

I suppose errors returned by pam_putenv shouldn't be ignored.
What do you think about the following fix:

-    return retval;
+    return retval == PAM_SUCCESS ? PAM_IGNORE : retval;

rbalint:
OK, I've also updated the man page.

modules/pam_motd/pam_motd.8.xml
```diff
@@ -132,11 +132,27 @@
   <refsect1 id='pam_motd-return_values'>
     <title>RETURN VALUES</title>
     <variablelist>
+      <varlistentry>
+        <term>PAM_ABORT</term>
+        <listitem>
+           <para>
+             Not all relevant data or options could be gotten.
```
t8m: got or obtained - although sometimes used gotten is weird.

rbalint: I agree. I copied this over from `pam_env` and the same phrasing is present in other modules, too, thus I did not want to diverge.

PR #158, opened 2019-12-17
pam_securetty: silently ignore missing securetty file

rbalint:
Fedora stopped shipping /etc/securetty years ago and it is also not shipped in
Debian and Debian derivatives either because nowadays it is not found to be
useful.
Since the default state is not having /etc/securetty on the system PAM noticing
its absence also should not show up in syslog either.

### DIFF
--- a/modules/pam_securetty/pam_securetty.c
+++ b/modules/pam_securetty/pam_securetty.c
@@ -119,19 +119,14 @@ securetty_perform_check (pam_handle_t *pamh, int ctrl,
 #ifdef VENDORDIR
       if (errno == ENOENT) {
 	if (stat(SECURETTY2_FILE, &ttyfileinfo)) {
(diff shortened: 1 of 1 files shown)


thkukuk:
I disagree. If you configure pam_securetty, then you expect that it will work and report an error and not silently stops working.
If you don't need pam_securetty, don't configure it.
I'm pretty sure that Debian and Fedora don't configure pam_securetty by default if they don't ship a /etc/securetty file.

rbalint:
@thkukuk ACK.
The state in Debian is actually an open question and I wanted to get feedback about just silencing it.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=932376
I'm working towards disabling it then.

zeha:
This has now happened in e842a5fc075002f46672ebcd8e896624f1ec8068 instead. ...

PR #162, opened 2019-12-24
Fix (IPv6) address prefix size matching

### LINKED ISSUE #161: IPv6 prefix size matching doesn't allow /128
The code in [pam_access.c line 740](https://github.com/linux-pam/linux-pam/blob/master/modules/pam_access/pam_access.c#L740) incorrectly discards address prefix sizes `>= 128`, instead of `> 128`. This means we cannot specify a host address such as `::1/128`.
Additionally, the current code does not distinguish between IPv4 and IPv6 addresses, since the same range check is used for both. We can therefore match addresses such as `127.0.0.1/32`. We can actually even match `127.0.0.1/127` which should have failed, but is handled as equivalent to `/32`.

msalle:
IPv6 address prefix sizes larger than 128 (i.e. not larger or equal to) should
be discarded. Additionally, for IPv4 addresses, the largest valid prefix size
should be 32.
This fixes #161

### DIFF
--- a/modules/pam_access/pam_access.c
+++ b/modules/pam_access/pam_access.c
@@ -737,7 +737,9 @@ network_netmask_match (pam_handle_t *pamh,
 		{ /* invalid netmask value */
 		  return NO;
 		}
-	    if ((netmask < 0) || (netmask >= 128))
+	    if ( (netmask < 0) ||
+		 (addr_type==AF_INET && netmask > 32) ||
+		 (addr_type==AF_INET6 && netmask > 128) )
 		{ /* netmask value out of range */
 		  return NO;
 		}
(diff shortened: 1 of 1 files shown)


modules/pam_access/pam_access.c
```diff
@@ -737,7 +737,9 @@ network_netmask_match (pam_handle_t *pamh,
 		{ /* invalid netmask value */
 		  return NO;
 		}
-	    if ((netmask < 0) || (netmask >= 128))
+	    if ( (netmask < 0) ||
```
t8m: Just please fix these formatting nits: Do not put spaces after the initial parenthesis after if and before the last parenthesis. Put spaces around `==` and move the `||` at the beginning of the line for better readability.

msalle: Hi, thanks for the comments, I've fixed the formatting.

PR #164, opened 2020-01-02
pam_umask: build-time usergroups option default

andhe:
This change adds a configure option to set the default value of the
usergroups option (of the pam_umask module) at build-time.

Distributions usually makes the decision if usergroups should be used or
not. This allows them to control the built-in default value, without
having to ship the value in a config file (cluttering up the view
of actually relevant user/system configuration overrides).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -487,6 +487,17 @@ AS_IF([test "x$enable_nis" != "xno"], [
 AC_SUBST([NIS_CFLAGS])
 AC_SUBST([NIS_LIBS])
 
+AC_ARG_ENABLE([usergroups],
+  AS_HELP_STRING([--enable-usergroups], [sets the usergroups option default to enabled]),
+  [WITH_USERGROUPS=$enableval], WITH_USERGROUPS=no)
+if test "$WITH_USERGROUPS" = "yes" ; then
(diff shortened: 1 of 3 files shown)


andhe:
With this distributions no longer needs to ship a file to have a sane default, and thus hopefully Debian can finally (after a decade of discussion/consideration!) have a sane default when we no longer need to discuss where the file/setting should live in the filesystem.

While I prefer this approach over https://github.com/linux-pam/linux-pam/pull/97 this change acually increases the value of merging that one as well, as without #97 there's no option to disable usergroups if the distribution built with it enabled by default.

t8m:
The approach is good however if enabled by default I would like to see an option to turn it off from the PAM command line.

andhe:
> The approach is good however if enabled by default I would like to see an option to turn it off from the PAM command line.

Thanks for your feedback. As mentioned #97 would solve that concern, but I guess you want this PR to be self-contained and really add a command line argument for it so I pushed an additional commit adding a 'nousergroups' that is similar to the pre-existing 'usergroups' argument. Hope this meets all your expectations and if there's anything else needed to get this merged please tell me and I'll try to work on it!

t8m:
Merged. Thank you for the PR.

PR #166, opened 2020-01-14
pam_usertype: new module to tell if uid is in login.defs ranges

pbrezina:
This module will check if the user account type is system or regular based
on its uid. To evaluate the condition it will use 0-99 reserved range
together with `SYS_UID_MIN` and `SYS_UID_MAX` values from `/etc/login.defs`.

If these values are not set, it uses configure-time defaults
`--with-sys-uid-min` and `--with-uid-min` (according to `login.defs` man page
`SYS_UID_MAX` defaults to `UID_MIN - 1`.

This information can be used to skip specific module in pam stack
based on the account type. `pam_succeed_if uid < 1000` is used at the moment
however it does not reflect changes to `login.defs`.

### DIFF
--- a/modules/pam_usertype/Makefile.am
+++ b/modules/pam_usertype/Makefile.am
@@ -1,0 +1,34 @@
+#
+# Copyright (c) 2005, 2006, 2009 Thorsten Kukuk <kukuk@suse.de>
+# Copyright (c) 2020 Red Hat, Inc.
+#
+
+CLEANFILES = *~
+MAINTAINERCLEANFILES = $(MANS) README
+
+EXTRA_DIST = README ${MANS} ${XMLS} tst-pam_usertype
+
+TESTS = tst-pam_usertype
+
+man_MANS = pam_usertype.8
+
+XMLS = README.xml pam_usertype.8.xml
+
+securelibdir = $(SECUREDIR)
+secureconfdir = $(SCONFIGDIR)
+
+AM_CFLAGS = -I$(top_srcdir)/libpam/include -I$(top_srcdir)/libpamc/include \
+	$(WARN_CFLAGS)
+AM_LDFLAGS = -no-undefined -avoid-version -module
+if HAVE_VERSIONING
+  AM_LDFLAGS += -Wl,--version-script=$(srcdir)/../modules.map
+endif
+
+securelib_LTLIBRARIES = pam_usertype.la
+pam_usertype_la_LIBADD = $(top_builddir)/libpam/libpam.la
+
+if ENABLE_REGENERATE_MAN
+noinst_DATA = README
+README: pam_usertype.8.xml
+-include $(top_srcdir)/Make.xml.rules
+endif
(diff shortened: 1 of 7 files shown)


modules/pam_usertype/Makefile.am
```diff
@@ -0,0 +1,33 @@
+#
+# Copyright (c) 2005, 2006, 2009 Thorsten Kukuk <kukuk@suse.de>
```
t8m: Add a fresh copyright notice?

modules/pam_usertype/pam_usertype.c
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    if (uid == PAM_USERTYPE_OVERFLOW_UID) {
+        /* nobody */
+        return PAM_SUCCESS;
+    }
+
+    if (uid == (uid_t)-1) {
+        pam_syslog(pamh, LOG_WARNING, "invalid uid");
+        return PAM_AUTH_ERR;
```
t8m: Either return PAM_SUCCESS here - as it is probably better to handle the invalid uid as a system user than as a regular user or return PAM_USER_UNKNOWN or some other return value and properly handle it in the pam_usertype_is_regular() as well.

thkukuk:
> @ldv-alt @thkukuk are you OK with adding it?

I'm questioning, why do we need yet another module and cannot enhance pam_succeed_if?

t8m:
Heh, I actually persuaded Pavel that rather than complicating already complicated (in terms of supported options/conditions) pam_succeed_if more it would be more clear if a new module was added. Up to now pam_succeed_if also did not read login.defs so I wanted to keep it so.

modules/pam_usertype/Makefile.am
pbrezina: Right.

pbrezina:
Patches were updated to address requested changes.

thkukuk:
On Thu, Jan 16, Tomáš Mráz wrote:

> Heh, I actually persuaded Pavel that rather than complicating already
> complicated (in terms of supported options/conditions) pam_succeed_if more it
> would be more clear if a new module was added. Up to now pam_succeed_if also
> did not read login.defs so I wanted to keep it so.

I still think that merging this in one module is more user friendly
and since reading of login.defs is done by libpam, that's not an 
argument for me.
But I'm also fine with an extra module

t8m:
Could you be please more concrete why do you think merging this into pam_succeed_if would be more user friendly? I do not see it that way. For example from the perspective of discoverability of the new functionality it is easier to find out there is a new module added rather than some existing module having a new option supported.

modules/pam_usertype/Makefile.am
```diff
@@ -0,0 +1,33 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
```
t8m: As this is derived from other makefiles, you should have kept the original copyright notice and just add the new one.

pbrezina: Done.

pbrezina:
Thank you. @thkukuk ping can you please look at Tomas' question?

thkukuk:
> LGTM. I would still like to see another review before merging. @thkukuk Did you do a review or the comment was just about separate module approach?

I only commented about the separate module approach. But the code looks fine to me.

t8m:
OK, let's merge then.

PR #168, opened 2020-01-17
MOD modules/pam_unix/passverify.c

DZH-CS:
if (curdays == spent->sp_lstchg && spent->sp_max == spent->sp_warn)
            no warn


t8m:
I do not think this is a right change - this way if the sp_warn is set to for example 1 - to warn 1 day before the actual password expiration - the warning would be given for 2 days before the actual expiration happens. Because if sp_max is for example 3 - the expiration happens on beginning of the 4th day and not on the 3rd day.

PR #170, opened 2020-01-25
configure.ac: add --enable-doc option

ffontaine:
Allow the user to disable documentation through --disable-doc (enabled
by default), this is especially useful when cross-compiling for embedded
targets

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,11 @@
 
 AUTOMAKE_OPTIONS = 1.9 gnu dist-bzip2 dist-xz check-news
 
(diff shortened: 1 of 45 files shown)


t8m:
Perhaps this should also handle the manual pages? Or there should at least be a dependency that with --disable-doc would also provide the --disable-regenerate-docu.

ffontaine:
I updated the PR as requested, --enable-doc will disable install on man pages as well as regenerate-docu

t8m:
LGTM now. Thank you for the PR. I'll merge it.

PR #171, opened 2020-01-26
Updating man pages to indicate account leakage without silent

### LINKED ISSUE #160: pam_tally2: SSH account locking reveals real users
This was asked on the StackExchange:
https://unix.stackexchange.com/questions/387880/ssh-account-locking-reveals-real-users

> On my ssh server, I'm trying to hide the existence of all the users except for one. For any user besides that one user, I will get a Password: prompt which will never allow the user to enter. That's good. But for real users, after a certain number of attempts, it will start printing "Account locked due to XX failed logins". It will not do this for non-existent users.

Looks like this is still an issue.

edneville:
Update man page to indicate silent hides account lock notices.

Closes #160.

### DIFF
--- a/modules/pam_tally/pam_tally.8.xml
+++ b/modules/pam_tally/pam_tally.8.xml
@@ -168,7 +168,7 @@
               </term>
               <listitem>
                 <para>
-                  Don't print informative messages. Without <emphasis>silent</emphasis> leaks presence of accounts on the system.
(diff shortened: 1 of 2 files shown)


t8m:
Perhaps reword it a little? `The messages printed without the silent option leak presence of accounts on the system because they are never printed for non-existing accounts.`

t8m:
Thank you for the PR. Merging now.

PR #173, opened 2020-02-03
Remove redundant header file inclusion

miztake:
There are some source code including the same header file redundantly.
We remove these redundant header file inclusion.


t8m:
Merged, thank you for the PR.

PR #176, opened 2020-02-26
pam_tty_audit: if kernel audit is disabled don't fail when user tries to log in

ikerexxe:
if kernel audit is disabled and pam audit is enabled don't fail when user tries to log in, instead syslog a warning and succeed

### DIFF
--- a/modules/pam_tty_audit/pam_tty_audit.c
+++ b/modules/pam_tty_audit/pam_tty_audit.c
@@ -351,6 +351,14 @@ pam_sm_open_session (pam_handle_t *pamh, int flags, int argc, const char **argv)
 
   fd = nl_open ();
   if (fd == -1
+      && errno == EPROTONOSUPPORT)
+    {
+      pam_syslog (pamh, LOG_WARNING, "unable to open audit socket, audit not "
+                  "supported; tty_audit skipped");
+      free (old_status);
+      return PAM_IGNORE;
+    }
+  else if (fd == -1
       || nl_send (fd, AUDIT_TTY_GET, 0, NULL, 0) != 0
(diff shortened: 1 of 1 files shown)


modules/pam_tty_audit/pam_tty_audit.c
```diff
@@ -351,6 +351,15 @@ pam_sm_open_session (pam_handle_t *pamh, int flags, int argc, const char **argv)
 
   fd = nl_open ();
   if (fd == -1
+      && errno == EPROTONOSUPPORT)
+    {
+      pam_syslog (pamh, LOG_WARNING, "warning: unable to open communication "
```
t8m: Could you please change the wording of the log message to something like: `unable to open audit socket, audit not supported; tty_audit skipped`

ikerexxe: Done

modules/pam_tty_audit/pam_tty_audit.c
```diff
@@ -351,6 +351,14 @@ pam_sm_open_session (pam_handle_t *pamh, int flags, int argc, const char **argv)
 
   fd = nl_open ();
   if (fd == -1
+      && errno == EPROTONOSUPPORT)
+    {
+      pam_syslog (pamh, LOG_WARNING, "unable to open audit socket, audit not "
+                  "supported; tty_audit skipped");
+      free (old_status);
+      return PAM_SUCCESS;
```
t8m: Thinking about it more. Please change the return value to PAM_IGNORE in this case. This way it will behave like PAM_SUCCESS with required and so but it will be possible to distinguish the error from other errors or success.

ikerexxe: Done

t8m:
Merged, thank you.

PR #177, opened 2020-02-28
Remove support for legacy xcrypt.

besser82:
Since many distributions are shipping a version of libxcrypt >= 4.0.0 as a replacement for glibc's libcrypt now, older versions of xcrypt, which could be installed in parallel, are not relevant anymore.

* configure.ac: Remove support for legacy xcrypt.
* ci/install-dependencies.sh: Likewise.
* modules/pam_cracklib/pam_cracklib.c: Likewise.
* modules/pam_pwhistory/opasswd.c: Likewise.
* modules/pam_unix/bigcrypt.c: Likewise.
* modules/pam_unix/passverify.c: Likewise.
* modules/pam_userdb/pam_userdb.c: Likewise.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -359,23 +359,17 @@ else
 fi
 AC_SUBST(LIBAUDIT)
 
-AC_CHECK_HEADERS(xcrypt.h crypt.h)
-AS_IF([test "x$ac_cv_header_xcrypt_h" = "xyes"],
-  [crypt_libs="xcrypt crypt"],
-  [crypt_libs="crypt"])
+AC_CHECK_HEADERS(crypt.h)
 
 BACKUP_LIBS=$LIBS
-AC_SEARCH_LIBS([crypt],[$crypt_libs])
+AC_SEARCH_LIBS([crypt],[crypt])
 case "$ac_cv_search_crypt" in
 	-l*) LIBCRYPT="$ac_cv_search_crypt" ;;
 	*) LIBCRYPT="" ;;
 esac
-AC_CHECK_FUNCS(crypt_r crypt_gensalt_r)
+AC_CHECK_FUNCS([crypt_r])
 LIBS=$BACKUP_LIBS
 AC_SUBST(LIBCRYPT)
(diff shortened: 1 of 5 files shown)


besser82:
@t8m: How do you feel about merging this before the 1.4.0 release?

t8m:
> @t8m: How do you feel about merging this before the 1.4.0 release?

I am not sure about this. We might need to support older systems with separate libxcrypt for a while.

ldv-alt:
Maybe it's time to remove support for legacy xcrypt.

besser82:
> > @t8m: How do you feel about merging this before the 1.4.0 release?
> 
> I am not sure about this. We might need to support older systems with separate libxcrypt for a while.

Are there any actively maintained distributions using libxcrypt < 4.0 left?

https://repology.org/project/libxcrypt/versions

From what I can see here, there are only those left with libxcrypt 2.4, that are in the best case just getting occasional security maintainace…

besser82:
> > > @t8m: How do you feel about merging this before the 1.4.0 release?
> > 
> > 
> > I am not sure about this. We might need to support older systems with separate libxcrypt for a while.
> 
> Are there any actively maintained distributions using libxcrypt < 4.0 left?
> 
> https://repology.org/project/libxcrypt/versions
> 
> From what I can see here, there are only those left with libxcrypt 2.4, that are in the best case just getting occasional security maintainace…

And those are very, very unlikely to ever get an update to most recent release of Linux-PAM.

t8m:
What is the relation to #370 ?

besser82:
> What is the relation to #370 ?

There is no relation.  The PR you've mentioned, is about some problem with uclibc-ng, which provides the `crypt()` function in its `libc.so`, and thus causes problems, if its original `crypt.h` header is replaced with one of libxcrypt >= 4.0.

Solving that is not related to this PR, as we here just remove the legacy bits for libxcrypt < 4.0.

ldv-alt (approved):
I agree the legacy xcrypt support can be safely removed.

t8m (approved):

t8m:

PR #182, opened 2020-03-03
Use cached 'crypt' library result correctly

cpackham:
Configure script incorrectly used a non-cached variable (ac_lib) in the
cached code path. This results in no -lcrypt being defined resulting in
link errors on a re-build.

Update configure.ac to use ac_cv_search_crypt (via ac_res) to setup the
correct library arguments.

Signed-off-by: Mark Wutzke <mark.wutzke@alliedtelesis.co.nz>
Signed-off-by: Chris Packham <chris.packham@alliedtelesis.co.nz>


t8m:
Merged. Thank you for the PR.

PR #194, opened 2020-03-17
pam_selinux update

cgzones:
* fall back to log to syslog if audit logging fails
* substitute legacy security_context_t type
* fix indentation


ldv-alt:
Is there any particular reason for one of three commits (pam_selinux: fall back to log to syslog if audit logging fails) to include two unrelated changes?  If not, could you separate them, please?

cgzones:
Commit separated

Issue #131, opened 2019-08-08
pam_exec: program is run before login prompt after first failed attempt


user318:
Hi,

I found out that if first attempt of login is failed, further attempts run pam_exec module before login prompt and the program receives an empty user name. I also see its stdout before I see "login:" prompt.
I looked into pam_unix code and borrowed from there a pam_get_user call. Not sure if it is completely correct, but it fixes the issue for me.
Please check the pathch in my fork repo: https://github.com/user318/linux-pam/commit/7daf2856b979c68fbefa9cb9fb2ac236076311c9

PR #195, opened 2020-03-22
pam_exec: check if user name is already available during auth

user318:
On second and futher iterations, pam_exec module is called
before the login prompt and thus command is called without
a user name. This fix asks PAM for the user name when type
is auth.
This fix addresses the issue #131.

### DIFF
--- a/modules/pam_exec/pam_exec.c
+++ b/modules/pam_exec/pam_exec.c
@@ -108,6 +108,8 @@ call_exec (const char *pam_type, pam_handle_t *pamh,
   int fds[2];
   int stdout_fds[2];
   FILE *stdout_file = NULL;
+  int retval;
+  const char *name;
 
   if (argc < 1) {
     pam_syslog (pamh, LOG_ERR,
@@ -143,6 +145,16 @@ call_exec (const char *pam_type, pam_handle_t *pamh,
 	break; /* Unknown option, assume program to execute. */
     }
 
+  /* Request user name to be available. */
+
+  retval = pam_get_user(pamh, &name, NULL);
+  if (retval != PAM_SUCCESS)
+    {
+      if (retval == PAM_CONV_AGAIN)
+        retval = PAM_INCOMPLETE;
+      return retval;
+    }
+
   if (expose_authtok == 1)
     {
       if (strcmp (pam_type, "auth") != 0)
@@ -154,7 +166,6 @@ call_exec (const char *pam_type, pam_handle_t *pamh,
       else
 	{
 	  const void *void_pass;
-	  int retval;
 
 	  retval = pam_get_item (pamh, PAM_AUTHTOK, &void_pass);
 	  if (retval != PAM_SUCCESS)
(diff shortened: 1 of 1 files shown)


user318:
Accidentally removed the branch from my repo for a short time. This cause the PR closed.

user318:
Fixed patch to compile with `-Werror` - it would be two patches with fixes, that is why I updated the whole branch with new version of the patch.

ldv-alt:
Although pam_exec can work without username, the subsequent absence of PAM_USER environment variable can indeed make the called program upset.

The mentioning of "On second and futher iterations" in the commit message is very misleading, though.  Apparently, it describes the behaviour of an application not mentioned in the commit message.

I don't mind accepting this change, but could you reword the commit message, squash commits into a single one, and rebase it on top of the current master branch, please?

user318:
Hmm, I did not do many tests, but in my setup I was getting the behaviour like I described. The first invocation of pam_exec was getting a username for some reason. So I was asked for the username in the terminal before pam_exec invocation. And after failed attempt - futher invocation was not asking for the username. But of course, I can rephrase it. :)

ldv-alt:
Most applications do not iterate this way at all.

user318:
Updated the branch with your suggestions. Please check if it is OK now.

ldv-alt:
On Mon, Mar 23, 2020 at 10:23:29AM -0700, Alexander Zubkov wrote:
> Updated the branch with your suggestions. Please check if it is OK now.

Would it make sense to extend the change beyond auth stage?
For example, what if pam_exec is invoked at another stage and nothing has requested a user name before?

I'd also suggest to replace "login" with "user name" in the commit message.

user318:
> Would it make sense to extend the change beyond auth stage? For example, what if pam_exec is invoked at another stage and nothing has requested a user name before?

Yes, you are right. I agree that it should make sence for other types too. I just used it for auth myself and did not want to meddle much.

> I'd also suggest to replace "login" with "user name" in the commit message.

OK. I'll make a new branch with these changes.

ldv-alt:
On Mon, Mar 23, 2020 at 11:07:20AM -0700, Alexander Zubkov wrote:
> > I'd also suggest to replace "login" with "user name" in the commit message.
> 
> OK. I'll make a new branch with these changes.

Thanks.  There is no need for a new branch, I suppose you can just amend this one.

user318:
> Thanks.  There is no need for a new branch, I suppose you can just amend this one.

No problem with a new branch forced push. Anyway you suggested to imrove the description.
I had to rename one further usage of the retval variable, because now without "if" I had to define retval in the top level scope of the function body and that would cause an error with "-Werror".

oxr463:
@user318 what are you using to test? I would recommend providing an example of what you mean via [`pamtester`](http://pamtester.sourceforge.net).

user318:
> @user318 what are you using to test?

@oxr463 It was agetty and me trying to login via serial console. :)

user318:
As I see pamtester takes username on its command line. And I suppose it may not be suitable to test this. Because it should be able to prompt for the user name. The idea is if the main program has not still requested the username from the user while calling the pam_exec module, that it will do it when the module asks pam for the user name.
Probably my getty version makes 3 tries to authenticate and first time it prompts for the username on its own before running pam, but during further tries it does not do it in advance, but only when the module requests it from pam.

PR #208, opened 2020-04-10
pam_unix: modernize example in manual page

topimiettinen:
According to crypt(5), md5 should not be used for new hashes. Let's
give a more modern example with sha512 and 500000 rounds (default 5000
is too weak).

Signed-off-by: Topi Miettinen <toiwoton@gmail.com>

### DIFF
--- a/modules/pam_unix/pam_unix.8.xml
+++ b/modules/pam_unix/pam_unix.8.xml
@@ -467,7 +467,7 @@ account    required   pam_unix.so
 # Change the user's password, but at first check the strength
 # with pam_cracklib(8)
 password   required   pam_cracklib.so retry=3 minlen=6 difok=3
-password   required   pam_unix.so use_authtok nullok md5
+password   required   pam_unix.so use_authtok nullok yescrypt
 session    required   pam_unix.so
       </programlisting>
     </para>


ldv-alt:
On Fri, Apr 10, 2020 at 03:49:31AM -0700, Topi Miettinen wrote:
> According to crypt(5), md5 should not be used for new hashes. Let&#39;s
> give a more modern example with sha512 and 500000 rounds (default 5000
> is too weak).

In libxcrypt >= 4.2.0 the new default method used for new hashes
is yescrypt, so I suggest using yescrypt instead.

topimiettinen:
I changed the algorithm in pam.d configuration to yescrypt:
```
password        [success=1 default=ignore]      pam_unix.so obscure yescrypt
```
Also login.defs:
```
ENCRYPT_METHOD yescrypt
```

But this does not work here (Debian bullseye):
```
passwd[35019]: pam_unix(passwd:chauthtok): unrecognized option [yescrypt]
passwd[35019]: pam_unix(passwd:chauthtok): unrecognized ENCRYPT_METHOD value [yescrypt]
```

topimiettinen:
Yescrypt doesn't work on Debian because the version of the package is v1.3.1 (latest released version) and yescrypt was added to yet unreleased master after v1.3.1 and I foolishly assumed for some reason that it was older. I'll update the example.

PR #217, opened 2020-04-27
pam_faillock: New module for locking after multiple auth failures

t8m:


ldv-alt:
Tomáš, I was going to change modules/Makefile.am to facilitate conditional builds.
While those changes are pretty much straightforward, they affect this PR slightly.

Would you mind if I merge those changes (currently in ldv/next branch) first?

t8m:
No problem with that. I'll rebase the PR after the merge.

I am also curious about the CI failures as I can't find any actual error message in the logs apart from the make failure. But there is no command that would fail.

ldv-alt:
> No problem with that. I'll rebase the PR after the merge.

Pushed.  When rebasing, see also commit 76916913fc3802ccb5348a69a56fa2bc3946439d.

modules/pam_faillock/Makefile.am
```diff
@@ -0,0 +1,46 @@
...
+# Copyright (c) 2008, 2018 Red Hat, Inc.
+# Copyright (c) 2010 Tomas Mraz <tmraz@redhat.com>
+#
+
+CLEANFILES = *~
+MAINTAINERCLEANFILES = $(MANS) README
+
+EXTRA_DIST = README $(MANS) $(XMLS) faillock.conf tst-pam_faillock
+
+man_MANS = pam_faillock.8 faillock.8 faillock.conf.5
```
ldv-alt, line 12: Nowadays it has to be
```if HAVE_DOC
man_MANS = pam_faillock.8 faillock.8 faillock.conf.5
endif```

File: modules/pam_faillock/Makefile.am [hunk 2]
```diff
@@ -0,0 +1,46 @@
...
+
+TESTS = tst-pam_faillock
+
+securelibdir = $(SECUREDIR)
+secureconfdir = $(SCONFIGDIR)
+
+noinst_HEADERS = faillock.h
+
+faillock_CFLAGS = -I$(top_srcdir)/libpam/include -I$(top_srcdir)/libpamc/include @PIE_CFLAGS@
+pam_faillock_la_CFLAGS = -I$(top_srcdir)/libpam/include -I$(top_srcdir)/libpamc/include
```
ldv-alt, line 23: $(WARN_CFLAGS) is missing, I suggest adding common AM_CFLAGS and using it in other *_CFLAGS.

t8m:
Fixup commit pushed

modules/pam_faillock/pam_faillock.8.xml
```diff
@@ -0,0 +1,322 @@
...
+                  The <emphasis>authfail</emphasis> argument must be used when the module
+                  is called after the modules which determine the authentication outcome,
+                  failed. Unless the user is already blocked due to previous authentication
+                  failures, the module will record the failure into the appropriate user
+                  tally file.
+                </para>
+                <para>
+                  The <emphasis>authsucc</emphasis> argument must be used when the module
+                  is called after the modules which determine the authentication outcome,
+                  succeded. Unless the user is already blocked due to previous authentication
```
ldv-alt, line 116: s/succeded/succeeded/

modules/pam_faillock/pam_faillock.8.xml
```diff
@@ -0,0 +1,322 @@
...
+  </refsect1>
+
+  <refsect1 id='pam_faillock-return_values'>
+    <title>RETURN VALUES</title>
+    <variablelist>
+      <varlistentry>
+        <term>PAM_AUTH_ERR</term>
+        <listitem>
+          <para>
+            A invalid option was given, the module was not able
```
ldv-alt, line 153: s/A invalid/An invalid/

modules/pam_faillock/pam_faillock.8.xml
```diff
@@ -0,0 +1,322 @@
...
+      the user. This allows <emphasis remap='B'>pam_faillock.so</emphasis> module
+      to work correctly when it is called from a screensaver.
+    </para>
+    <para>
+      Note that using the module in <option>preauth</option> without the
+      <option>silent</option> option specified in <filename>/etc/security/faillock.conf</filename>
+      or with <emphasis>requisite</emphasis> control field leaks an information about
+      existence or non-existence of an user account in the system because
+      the failures are not recorded for the unknown users. The message
+      about the user account being locked is never displayed for nonexisting
```
ldv-alt, line 199: s/nonexisting/non-existing/

modules/pam_faillock/pam_faillock.8.xml
```diff
@@ -0,0 +1,322 @@
...
+    <title>EXAMPLES</title>
+    <para>
+      Here are two possible configuration examples for <filename>/etc/pam.d/login</filename>.
+      They make <emphasis>pam_faillock</emphasis> to lock the account after 4 consecutive
+      failed logins during the default interval of 15 minutes. Root account will be locked
+      as well. The accounts will be automatically unlocked after 20 minutes.
+    </para>
+    <para>
+      In the first example the module is called only in the <emphasis>auth</emphasis>
+      phase and the module does not print any information about the account blocking
```
ldv-alt, line 215: s/blocking by/being blocked by/

modules/pam_faillock/pam_faillock.8.xml
```diff
@@ -0,0 +1,322 @@
...
+      by <emphasis>pam_faillock</emphasis>. The <emphasis>preauth</emphasis> call can
+      be added to tell the user that his login is blocked by the module and also to abort
```
ldv-alt, line 217: s/tell the user that his login is blocked/tell users that their logins are blocked/

modules/pam_faillock/pam_faillock.8.xml
```diff
@@ -0,0 +1,322 @@
...
+password required       pam_unix.so shadow
+session  required       pam_selinux.so close
+session  required       pam_loginuid.so
+session  required       pam_unix.so
+session  required       pam_selinux.so open
+    </programlisting>
+    <para>
+      In the second example the module is called both in the <emphasis>auth</emphasis>
+      and <emphasis>account</emphasis> phases and the module gives the authenticating
+      user message when the account is locked if <option>silent</option> option is not
```
ldv-alt, line 251: s/gives the authenticating user message when/informs the authenticating user when/

modules/pam_faillock/faillock.conf.5.xml
```diff
@@ -0,0 +1,243 @@
...
+                <option>unlock_time=<replaceable>n</replaceable></option>
+              </term>
+              <listitem>
+                <para>
+                  The access will be reenabled after
+                  <replaceable>n</replaceable> seconds after the lock out.
+                  The value 0 has the same meaning as value
+                  <emphasis>never</emphasis> - the access
+                  will not be reenabled without resetting the faillock
+                  entries by the <citerefentry><refentrytitle>faillock</refentrytitle><manvolnum>8</manvolnum></citerefentry> command.
```
ldv-alt, line 133: s/reenabled/re-enabled/g

modules/pam_faillock/faillock.conf.5.xml
```diff
@@ -0,0 +1,243 @@
...
+                </para>
+                <para>
+                  Note that the default directory that <emphasis>pam_faillock</emphasis>
+                  uses is usually cleared on system boot so the access will be also reenabled
+                  after system reboot. If that is undesirable a different tally directory
+                  must be set with the <option>dir</option> option.
+                </para>
+                <para>
+                  Also note that it is usually undesirable to permanently lock
+                  out the users as they can become easily a target of denial of service
```
ldv-alt, line 144: s/the users/users/

modules/pam_faillock/faillock.8.xml
```diff
@@ -0,0 +1,123 @@
...
+      failed authentication attempts per user during a specified interval
+      and locks the account in case there were more than
+      <replaceable>deny</replaceable> consecutive failed authentications.
+      It stores the failure records into per-user files in the tally
+      directory.
+    </para>
+    <para>
+      The <command>faillock</command> command is an application which
+      can be used to examine and modify the contents of the
+      the tally files. It can display the recent failed authentication
```
ldv-alt, line 48: s/the the/the/

modules/pam_faillock/faillock.conf
```diff
@@ -0,0 +1,62 @@
...
+#
+# The length of the interval during which the consecutive
+# authentication failures must happen for the user account
+# lock out is <replaceable>n</replaceable> seconds.
+# The default is 900 (15 minutes).
+# fail_interval = 900
+#
+# The access will be reenabled after n seconds after the lock out.
+# The value 0 has the same meaning as value `never` - the access
+# will not be reenabled without resetting the faillock
```
ldv-alt, line 42: s/reenabled/re-enabled/g

modules/pam_faillock/faillock.h
```diff
@@ -0,0 +1,73 @@
...
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/*
+ * faillock.h - authentication failure data file record structure
+ *
+ * Each record in the file represents an instance of login failure of
+ * the user at the recorded time
```
ldv-alt, line 40: I'd like to see a terminating period at the end of the sentence.

modules/pam_faillock/faillock.h
```diff
@@ -0,0 +1,73 @@
...
+
+#ifndef _FAILLOCK_H
+#define _FAILLOCK_H
+
+#include <stdint.h>
+#include <sys/types.h>
+
+#define TALLY_STATUS_VALID     0x1       /* the tally file entry is valid */
+#define TALLY_STATUS_RHOST     0x2       /* the source is rhost */
+#define TALLY_STATUS_TTY       0x4       /* the source is tty - if both TALLY_FLAG_RHOST and TALLY_FLAG_TTY are not set the source is service */
```
ldv-alt, line 52: This line is way too long.

modules/pam_faillock/main.c
```diff
@@ -0,0 +1,232 @@
...
+	fd = open_tally(opts->dir, user, pwd != NULL ? pwd->pw_uid : 0, 0);
+
+	if (fd == -1) {
+		if (errno == ENOENT) {
+			return 0;
+		}
+		else {
+			fprintf(stderr, "%s: Error opening the tally file for %s:",
+				opts->progname, user);
+			perror(NULL);
```
ldv-alt, line 125: %m would look nicer than perror(NULL)

modules/pam_faillock/main.c
```diff
@@ -0,0 +1,232 @@
...
+#ifdef HAVE_LIBAUDIT
+		}
+		if ((audit_fd=audit_open()) >= 0) {
+
+			if (pwd != NULL) {
+				audit_log_acct_message(audit_fd, AUDIT_USER_MGMT, NULL,
+					"faillock-reset", NULL, pwd->pw_uid, NULL, NULL, NULL, rv == 0);
+			}
+			close(audit_fd);
+		}
```
ldv-alt, line 148: no need to bother with audit_open if pwd == NULL?

t8m:
@ldv-alt I've hopefully addressed all your comments. OK to merge?

ldv-alt:
On Tue, Apr 28, 2020 at 04:00:11AM -0700, Tomáš Mráz wrote:
> @ldv-alt I've hopefully addressed all your comments. OK to merge?

Wait a bit, I haven't looked into pam_faillock.c yet! :)

modules/pam_faillock/main.c
```diff
@@ -0,0 +1,232 @@
...
+#include <dirent.h>
+#include <errno.h>
+#include <pwd.h>
+#include <time.h>
+#include <sys/types.h>
+#include <unistd.h>
+#ifdef HAVE_LIBAUDIT
+#include <libaudit.h>
+
+#define AUDIT_NO_ID     ((unsigned int) -1)
```
ldv-alt, line 51: -1U ?

t8m: What's wrong with the cast? -1U looks weird to me although it is also correct.

ldv-alt:
On Tue, Apr 28, 2020 at 06:22:27AM -0700, Tomáš Mráz wrote:
> +#define AUDIT_NO_ID     ((unsigned int) -1)
> 
> What's wrong with the cast? -1U looks weird to me although it is also correct.

I prefer -1U, it's the most concise form I know.

t8m:
I am going to keep the cast. I do not really see it wrong or confusing or anything.

modules/pam_faillock/pam_faillock.c
```diff
@@ -0,0 +1,771 @@
...
+			oldtime = tallies->records[i].time;
+			oldest = i;
+		}
+		if (opts->flags & FAILLOCK_FLAG_UNLOCKED ||
+			opts->now - tallies->records[i].time >= opts->fail_interval ) {
+			tallies->records[i].status &= ~TALLY_STATUS_VALID;
+		} else {
+			++failures;
+		}
+	}
```
ldv-alt, line 549: Assuming that ```tallies->records[i].time >= 0```,
```(tallies->records[i].time < oldtime)``` is always false,
hence ```oldtime == oldest == 0``` after this loop.

What was the intent behind this piece of code?

File: modules/pam_faillock/pam_faillock.c [hunk 19]
```diff
@@ -0,0 +1,771 @@
...
+	struct options *opts,
+	const char *cfgfile
+);
+
+void set_conf_opt(
+	pam_handle_t *pamh,
+	struct options *opts,
+	const char *name,
+	const char *value
+);
```
ldv-alt, line 110: Since these two functions and all their users are in the same file, these functions should rather be static.

modules/pam_faillock/pam_faillock.c
```diff
@@ -0,0 +1,771 @@
...
+static void
+args_parse(pam_handle_t *pamh, int argc, const char **argv,
+		int flags, struct options *opts)
+{
+	int i;
+	int rv;
+	memset(opts, 0, sizeof(*opts));
+
+	opts->dir = strdup(FAILLOCK_DEFAULT_TALLYDIR);
+	opts->conf = FAILLOCK_DEFAULT_CONF;
```
ldv-alt, line 121: This is the only place where struct options.conf is assigned,
which effectively means that no files besides FAILLOCK_DEFAULT_CONF are supported.
Do we need struct options.conf?

modules/pam_faillock/pam_faillock.c
```diff
@@ -0,0 +1,771 @@
...
+read_config_file(pam_handle_t *pamh, struct options *opts, const char *cfgfile)
+{
+	FILE *f;
+	char linebuf[FAILLOCK_CONF_MAX_LINELEN+1];
+
+	f = fopen(cfgfile, "r");
+	if (f == NULL) {
+		/* ignore non-existent default config file */
+		if (errno == ENOENT && strcmp(cfgfile, FAILLOCK_DEFAULT_CONF) == 0)
+			return 0;
```
ldv-alt, line 183: Apparently, cfgfile is always FAILLOCK_DEFAULT_CONF.

modules/pam_faillock/pam_faillock.c
```diff
@@ -0,0 +1,771 @@
...
+	opts->conf = FAILLOCK_DEFAULT_CONF;
+	opts->deny = 3;
+	opts->fail_interval = 900;
+	opts->unlock_time = 600;
+	opts->root_unlock_time = MAX_TIME_INTERVAL+1;
+
+	if ((rv=read_config_file(pamh, opts, opts->conf)) != PAM_SUCCESS) {
+		pam_syslog(pamh, LOG_DEBUG,
+					"Configuration file missing");
+	}
```
ldv-alt, line 130: This is confusing because read_config_file returns 0 in case of errno == ENOENT.

modules/pam_faillock/pam_faillock.c
```diff
@@ -0,0 +1,771 @@
...
+
+	if (read_tally(tfd, tallies) != 0) {
+		pam_syslog(pamh, LOG_ERR, "Error reading the tally file for %s: %m", opts->user);
+		return PAM_SYSTEM_ERR;
+	}
+
+	if (opts->is_admin && !(opts->flags & FAILLOCK_FLAG_DENY_ROOT)) {
+		return PAM_SUCCESS;
+	}
```
ldv-alt, line 446: Why do we check is_admin so late?
Shouldn't we check this right after get_pam_user call?

modules/pam_faillock/pam_faillock.c
```diff
@@ -0,0 +1,771 @@
...
+	uint64_t oldtime;
+	const void *source = NULL;
+
+	if (*fd == -1) {
+		*fd = open_tally(opts->dir, opts->user, opts->uid, 1);
+	}
+	if (*fd == -1) {
+		if (errno == EACCES) {
+			return PAM_SUCCESS;
+		}
```
ldv-alt, line 529: In check_tally there is a similar check where ENOENT is ignored the same way as EACCES.
In write_tally only EACCES is ignored. Why?

modules/pam_faillock/pam_faillock.c
t8m: The intent was to find the oldest record to potentially overwrite it with the new data if the record is not valid anymore. This apparently will not happen and the record 0 will be overwritten instead  if it is not valid anymore. It is not critical for the pam_faillock work however it should be fixed.

modules/pam_faillock/pam_faillock.c
t8m: I'd prefer keeping it if anybody wants to implement option to load non-default config file.

modules/pam_faillock/pam_faillock.c
t8m: I'd prefer keeping it as preparation for non-default conf option implementation.

modules/pam_faillock/pam_faillock.c
t8m: It would not if the non-default conf option was implemented. I can submit it in a separate PR later.

modules/pam_faillock/pam_faillock.c
t8m: No, we want to keep the records updated for admin as well. We just do not block him.

modules/pam_faillock/pam_faillock.c
t8m: When reading there might be no tally file for an user yet. So we need to ignore ENOENT. When writing we however want to write a tally if we are allowed by permissions. The file will be created if it does not exist so there is no point in ignoring ENOENT.

t8m:
So I believe I've addressed your comments by fixing or a reply to the comment.

modules/pam_faillock/pam_faillock.c
ldv-alt: If a config option is ever supported, would it make sense to skip the config file specified in that option only if it is the same as FAILLOCK_DEFAULT_CONF?

t8m: Yes, the intent is to ignore the default faillock.conf if it is missing because that can be intentional - the reason is that the faillock.conf support was added only in later versions of the module - before that there was no such file and only pam module arguments were used to configure it. However if there was a non-default config file option and it was specified, it seems to be appropriate to error out if the file is missing.

ldv-alt: In that case ```strcmp(cfgfile, FAILLOCK_DEFAULT_CONF)``` wouldn't be very helpful.
If opts->conf was initialized to NULL, then cfgfile == NULL would mean the default config.
btw, does read_config_file really need its cfgfile parameter?

modules/pam_faillock/pam_faillock.c
```diff
@@ -0,0 +1,774 @@
...
+			   PATH_PASSWD);
+		return -1;
+	}
+
+	for (;;) {
+		errn = fgetpwent_r(fp, &pw, buf, sizeof (buf), &pwp);
+		if (errn == ERANGE) {
+			pam_syslog(pamh, LOG_WARNING, "%s contains very long lines; corrupted?",
+				   PATH_PASSWD);
+			/* we can continue here as next call will read further */
```
ldv-alt, line 359: The comment is misleading, we cannot proceed with the next fgetpwent_r call as if nothing has happened.

If fgetpwent_r returned ERANGE, it means either read error or sizeof(buf) bytes were read without a terminating newline. In the first case any subsequent fgetpwent_r call is likely to produce the same error, in the second case a subsequent fgetpwent_r call will produce garbage.

t8m: OK, let's just make it an error then.

t8m:
So I believe I've addressed all the comments now. I've added also the conf option support so we do not carry a dead code.

t8m:
@ldv-alt any further comments?

modules/pam_faillock/Makefile.am
```diff
@@ -0,0 +1,50 @@
+#
+# Copyright (c) 2005, 2006, 2007, 2009 Thorsten Kukuk <kukuk@thkukuk.de>
+# Copyright (c) 2008, 2018, 2020 Red Hat, Inc.
+# Copyright (c) 2010 Tomas Mraz <tmraz@redhat.com>
+#
+
+CLEANFILES = *~
+MAINTAINERCLEANFILES = $(MANS) README
+
+EXTRA_DIST = README $(MANS) $(XMLS) faillock.conf tst-pam_faillock
```
ldv-alt, line 10: This EXTRA_DIST can be simplified to
```EXTRA_DIST = $(DATA) $(MANS) $(XMLS) $(TESTS)```
similar to recent cleanups in other makefiles.

File: modules/pam_faillock/Makefile.am [hunk 27]
```diff
@@ -0,0 +1,50 @@
...
+
+securelib_LTLIBRARIES = pam_faillock.la
+sbin_PROGRAMS = faillock
+
+pam_faillock_la_SOURCES = pam_faillock.c faillock.c
+faillock_SOURCES = main.c faillock.c
+
+if ENABLE_REGENERATE_MAN
+noinst_DATA = README
+README: pam_faillock.8.xml
```
ldv-alt, line 48: When rebasing to master, this
```README: pam_faillock.8.xml```
rule can be omitted (commit 1886b602 made it redundant).

ldv-alt:
> @ldv-alt any further comments?

Let's merge it. :)

t8m:
Merged manually.

Issue #224, opened 2020-05-12
[RFE]  PAM-level integration of prctl(PR_SET_NO_NEW_PRIVS, ...)


vcaputo:
I want to enforce this prctl() at login time in the same spirit of things found in /etc/security/limits.conf.

Would a PR adding a no_new_privs option to modules/pam_limits.c be considered for acceptance upstream?

LMK if this is already supported and I'm just missing something.  What I'm interested in is a user/group-oriented way of forcing PR_SET_NO_NEW_PRIVS @ login time.

t8m:
Do you mean a new item for limits.conf? In that case it should be probably named nonewprivs without underscores to align with the other items. I suppose it should be somewhat acceptable for pam_limits instead of creating a completely new module just for this purpose and duplicating all the user/group matching there.
@ldv-alt @thkukuk  What is your opinion?

vcaputo:
> Do you mean a new item for limits.conf?

Yep!  No opinion here on naming, with or without underscores, doesn't matter to me.  Seems like a trivial patch.

ldv-alt:
On Tue, May 12, 2020 at 04:37:20AM -0700, Tomáš Mráz wrote:
> Do you mean a new item for limits.conf? In that case it should be probably named nonewprivs without underscores to align with the other items. I suppose it should be somewhat acceptable for pam_limits instead of creating a completely new module just for this purpose and duplicating all the user/group matching there.
> @ldv-alt @thkukuk  What is your opinion?

I concur.

PR #225, opened 2020-05-12
modules/pam_limits: add support for nonewprivs

vcaputo:
Expose prctl(PR_SET_NO_NEW_PRIVS) as "nonewprivs" item

The valid values are a boolean toggle 0/1 to keep semi-consistent
with the other numeric limits.  It's slightly awkward as this is
an oddball relative to the other items in pam_limits but outside
of the item value itself this does seem at home in pam_limits.

Fixes https://github.com/linux-pam/linux-pam/issues/224

### DIFF
--- a/modules/pam_limits/limits.conf.5.xml
+++ b/modules/pam_limits/limits.conf.5.xml
@@ -228,6 +228,13 @@
               </listitem>
             </varlistentry>
             <varlistentry>
+              <term><option>nonewprivs</option></term>
+              <listitem>
+                <para>value of 0 or 1; if set to 1 disables acquiring new
+                  privileges by invoking prctl(PR_SET_NO_NEW_PRIVS)</para>
+              </listitem>
+            </varlistentry>
+            <varlistentry>
               <term><option>priority</option></term>
               <listitem>
                 <para>the priority to run user process with (negative
@@ -274,7 +281,8 @@
     <para>
       All items support the values <emphasis>-1</emphasis>,
       <emphasis>unlimited</emphasis> or <emphasis>infinity</emphasis> indicating no limit,
-      except for <emphasis remap='B'>priority</emphasis> and <emphasis remap='B'>nice</emphasis>.
+      except for <emphasis remap='B'>priority</emphasis>, <emphasis remap='B'>nice</emphasis>,
+      and <emphasis remap='B'>nonewprivs</emphasis>.
     </para>
     <para>
       If a hard limit or soft limit of a resource is set to a valid value,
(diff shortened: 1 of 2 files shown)


vcaputo:
I haven't tested this and am not familiar with the linux-pam coding style or conventions at all, but it seems like a reasonable start...

vcaputo:
Ok, my first half-asleep untested attempt at this was comically misguided as I tried sticking the flag in the module-wide ctrl mask, which the eventual testing quickly revealed was doing nothing useful!

It's since been rewritten to be a new item as appropriate.  I gotta say I'm a bit mystified by what's going on in the source code with regards to indentation and whitespace in general.  I have vim setup to visually distinguish tabs from spaces and this file is all over the place.

Some of my changes landed in a region where just tabs are being used, so I tried to be consistent with the surrounding lines.  Elsewhere the changes fell where a mix of 4+4 cumulative 8-space indentations are compressed into a single tab with 4-space fractional tabs filled out as spaces.  I tried to replicate that where appropriate.

With this PR I'm now able put something like the following in /etc/limits/security.conf:
```
web             -       nonewprivs      1
```

and the web user upon login loses the ability to acquire new priveleges.  Running the suid `su` command results in the following:
```
web@iridesce:~$ su
unix_chkpwd[2613]: check pass; user unknown
Password: 
unix_chkpwd[2614]: check pass; user unknown
unix_chkpwd[2614]: password check failed for user (root)
su: Authentication failure
web@iridesce:~$ 
```
Even with the correct root password on hand since `su` lacks the privileges to do the necessary things.  I'm very happy with this, as I use a separate user in a container for my web browsing, and now this user's processes couldn't gain root via su/sudo even if the session somehow managed to sniff my root password (I do still use Xorg after all).

There might be need to bikeshed over the item value, I used a boolean-style 0/1 to disable/enable the nonewprivs.  It seemed like just having nonewprivs present might be appropriate to enable it, but that'd be inconsistent with the other pam_limits items all having a value column, and I guess it's nice to allow retaining existing lines while simply flipping the 1/0 to control its effect, though commenting could achieve the same effect.  *shrug*

Promoting from draft to PR, PTAL, and thanks.

vcaputo:
@ldv-alt Force-pushed w/updated manual, but I assume you mean `pam_limits/limits.conf.5.xml`, thanks!

modules/pam_limits/limits.conf.5.xml
```diff
@@ -227,6 +227,13 @@
                   user with <emphasis>uid=0</emphasis>)</para>
               </listitem>
             </varlistentry>
+            <varlistentry>
+              <term><option>nonewprivs</option></term>
+              <listitem>
+                <para>disable acquiring new privileges by invoking
```
t8m, line 233: please add `value of 0 or 1; if set to 1` ....

vcaputo: Done.

For some reason I'm not receiving email notifications from github for these comments, otherwise I would have updated this quicker.

vcaputo:
Neat, thanks all!

PR #230, opened 2020-05-29
Update pam_setquota.c

jmoellers:
A colleague reported this bug description:
So this pam_setquota module iterates over all mounted file systems using
`setmntent()` and `getmntent()`. It tries to find the longest match of a file
system mounted on /home/$USER or above (except when the fs=/some/path
parameter is passed to the pam module).

The thing is that /home/$USER is owned by the unprivileged user. And there
exist tools like fusermount from libfuse which is by default installed
setuid-root for everybody. fusermount allows to mount a FUSE file system using
an arbitrary "source device name" as the unprivileged user.

Thus considering the following use case:

1) there is only the root file system (/) or a file system is mounted on
   /home, but not on /home/$USER.
2) the attacker mounts a fake FUSE file system over its own home directory:

  ```
  user $ export _FUSE_COMMFD=0
  user $ fusermount $HOME -ononempty,fsname=/dev/sda1
  ```

  This will result in a mount entry in /proc/mounts looking like this:

  ```
  /dev/sda1 on /home/$USER type fuse (rw,nosuid,nodev,relatime,user_id=1000,group_id=100)
  ```
3) when the attacker now logs in with pam_setquota configured then
   pam_setquota will identify /dev/sda1 and the file system where to apply the
   user's quota on.

As a result an unprivileged user has full control over onto which block device
the quota is applied.
-eoq-
If the user's $HOME is on a separate partition, setting a quota on the user's $HOME does not really make sense, so this patch checks whether this is the case and does not set a quota, preventing the above mentioned bug as a side-effect (or vice-versa).

### DIFF
--- a/modules/pam_setquota/pam_setquota.c
+++ b/modules/pam_setquota/pam_setquota.c
@@ -171,7 +171,7 @@ pam_sm_open_session(pam_handle_t *pamh, int flags UNUSED,
 		    int argc, const char **argv)
 {
   int retval;
-  char *val, *mntdevice = NULL;
+  char *val, *mntdevice = NULL, *mntpoint = NULL;
   const void *user;
   const struct passwd *pwd;
   struct pam_params param = {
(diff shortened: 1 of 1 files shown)


t8m:
I believe this check could be very easily done by just dropping the `s[0] == '\0' condition at the line 278. This effectively means that mountpoints that completely match the home dir path are skipped.

ldv-alt:
On Fri, May 29, 2020 at 08:20:06AM -0700, Tomáš Mráz wrote:
> I believe this check could be very easily done by just dropping the `s[0] == '\0' condition at the line 278. This effectively means that mountpoints that completely match the home dir path are skipped.

Yes.

ldv-alt:
What do you think about draft commit 27ded89 mentioned above?

t8m:
I am fine with the @ldv-alt 's commit. The original one should be of course dropped.

jmoellers:
On 03.06.20 08:50, Tomáš Mráz wrote:
>
> I am fine with the @ldv-alt <https://github.com/ldv-alt> 's commit.
> The original one should be of course dropped.
>
I agree.

Josef

-- 
SUSE Software Solutions Germany GmbH
Maxfeldstr. 5
90409 Nürnberg
Germany

(HRB 36809, AG Nürnberg)
Geschäftsführer: Felix Imendörffer

PR #234, opened 2020-06-11
Fix build with musl libc.

jameshilliard:


ldv-alt:
I think we shouldn't emulate fgetpwent_r.

jameshilliard:
I guess #234 probably is a better fix.

ldv-alt:
On Thu, Jun 11, 2020 at 10:09:59AM -0700, James Hilliard wrote:
> I guess #234 probably is a better fix.

Sorry?

jameshilliard:
Oh, sorry, meant #237.

PR #237, opened 2020-06-11
fix build on musl

### LINKED ISSUE #236: pam_faillock fails to build on musl
pam_faillock fails to build because it uses the non standard fgetpwent_r which is undefined on musl:

```
libtool: link: /home/naourr/work/instance-0/output-1/host/bin/arm-buildroot-linux-musleabihf-gcc -shared  -fPIC -DPIC  .libs/pam_faillock.o .libs/faillock.o   -Wl,-rpath -Wl,/home/naourr/work/instance-0/output-1/build/linux-pam-1.4.0/libpam/.libs ../../libpam/.libs/libpam.so -lintl  -Os -Wl,--version-script=./../modules.map -Wl,--as-needed -Wl,--no-undefined -Wl,-O1   -Wl,-soname -Wl,pam_faillock.so -o .libs/pam_faillock.so
/home/naourr/work/instance-0/output-1/host/lib/gcc/arm-buildroot-linux-musleabihf/8.4.0/../../../../arm-buildroot-linux-musleabihf/bin/ld: .libs/pam_faillock.o: in function `check_local_user':
pam_faillock.c:(.text+0x168): undefined reference to `fgetpwent_r'
collect2: error: ld returned 1 exit status
```

Would it possible to replace this function by the standard (but non reentrant) fgetpwent?

The full log is available here:
http://autobuild.buildroot.org/results/226/2262a4d8b76899b60bbd6bf34d739d818aba3a02/build-end.log

ffontaine:
Rename check_user_in_passwd from pam_localuser.c to
pam_modutil_check_user_in_passwd and use it in pam_faillock.c instead of
fgetpwent_r which is not available on musl

Fix #236

Fixes:
 - http://autobuild.buildroot.org/results/0432736ffee376dd84757469434a4bbcfdcdaf4b

### DIFF
--- a/modules/pam_faillock/pam_faillock.c
+++ b/modules/pam_faillock/pam_faillock.c
@@ -348,42 +348,7 @@ set_conf_opt(pam_handle_t *pamh, struct options *opts, const char *name, const c
 static int
 check_local_user (pam_handle_t *pamh, const char *user)
 {
-	struct passwd pw, *pwp;
-	char buf[16384];
-	int found = 0;
-	FILE *fp;
-	int errn;
-
-	fp = fopen(PATH_PASSWD, "r");
-	if (fp == NULL) {
-		pam_syslog(pamh, LOG_ERR, "unable to open %s: %m",
-			   PATH_PASSWD);
-		return -1;
-	}
-
-	for (;;) {
-		errn = fgetpwent_r(fp, &pw, buf, sizeof (buf), &pwp);
-		if (errn == ERANGE) {
-			pam_syslog(pamh, LOG_WARNING, "%s contains very long lines; corrupted?",
-				   PATH_PASSWD);
-			break;
-		}
-		if (errn != 0)
-			break;
-		if (strcmp(pwp->pw_name, user) == 0) {
-			found = 1;
-			break;
-		}
-	}
-
-	fclose (fp);
-
-	if (errn != 0 && errn != ENOENT) {
-		pam_syslog(pamh, LOG_ERR, "unable to enumerate local accounts: %m");
-		return -1;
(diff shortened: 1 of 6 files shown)


modules/pam_faillock/pam_faillock.c
```diff
@@ -348,42 +348,7 @@ set_conf_opt(pam_handle_t *pamh, struct options *opts, const char *name, const c
...
-
-	fclose (fp);
-
-	if (errn != 0 && errn != ENOENT) {
-		pam_syslog(pamh, LOG_ERR, "unable to enumerate local accounts: %m");
-		return -1;
-	} else {
-		return found;
-	}
+	return pam_modutil_check_user_in_passwd(pamh, user, NULL);
```
ldv-alt, line 351: As I mentioned earlier, the semantics of pam_modutil_check_user_in_passwd return value differs from the semantics of check_local_user.  Probably the closest equivalent would be either
```
return pam_modutil_check_user_in_passwd(pamh, user, NULL) == PAM_SUCCESS;
```
or
```
return pam_modutil_check_user_in_passwd(pamh, user, NULL) != PAM_PERM_DENIED;
```

ldv-alt:
I'd probably split this commit in two, but besides that, I don't have any more comments.
@t8m , what do you think about this PR?

t8m:
I think the source file name pam_modutil_check_user_in_passwd.c is too long and maybe even too specific in case we would like to add some related function to it. Could you please invent some shorter and also perhaps more general one? Otherwise I am OK with this.

ldv-alt:
pam_modutil_check_user.c ?

PR #240, opened 2020-06-15
pam_faillock: change /run/faillock/$USER permissions to 0660

ikerexxe:
Nowadays, /run/faillock/$USER files have user:root ownership and 0600 permissions. This forces the process that writes to these files to have CAP_DAC_OVERRIDE capabilites. Just by changing the permissions to 0660 the capability can be removed, which leads to a more secure system.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1661822

### DIFF
--- a/modules/pam_faillock/faillock.c
+++ b/modules/pam_faillock/faillock.c
@@ -76,7 +76,7 @@ open_tally (const char *dir, const char *user, uid_t uid, int create)
 		flags |= O_CREAT;
 	}
 
-	fd = open(path, flags, 0600);
+	fd = open(path, flags, 0660);
 
 	free(path);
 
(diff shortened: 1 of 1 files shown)


modules/pam_faillock/faillock.c
```diff
@@ -76,13 +76,22 @@ open_tally (const char *dir, const char *user, uid_t uid, int create)
...
 
 	if (fd != -1) {
 		struct stat st;
 
+		/*
+		 * If umask is set to 022, as will probably in most systems, then the
+		 * group will not be able to write to the file. So, change the file
+		 * permissions just in case.
+		 */
+		if (fchmod(fd, 0660)) {
```
t8m, line 91: As this is auxiliary call I'd suggest using ignore_return() instead of returning error.
Perhaps also add note to the comment that the user will be set to the real user uid and the group would be root so later root processes will require CAP_DAC_OVERRIDE.

modules/pam_faillock/faillock.c
```diff
@@ -76,13 +76,23 @@ open_tally (const char *dir, const char *user, uid_t uid, int create)
...
+		 * permissions just in case.
+		 * Note: owners of this file are user:root, so if the permissions are
+		 * not changed the root process writing to this file will require
+		 * CAP_DAC_OVERRIDE.
+		 */
+		ignore_return(fchmod(fd, 0660));
+
 		while (flock(fd, LOCK_EX) == -1 && errno == EINTR);
 		if (fstat(fd, &st) == 0) {
 			if (st.st_uid != uid) {
```
ldv-alt, line 98: Should this fchmod be made along with fchown below?

t8m: Yes, perhaps it could also be checked whether the group write permission isn't set already as we have the information readily available.

ikerexxe: I have changed it based on your comments.

PR #242, opened 2020-06-16
match the runtime until password prompt of existing and non-existing users

ikerexxe:
Taking a look at the time for the password prompt to appear it was possible to determine if a user existed in a system. Solved it by matching the runtime until the password prompt was shown by always checking the password hash for an existing and a non-existing user.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1629598

### DIFF
--- a/modules/pam_unix/support.c
+++ b/modules/pam_unix/support.c
@@ -601,6 +601,8 @@ _unix_blankpasswd (pam_handle_t *pamh, unsigned long long ctrl, const char *name
 	char *salt = NULL;
 	int daysleft;
 	int retval;
+	int execloop = 1;
+	int nonexistent = 1;
 
 	D(("called"));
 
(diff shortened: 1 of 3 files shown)


modules/pam_unix/support.c
```diff
@@ -624,14 +626,31 @@ _unix_blankpasswd (pam_handle_t *pamh, unsigned long long ctrl, const char *name
...
-	if (retval == PAM_UNIX_RUN_HELPER) {
-		/* salt will not be set here so we can return immediately */
-		if (_unix_run_helper_binary(pamh, NULL, ctrl, name) == PAM_SUCCESS)
-			return 1;
-		else
-			return 0;
+		if (retval == PAM_UNIX_RUN_HELPER) {
+			execloop = 0;
+			if(nonexistent) {
+				get_pwd_hash(pamh, "non_existent", &pwd, &salt);
```
t8m, line 641: Perhaps use something more weird to avoid a collision - this looks like a user name that some sysadmin could create to test things for example. Something like non_pam_unix_existent or so.

modules/pam_usertype/pam_usertype.c
```diff
@@ -139,8 +139,11 @@ pam_usertype_get_uid(struct pam_usertype_opts *opts,
                        "error retrieving information about user %s", username);
         }
 
+        pam_modutil_getpwnam(pamh, "root");
+
         return PAM_USER_UNKNOWN;
     }
+    pam_modutil_getpwnam(pamh, "non_existent");
```
t8m, line 146: same here

ikerexxe:
I've been checking it and pam_localuser doesn't use getpwnam_r(), so I think it's safe on that side. As for pam_succeed_if, it uses the API but it's not used as an authentication stack module in Fedora/RHEL, thus it will not impact on password prompt times. Though, I'm not sure about other distributions.

ldv-alt:
Yes, pam_localuser also needs some treatment.  I'll prepare a patch for pam_modutil_check_user_in_passwd.

ldv-alt:
On Tue, Jun 16, 2020 at 05:54:39AM -0700, Tomáš Mráz wrote:
> +				get_pwd_hash(pamh, "non_existent", &pwd, &salt);
> 
> Perhaps use something more weird to avoid a collision - this looks like a user name that some sysadmin could create to test things for example. Something like non_pam_unix_existent or so.

What about something containing a colon, e.g. ":" or "root:"?

ldv-alt:
Back in 2003 I implemented a timing attack protection in pam_tcb, see http://git.altlinux.org/gears/t/...git?p=tcb.git;a=commitdiff;h=0.9.8.4-owl1
Since pam_tcb is the pam_unix replacement we use in ALT, that approach may still be relevant for pam_unix, too.

ldv-alt:
On Tue, Jun 16, 2020 at 07:49:22AM -0700, Dmitry V. Levin wrote:
> Yes, pam_localuser also needs some treatment.  I'll prepare a patch for pam_modutil_check_user_in_passwd.

Commit efd2a79c11982d0feebebbf740506c9555120b97.

ikerexxe:
> Back in 2003 I implemented a timing attack protection in pam_tcb, see http://git.altlinux.org/gears/t/...git?p=tcb.git;a=commitdiff;h=0.9.8.4-owl1
> Since pam_tcb is the pam_unix replacement we use in ALT, that approach may still be relevant for pam_unix, too.

I had already tried something similar to this approach but it was very difficult to get the same executions times. Just creating a fake salt and comparing it took more time than the one needed for the execution of existing users flow.

solardiz:
@ikerexxe wrote:

> creating a fake salt and comparing it took more time than the one needed for the execution of existing users flow.

This is why @ldv-alt's pam_tcb changes that he referenced above create a fake salt unconditionally, even for existing users. Of course, there are still performance differences (perhaps most notably if a multi-line file like `/etc/shadow` is read until first match by the underlying NSS module), but at least this one component (time to create a fake salt) doesn't have to be part of them.

PR #245, opened 2020-06-25
Fixed coverity issues

ikerexxe:
The issues were solved downstream some time ago and have been widely
tested in different versions of Fedora

### DIFF
--- a/libpam_misc/misc_conv.c
+++ b/libpam_misc/misc_conv.c
@@ -244,7 +244,7 @@ static int read_string(int echo, const char *prompt, char **retstr)
     D(("the timer appears to have expired"));
 
     *retstr = NULL;
-    _pam_overwrite(line);
+    _pam_overwrite_n(line, sizeof(line));
 
  cleanexit:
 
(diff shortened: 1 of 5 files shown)


ldv-alt:
On Thu, Jun 25, 2020 at 02:59:47AM -0700, Iker Pedrosa wrote:
> The issues were solved downstream some time ago and have been widely
> tested in different versions of Fedora

What are the issues being fixed?
Please provide some details in the commit message.

ikerexxe:
I'll work on the things that you mentioned and come back with them fixed.

ikerexxe:
I've updated the PR with several commits, each including only a class of issues pointed by coverity. It also includes the description included in the tool.

There's an exception, which is the last commit. I've been unable to find the coverity report that triggered the change. As indicated in the first comment of this PR, it has been in our downstream repository for some time. Anyways, after having a look at it I'm pretty sure that the variables involved in this commit are read-only pointers to read-only chars. That's why I'd like to include the change. In any case, if you find this last commit not suitable I'm also willing to drop it.

t8m:
@ikerexxe the last commit is bogus, please just drop it. I do not know where it came from, if it just some mistake of mine or it came from somewhere else.

ikerexxe:
> @ikerexxe the last commit is bogus, please just drop it. I do not know where it came from, if it was just some mistake of mine or it came from somewhere else.

Thanks! I just dropped it.

ldv-alt:
The first commit in the series that replaces _pam_overwrite(line) with _pam_overwrite_n(line, sizeof(line)) is problematic: if line is not null-terminated, then the preceding strdup(line) is also not valid.  I'll have a look.

ikerexxe:
> Also, could you change commit messages to the present tense, please?

Done

ldv-alt:
Thanks, I've applied all these commits.
I decided to explain these changes further, so I've reworded commit messages.

PR #247, opened 2020-06-25
pam_pwhistory: added helper to handle SELinux

ikerexxe:
The purpose of the helper is to enable tighter confinement of login and
password changing services. The helper is thus called only when SELinux
is enabled on the system.

The issues were solved downstream some time ago and have been widely
tested in different versions of Fedora.

### DIFF
--- a/modules/pam_pwhistory/opasswd.c
+++ b/modules/pam_pwhistory/opasswd.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2008 Thorsten Kukuk <kukuk@suse.de>
+ * Copyright (c) 2013 Red Hat, Inc.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -38,6 +39,7 @@
 #endif
 
 #include <pwd.h>
+#include <shadow.h>
 #include <time.h>
 #include <ctype.h>
 #include <errno.h>
@@ -47,6 +49,9 @@
 #include <string.h>
 #include <stdlib.h>
 #include <syslog.h>
+#ifdef HELPER_COMPILE
+#include <stdarg.h>
+#endif
 #include <sys/stat.h>
 
 #if defined HAVE_LIBXCRYPT
@@ -55,7 +60,14 @@
 #include <crypt.h>
 #endif
 
+#ifdef HELPER_COMPILE
+#define pam_modutil_getpwnam(h,n) getpwnam(n)
+#define pam_modutil_getspnam(h,n) getspnam(n)
+#define pam_syslog(h,a,...) helper_log_err(a,__VA_ARGS__)
+#else
+#include <security/pam_modutil.h>
 #include <security/pam_ext.h>
+#endif
 #include <security/pam_modules.h>
 
 #include "opasswd.h"
@@ -76,6 +88,19 @@ typedef struct {
   char *old_passwords;
 } opwd;
 
+#ifdef HELPER_COMPILE
+void
+helper_log_err(int err, const char *format, ...)
+{
+  va_list args;
+
+  va_start(args, format);
+  openlog(HELPER_COMPILE, LOG_CONS | LOG_PID, LOG_AUTHPRIV);
+  vsyslog(err, format, args);
+  va_end(args);
+  closelog();
+}
+#endif
 
 static int
 parse_entry (char *line, opwd *data)
@@ -117,9 +142,8 @@ compare_password(const char *newpass, const char *oldpass)
 }
 
 /* Check, if the new password is already in the opasswd file.  */
-int
-check_old_pass (pam_handle_t *pamh, const char *user,
-		const char *newpass, int debug)
+PAMH_ARG_DECL(int
+check_old_pass, const char *user, const char *newpass, int debug)
 {
   int retval = PAM_SUCCESS;
   FILE *oldpf;
@@ -127,6 +151,11 @@ check_old_pass (pam_handle_t *pamh, const char *user,
   size_t buflen = 0;
   opwd entry;
   int found = 0;
+
+#ifndef HELPER_COMPILE
+  if (SELINUX_ENABLED)
+    return PAM_PWHISTORY_RUN_HELPER;
+#endif
 
   if ((oldpf = fopen (OLD_PASSWORDS_FILE, "r")) == NULL)
     {
@@ -213,9 +242,8 @@ check_old_pass (pam_handle_t *pamh, const char *user,
   return retval;
 }
 
-int
-save_old_pass (pam_handle_t *pamh, const char *user, uid_t uid,
-	       const char *oldpass, int howmany, int debug UNUSED)
+PAMH_ARG_DECL(int
+save_old_pass, const char *user, int howmany, int debug UNUSED)
 {
   char opasswd_tmp[] = TMP_PASSWORDS_FILE;
   struct stat opasswd_stat;
@@ -226,9 +254,34 @@ save_old_pass (pam_handle_t *pamh, const char *user, uid_t uid,
   char *buf = NULL;
   size_t buflen = 0;
   int found = 0;
+  struct passwd *pwd;
+  const char *oldpass;
+
+  pwd = pam_modutil_getpwnam (pamh, user);
+  if (pwd == NULL)
+    return PAM_USER_UNKNOWN;
 
   if (howmany <= 0)
     return PAM_SUCCESS;
+
+#ifndef HELPER_COMPILE
+  if (SELINUX_ENABLED)
+    return PAM_PWHISTORY_RUN_HELPER;
+#endif
+
+  if ((strcmp(pwd->pw_passwd, "x") == 0)  ||
+      ((pwd->pw_passwd[0] == '#') &&
+       (pwd->pw_passwd[1] == '#') &&
+       (strcmp(pwd->pw_name, pwd->pw_passwd + 2) == 0)))
+    {
+      struct spwd *spw = pam_modutil_getspnam (pamh, user);
+
+      if (spw == NULL)
+        return PAM_USER_UNKNOWN;
+      oldpass = spw->sp_pwdp;
+    }
+  else
+      oldpass = pwd->pw_passwd;
 
   if (oldpass == NULL || *oldpass == '\0')
     return PAM_SUCCESS;
@@ -452,7 +505,7 @@ save_old_pass (pam_handle_t *pamh, const char *user, uid_t uid,
     {
       char *out;
 
-      if (asprintf (&out, "%s:%d:1:%s\n", user, uid, oldpass) < 0)
+      if (asprintf (&out, "%s:%d:1:%s\n", user, pwd->pw_uid, oldpass) < 0)
 	{
 	  retval = PAM_AUTHTOK_ERR;
 	  if (oldpf)
--- a/modules/pam_pwhistory/opasswd.h
+++ b/modules/pam_pwhistory/opasswd.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2008 Thorsten Kukuk <kukuk@suse.de>
+ * Copyright (c) 2013 Red Hat, Inc.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -36,10 +37,30 @@
 #ifndef __OPASSWD_H__
 #define __OPASSWD_H__
 
-extern int check_old_pass (pam_handle_t *pamh, const char *user,
-			   const char *newpass, int debug);
-extern int save_old_pass (pam_handle_t *pamh, const char *user,
-			  uid_t uid, const char *oldpass,
-			  int howmany, int debug);
+#define PAM_PWHISTORY_RUN_HELPER PAM_CRED_INSUFFICIENT
+
+#ifdef WITH_SELINUX
+#include <selinux/selinux.h>
+#define SELINUX_ENABLED (is_selinux_enabled()>0)
+#else
+#define SELINUX_ENABLED 0
+#endif
+
+#ifdef HELPER_COMPILE
+#define PAMH_ARG_DECL(fname, ...) fname(__VA_ARGS__)
+#else
+#define PAMH_ARG_DECL(fname, ...) fname(pam_handle_t *pamh, __VA_ARGS__)
+#endif
+
+#ifdef HELPER_COMPILE
+void
+helper_log_err(int err, const char *format, ...);
+#endif
+
+PAMH_ARG_DECL(int
+check_old_pass, const char *user, const char *newpass, int debug);
+
+PAMH_ARG_DECL(int
+save_old_pass, const char *user, int howmany, int debug);
 
 #endif /* __OPASSWD_H__ */
--- a/modules/pam_pwhistory/pam_pwhistory.c
+++ b/modules/pam_pwhistory/pam_pwhistory.c
@@ -3,6 +3,7 @@
  *
  * Copyright (c) 2008, 2012 Thorsten Kukuk
  * Author: Thorsten Kukuk <kukuk@thkukuk.de>
+ * Copyright (c) 2013 Red Hat, Inc.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -46,10 +47,14 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
-#include <shadow.h>
 #include <syslog.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/time.h>
+#include <sys/resource.h>
+#include <sys/wait.h>
+#include <signal.h>
+#include <fcntl.h>
 
 #include <security/pam_modules.h>
 #include <security/pam_modutil.h>
@@ -58,8 +63,6 @@
 
 #include "opasswd.h"
 #include "pam_inline.h"
-
-#define DEFAULT_BUFLEN 2048
 
 struct options_t {
   int debug;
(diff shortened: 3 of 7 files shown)


modules/pam_pwhistory/pwhistory_helper.c
```diff
@@ -0,0 +1,208 @@
...
+#include <string.h>
+#include <syslog.h>
+#include <errno.h>
+#include <unistd.h>
+#include <signal.h>
+#include <security/_pam_types.h>
+#include <security/_pam_macros.h>
+#include "opasswd.h"
+
+#define MAXPASS 200
```
ldv-alt, line 50: Why 200?  How does this limitation correspond to the limit imposed by the module (if any)?

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -105,14 +111,203 @@ parse_option (pam_handle_t *pamh, const char *argv, options_t *options)
...
+      _exit(PAM_SYSTEM_ERR);
+    }
+  else if (child > 0)
+    {
+      /* wait for child */
+      int rc = 0;
+      rc = waitpid(child, &retval, 0);  /* wait for helper to complete */
+      if (rc < 0)
+        {
+          pam_syslog(pamh, LOG_ERR, "pwhistory_helper save waitpid returned %d: %m", rc);
```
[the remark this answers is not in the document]
t8m: "save" is the command line argument so we can differentiate from inadvertant invocation of the helper without any arguments. Yeah, logging the -1 does not make much sense.

modules/pam_pwhistory/pwhistory_helper.c
```diff
@@ -0,0 +1,208 @@
...
+  action.sa_handler = SIG_IGN;
+  action.sa_flags = 0;
+  (void) sigaction(SIGTERM, &action, NULL);
+  (void) sigaction(SIGHUP, &action, NULL);
+  (void) sigaction(SIGINT, &action, NULL);
+  (void) sigaction(SIGQUIT, &action, NULL);
+}
+
+static int
+read_passwords(int fd, int npass, char **passwords)
```
ldv-alt, line 95: Do you really need this function?  Isn't pam_modutil_read enough?

modules/pam_pwhistory/pam_pwhistory.c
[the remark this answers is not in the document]
ldv-alt: I suggest the following wording then:
"pwhistory_helper save: waitpid: %m"

modules/pam_pwhistory/pwhistory_helper.c
t8m: I think it was wrong to replace read_passwords with pam_modutil_read. The return value is different. If that is fixed, it will probably somewhat work but the read_passwords handles reads from pipe better because it does not try to read until eof but just until it encounters the `\0` character. We have the same function in the pam_unix helpers.

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -105,14 +111,180 @@ parse_option (pam_handle_t *pamh, const char *argv, options_t *options)
...
+
+      if (pam_modutil_sanitize_helper_fds(pamh, PAM_MODUTIL_PIPE_FD,
+          PAM_MODUTIL_PIPE_FD,
+          PAM_MODUTIL_PIPE_FD) < 0)
+        {
+          _exit(PAM_SYSTEM_ERR);
+        }
+
+      /* exec binary helper */
+      args[0] = PWHISTORY_HELPER;
```
t8m, line 139: You need to cast out the const.

modules/pam_pwhistory/pwhistory_helper.c
```diff
@@ -0,0 +1,165 @@
...
+static int
+check_history(const char *user, const char *debug)
+{
+  char pass[MAXPASS + 1];
+  int npass;
+  int dbg = atoi(debug); /* no need to be too fancy here */
+  int retval;
+
+  /* read the password from stdin (a pipe from the pam_pwhistory module) */
+  npass = pam_modutil_read(STDIN_FILENO, pass, MAXPASS + 1);
```
t8m, line 105: This is wrong, please revert back to the read_passwod.

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -105,14 +111,180 @@ parse_option (pam_handle_t *pamh, const char *argv, options_t *options)
...
+
+      if (pam_modutil_sanitize_helper_fds(pamh, PAM_MODUTIL_IGNORE_FD,
+          PAM_MODUTIL_PIPE_FD,
+          PAM_MODUTIL_PIPE_FD) < 0)
+        {
+          _exit(PAM_SYSTEM_ERR);
+        }
+
+      /* exec binary helper */
+      args[0] = PWHISTORY_HELPER;
```
t8m, line 224: and here

modules/pam_pwhistory/pwhistory_helper.c
ldv-alt: pam_modutil_read has a different semantics of the return value, and it doesn't null-terminate.
However, I still think read_passwords is overkill here: we don't need multiple passwords, and terminating the loop on reading '\0' is not something we could benefit from.

t8m: Why we would not benefit?

ldv-alt: Do you want to impose a requirement that the data passed through this pipe must be null-terminated?

t8m: The read_passwords handled that properly. And yes, that requirement was there and is there for the other helpers as well. Anything that is passed and is not null terminated is ignored.

ldv-alt: BTW, why this implementation of read_passwords differs from the one defined in modules/pam_unix/passverify.c?
For example,
```
-        /* The passwords array must contain npass preallocated
-         * buffers of length MAXPASS + 1
-         */
         int rbytes = 0;
         int offset = 0;
         int i = 0;
         char *pptr;
-        while (npass > 0) {
-                rbytes = read(fd, passwords[i]+offset, MAXPASS+1-offset);
+  while (npass > 0)
+    {
+      rbytes = read(fd, passwords[i]+offset, MAXPASS-offset);
```
Looks like commit e89d4c973 is missing in this version.

Can we have a single implementation instead of two different copies, please?

t8m: Can we use inline functions? I'd like to avoid dependency on libpam in helpers.
Or perhaps have some helper_common.c in modules/ directory.

ldv-alt: Yes, we can use inline functions.  In fact, we already use them, see e.g. pam_str_skip_prefix_len.
We could introduce a convenience library for use in helpers, but in this case an inline function should be good enough.

ikerexxe: I've moved read_passwords from pam_unix to pam_inline, which includes [e89d4c9](https://github.com/linux-pam/linux-pam/commit/e89d4c97385ff8180e6e81e84c5aa745daf28a79), and renamed the function to pam_read_passwords.

ldv-alt: Thanks.  Now I see that MAXPASS macro is defined both in modules/pam_unix/passverify.h and libpam/include/pam_inline.h files.  Could we just drop it and use PAM_MAX_RESP_SIZE instead?

ldv-alt: Something like commit b36351dd0137034e79194023c8d687b495e785c4 in ldv/next branch.

modules/pam_pwhistory/pwhistory_helper.c
```diff
@@ -0,0 +1,167 @@
...
+int
+main(int argc, char *argv[])
+{
+  const char *option;
+  const char *user;
+
+  /*
+   * Catch or ignore as many signal as possible.
+   */
+  setup_signals();
```
ldv-alt, line 143: I'm aware of pam_unix/unix_chkpwd doing this thing from the very first commit ea488580c42e8918445a945484de3c8a5addc761, but I wonder why this is needed.

modules/pam_pwhistory/pwhistory_helper.c
```diff
@@ -0,0 +1,167 @@
...
+   * this is to discourage casual use.
+   */
+
+  if (isatty(STDIN_FILENO) || argc < 4)
+    {
+      fprintf(stderr,
+		"This binary is not designed for running in this way.\n");
+      sleep(10);	/* this should discourage/annoy the user */
+      return PAM_SYSTEM_ERR;
+    }
```
ldv-alt, line 156: Also, unix_chkpwd is a privileged helper while pwhistory_helper is not, so I suppose it can be packaged with appropriate permissions, e.g. 0700, instead of discouraging and annoying the user.

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -60,6 +65,7 @@
 #include "pam_inline.h"
 
 #define DEFAULT_BUFLEN 2048
+#define MAX_FD_NO 20000
```
ldv-alt, line 68: MAX_FD_NO is unused.

modules/pam_pwhistory/pwhistory_helper.c
t8m: Yeah, I have no issue with dropping the sleep().

modules/pam_pwhistory/pwhistory_helper.c
t8m: So it isn't inadvertently terminated by signals in the middle of operation? I am not saying it provides any extra security but is it really bad to have it there?

modules/pam_pwhistory/pwhistory_helper.c
ikerexxe: I've rebased my changes on top of that commit, so you'll see it in this PR.

modules/pam_pwhistory/pam_pwhistory.c
ikerexxe: DEFAULT_BUFLEN was also unused so I've deleted both of them.

modules/pam_pwhistory/pwhistory_helper.c
ldv-alt: No, it doesn't look as a reliable method of protection against terminating signals.
Firstly, SIGILL, SIGTRAP, SIGBUS, and SIGSERV are silently converted into _exit(sig).
Secondly, only four signals (SIGTERM, SIGHUP, SIGINT, and SIGQUIT) are ignored, remaining terminating signals are not changed.
Thirdly, none of this is applied in case when the helper is not used.
The last but not least, I've had a look at save_old_pass - the only operation that modifies something, and it seems to be safe: OLD_PASSWORDS_FILE is modified atomically by rename syscall.

So no, I still don't see why setup_signals is needed in this helper.

t8m: I am fine with dropping it.

modules/pam_pwhistory/pwhistory_helper.c
ldv-alt: Would it be OK for this PR if I push these pam_read_passwords related commits from ldv/next to master?
This PR would have to be rebased to master anyway.

ikerexxe: Yes, no problem. Just tell me when they are ready.

ldv-alt: I think they are ready.

modules/pam_pwhistory/opasswd.c
```diff
@@ -117,8 +142,8 @@ compare_password(const char *newpass, const char *oldpass)
 }
 
 /* Check, if the new password is already in the opasswd file.  */
-int
-check_old_pass (pam_handle_t *pamh, const char *user,
+PAMH_ARG_DECL(int
+check_old_pass, const char *user,
 		const char *newpass, int debug)
```
ldv-alt, line 147: Please merge these two lines into a single one.

modules/pam_pwhistory/opasswd.c
```diff
@@ -213,9 +243,9 @@ check_old_pass (pam_handle_t *pamh, const char *user,
   return retval;
 }
 
-int
-save_old_pass (pam_handle_t *pamh, const char *user, uid_t uid,
-	       const char *oldpass, int howmany, int debug UNUSED)
+PAMH_ARG_DECL(int
+save_old_pass, const char *user,
+	       int howmany, int debug UNUSED)
```
ldv-alt, line 248: Please merge these two lines into a single one.

modules/pam_pwhistory/opasswd.h
```diff
@@ -36,10 +37,32 @@
...
+#define SELINUX_ENABLED 0
+#endif
+
+#ifdef HELPER_COMPILE
+#define PAMH_ARG_DECL(fname, ...) fname(__VA_ARGS__)
+#define PAMH_ARG(...)               __VA_ARGS__
+#else
+#define PAMH_ARG_DECL(fname, ...) fname(pam_handle_t *pamh, __VA_ARGS__)
+#define PAMH_ARG(...)               pamh, __VA_ARGS__
+#endif
```
ldv-alt, line 55: PAMH_ARG seems to be unused.

modules/pam_pwhistory/opasswd.h
```diff
@@ -36,10 +37,32 @@
...
-extern int check_old_pass (pam_handle_t *pamh, const char *user,
-			   const char *newpass, int debug);
-extern int save_old_pass (pam_handle_t *pamh, const char *user,
-			  uid_t uid, const char *oldpass,
-			  int howmany, int debug);
+#define PAM_PWHISTORY_RUN_HELPER PAM_CRED_INSUFFICIENT
+
+#ifdef WITH_SELINUX
+#include <selinux/selinux.h>
+#define SELINUX_ENABLED is_selinux_enabled()>0
```
ldv-alt, line 44:
```
#define SELINUX_ENABLED (is_selinux_enabled() > 0)
```

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -105,14 +108,184 @@ parse_option (pam_handle_t *pamh, const char *argv, options_t *options)
...
+      if (asprintf(&args[3], "%d", howmany) < 0)
+        {
+          pam_syslog(pamh, LOG_ERR, "asprintf on args[3] failed");
+          _exit(PAM_SYSTEM_ERR);
+        }
+      if (asprintf(&args[4], "%d", debug) < 0)
+        {
+          pam_syslog(pamh, LOG_ERR, "asprintf on args[4] failed");
+          _exit(PAM_SYSTEM_ERR);
+        }
```
ldv-alt, line 150: Error diagnostics here is too verbose.  Something more concise like
```
if (asprintf(&args[3], "%d", howmany) < 0 ||
    asprintf(&args[4], "%d", debug) < 0)
  {
    pam_syslog(pamh, LOG_ERR, "asprintf: %m");
    _exit(PAM_SYSTEM_ERR);
  }
```
should be enough.

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -105,14 +108,184 @@ parse_option (pam_handle_t *pamh, const char *argv, options_t *options)
...
+
+      execve(args[0], args, envp);
```
ldv-alt, line 152: Please add some error diagnostics here, e.g.
```
pam_syslog(pamh, LOG_ERR, "execve: %s: %m", args[0]);
```

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -105,14 +108,184 @@ parse_option (pam_handle_t *pamh, const char *argv, options_t *options)
...
+          retval = PAM_SYSTEM_ERR;
+        }
+      else
+        {
+          retval = WEXITSTATUS(retval);
+        }
+    }
+  else
+    {
+      retval = PAM_SYSTEM_ERR;
```
ldv-alt, line 179: Please add some error diagnostics here, e.g.
```
pam_syslog(pamh, LOG_ERR, "fork: %m");
```

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -105,14 +108,184 @@ parse_option (pam_handle_t *pamh, const char *argv, options_t *options)
...
+      args[1] = (char *)"check";
+      args[2] = (char *)user;
+      DIAG_POP_IGNORE_CAST_QUAL;
+      if (asprintf(&args[3], "%d", debug) < 0)
+        {
+          pam_syslog(pamh, LOG_ERR, "asprintf on args[3] failed");
+          _exit(PAM_SYSTEM_ERR);
+        }
+
+      execve(args[0], args, envp);
```
ldv-alt, line 234: Likewise, please add some error diagnostics here.

modules/pam_pwhistory/pam_pwhistory.c
```diff
@@ -105,14 +108,184 @@ parse_option (pam_handle_t *pamh, const char *argv, options_t *options)
...
+      else
+        {
+          retval = WEXITSTATUS(retval);
+        }
+    }
+  else
+    {
+      close(fds[0]);
+      close(fds[1]);
+      retval = PAM_SYSTEM_ERR;
```
ldv-alt, line 275: Likewise, please add some error diagnostics here.

modules/pam_pwhistory/pwhistory_helper.c
```diff
@@ -0,0 +1,117 @@
...
+
+  option = argv[1];
+  user = argv[2];
+
+  if (strcmp(option, "check") == 0 && argc == 4)
+    return check_history(user, argv[3]);
+  else if (strcmp(option, "save") == 0 && argc == 5)
+    return save_history(user, argv[3], argv[4]);
+
+  return PAM_SYSTEM_ERR;
```
ldv-alt, line 116: It might be helpful to repeat the same usage diagnostics here.

ikerexxe: Do you mean something like:
```
pam_syslog(pamh, LOG_ERR, "pwhistory_helper executed without 'check' or 'save' option or without the correct number of arguments");
```

Or maybe something more generic:
```
pam_syslog(pamh, LOG_ERR, "pwhistory_helper incorrect arguments");
```

ldv-alt: That's up to you, I'd just re-used the same diagnostics that's printed in case of argc < 4.

modules/pam_pwhistory/opasswd.c
```diff
@@ -117,9 +142,8 @@ compare_password(const char *newpass, const char *oldpass)
 }
 
 /* Check, if the new password is already in the opasswd file.  */
-int
-check_old_pass (pam_handle_t *pamh, const char *user,
-		const char *newpass, int debug)
+PAMH_ARG_DECL(int
+check_old_pass, const char *user,	const char *newpass, int debug)
```
ldv-alt, line 146: Odd spacing after comma.

t8m:
I am OK with the patch as is.

PR #258, opened 2020-07-27
SELinux: replace deprecated types and functions

cgzones:


ldv-alt:
These commits also replace many selinux functions with their _raw counterparts,
could you add an explanation for this change into appropriate commit messages, please?

Or maybe also move this part of changes into separate commit(s).

Currently they say just "Also skip not needed context translation".

PR #259, opened 2020-08-01
Fix typo on --with-kernel-overflow-uid

concatime:
Or change documentation to `--with-kerneloverflowuid`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -671,7 +671,7 @@ test -n "$opt_sysuidmin" ||
           opt_sysuidmin=101
 AC_DEFINE_UNQUOTED(PAM_USERTYPE_SYSUIDMIN, $opt_sysuidmin, [Minimum system user uid.])
 
-AC_ARG_WITH([kerneloverflowuid], AS_HELP_STRING([--with-kernel-overflow-uid=<number>],[kernel overflow uid, default (uint16_t)-2=65534]), opt_kerneloverflowuid=$withval)
+AC_ARG_WITH([kernel-overflow-uid], AS_HELP_STRING([--with-kernel-overflow-uid=<number>],[kernel overflow uid, default (uint16_t)-2=65534]), opt_kerneloverflowuid=$withval)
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -671,7 +671,7 @@ test -n "$opt_sysuidmin" ||
           opt_sysuidmin=101
 AC_DEFINE_UNQUOTED(PAM_USERTYPE_SYSUIDMIN, $opt_sysuidmin, [Minimum system user uid.])
 
-AC_ARG_WITH([kerneloverflowuid], AS_HELP_STRING([--with-kernel-overflow-uid=<number>],[kernel overflow uid, default (uint16_t)-2=65534]), opt_kerneloverflowuid=$withval)
+AC_ARG_WITH([kernel-overflow-uid], AS_HELP_STRING([--with-kernel-overflow-uid=<number>],[kernel overflow uid, default (uint16_t)-2=65534]), opt_kerneloverflowuid=$withval)
```
ldv-alt, line 674: This was introduced by commit 926d7935edf35385e6c28bb97666aee443b71e46.
If anything, it doesn't look like a typo, although the variant you suggested looks nicer.

From the GNU Autoconf perspective, the first argument of AC_ARG_WITH should consist only of alphanumeric characters, dashes, plus signs, and dots.

@pbrezina @t8m what do you think about this proposal?

concatime: Well, the documentation says `--with-kernel-overflow-uid=` but that does not work. `--with-kerneloverflowuid=` does. So, either change the documentation to `--with-kerneloverflowuid=` or change the internal code to reflect `--with-kernel-overflow-uid=` like my commit.

ldv-alt: Thanks, now it's more clear what was the intent of the author.
Could you change the wording of your commit message to reflect that? e.g.
```
configure.ac: fix typo in --with-kernel-overflow-uid= option to match its documentation
```

concatime: Done.

PR #262, opened 2020-08-26
pam_wheel: if getlogin fails fallback to PAM_RUSER

ikerexxe:
modules/pam_wheel/pam_wheel.c: if getlogin fails to obtain the real user
ID, then try with PAM_RUSER. Besides, improve indentation.
modules/pam_wheel/pam_wheel.8.xml: indicate that use_uid option uses the
real uid of the calling process.

Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=1866866

### DIFF
--- a/modules/pam_wheel/pam_wheel.c
+++ b/modules/pam_wheel/pam_wheel.c
@@ -141,6 +141,16 @@ perform_check (pam_handle_t *pamh, int ctrl, const char *use_group)
     } else {
         fromsu = pam_modutil_getlogin(pamh);
 
+        /* if getlogin fails try a fallback to PAM_RUSER */
+        if (fromsu == NULL) {
+            const char *rhostname;
+
+            retval = pam_get_item(pamh, PAM_RHOST, (const void **)&rhostname);
+            if (retval != PAM_SUCCESS || rhostname == NULL) {
+                retval = pam_get_item(pamh, PAM_RUSER, (const void **)&fromsu);
+            }
+        }
+
         if (fromsu != NULL) {
             tpwd = pam_modutil_getpwnam (pamh, fromsu);
         }
--- a/modules/pam_wheel/pam_wheel.c
+++ b/modules/pam_wheel/pam_wheel.c
@@ -130,25 +130,27 @@ perform_check (pam_handle_t *pamh, int ctrl, const char *use_group)
     }
 
     if (ctrl & PAM_USE_UID_ARG) {
-	tpwd = pam_modutil_getpwuid (pamh, getuid());
-	if (!tpwd) {
-	    if (ctrl & PAM_DEBUG_ARG) {
+        tpwd = pam_modutil_getpwuid (pamh, getuid());
+        if (tpwd == NULL) {
+            if (ctrl & PAM_DEBUG_ARG) {
                 pam_syslog(pamh, LOG_NOTICE, "who is running me ?!");
-	    }
-	    return PAM_SERVICE_ERR;
-	}
-	fromsu = tpwd->pw_name;
+            }
+            return PAM_SERVICE_ERR;
+        }
+        fromsu = tpwd->pw_name;
     } else {
-	fromsu = pam_modutil_getlogin(pamh);
-	if (fromsu) {
-	    tpwd = pam_modutil_getpwnam (pamh, fromsu);
-	}
-	if (!fromsu || !tpwd) {
-	    if (ctrl & PAM_DEBUG_ARG) {
-		pam_syslog(pamh, LOG_NOTICE, "who is running me ?!");
-	    }
-	    return PAM_SERVICE_ERR;
-	}
+        fromsu = pam_modutil_getlogin(pamh);
+
+        if (fromsu != NULL) {
+            tpwd = pam_modutil_getpwnam (pamh, fromsu);
+        }
+
+        if (fromsu == NULL || tpwd == NULL) {
+            if (ctrl & PAM_DEBUG_ARG) {
+                pam_syslog(pamh, LOG_NOTICE, "who is running me ?!");
+            }
+            return PAM_SERVICE_ERR;
+        }
     }
 
     /*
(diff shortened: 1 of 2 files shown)


modules/pam_wheel/pam_wheel.c
```diff
@@ -130,25 +130,32 @@ perform_check (pam_handle_t *pamh, int ctrl, const char *use_group)
...
-	if (!fromsu || !tpwd) {
-	    if (ctrl & PAM_DEBUG_ARG) {
-		pam_syslog(pamh, LOG_NOTICE, "who is running me ?!");
-	    }
-	    return PAM_SERVICE_ERR;
-	}
+        fromsu = pam_modutil_getlogin(pamh);
+
+        /* if getlogin fails try a fallback to PAM_RUSER */
+        if(!fromsu) {
```
t8m, line 145: Please space after if.

t8m (commented):
Looks good but if you're already changing the indentation on the lines with if(), please also change the condition for NULL/non-NULL pointer values to explicit `== NULL` or `!= NULL`.

ikerexxe:
> Looks good but if you're already changing the indentation on the lines with if(), please also change the condition for NULL/non-NULL pointer values to explicit `== NULL` or `!= NULL`.

Updated with your comments.

ldv-alt:
On Wed, Aug 26, 2020 at 06:40:18AM -0700, Iker Pedrosa wrote:
> modules/pam_wheel/pam_wheel.c: if getlogin fails to obtain the real user
> ID, then try with PAM_RUSER. Besides, improve indentation.
> 
> Resolves:
> https://bugzilla.redhat.com/show_bug.cgi?id=1866866

What's the rationale for this change?  The bugzilla bug is not accessible.

Can PAM_RUSER be trusted at all?
Also, is PAM_RUSER valid for the purposes of this module if PAM_RHOST is also set?

ikerexxe:
> On Wed, Aug 26, 2020 at 06:40:18AM -0700, Iker Pedrosa wrote: modules/pam_wheel/pam_wheel.c: if getlogin fails to obtain the real user ID, then try with PAM_RUSER. Besides, improve indentation. Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1866866
> What's the rationale for this change? The bugzilla bug is not accessible. Can PAM_RUSER be trusted at all? Also, is PAM_RUSER valid for the purposes of this module if PAM_RHOST is also set?

Sorry, I didn't notice that it was set as private. Now, it's been changed and you should be able to see it.

t8m:
> What's the rationale for this change? The bugzilla bug is not accessible.

I've opened up the bug. The problem is that currently the terminal windows no longer provide a login session of their own.

> Can PAM_RUSER be trusted at all? Also, is PAM_RUSER valid for the purposes of this module if PAM_RHOST is also set?

Not sure if anything sets PAM_RUSER along with PAM_RHOST - at least I would not expect it to be set. But yeah, perhaps as a safety measure we should use PAM_RUSER only if PAM_RHOST is not set. As for trusting the PAM_RUSER - I do not see a reason why we should not trust it. If a PAM application is setting PAM_RUSER randomly, it is a bug (and security issue) in the application, not in PAM.

ldv-alt:
On Thu, Aug 27, 2020 at 02:51:22AM -0700, Tomáš Mráz wrote:
> > What's the rationale for this change? The bugzilla bug is not accessible.
> 
> I've opened up the bug. The problem is that currently the terminal windows no longer provide a login session of their own.

Have these terminal emulators stopped using libutempter?
I wonder what was the rationale for this change.

t8m:
Unfortunately I have no idea.

ldv-alt:
On Thu, Aug 27, 2020 at 03:42:21AM -0700, Tomáš Mráz wrote:
> Unfortunately I have no idea.

Should the issue be raised with those programs which introduced the problem in the first place?
After all, pam_wheel is not the only place in the system where getlogin could be used.
For me the suggestion to honor PAM_RUSER in pam_wheel looks like a way to mitigate a bug introduced in other software.

t8m:
The thing is nowhere in the pam_wheel documentation we say that the getlogin is the source of the information on who is the requesting user. Also the use_uid option documentation seems to be fairly misleading. So yes, alternatively (or in addition) we should fix the documentation.

t8m:
It might be that the programs that introduced the change should be reverted to original behavior. On the other hand I do not see a reason why this should not be merged. @ldv-alt do you have any strong reason why this should not be merged?

ldv-alt:
On Mon, Sep 07, 2020 at 06:01:08AM -0700, Tomáš Mráz wrote:
> It might be that the programs that introduced the change should be reverted to original behavior. On the other hand I do not see a reason why this should not be merged. @ldv-alt do you have any strong reason why this should not be merged?

I think we should promote use_uid option instead, but no, this is not a sustained objection.  If this fallback is to be merged, please make sure PAM_RUSER is not trusted when PAM_RHOST is also set.

ikerexxe:
I've made two changes taking into account your comments. The first on is that if PAM_RHOST is set, then PAM_RUSER is not used to identify the user. The second one consists in updating the man page to show the actual behaviour of use_uid option.

t8m (approved):
LGTM

modules/pam_wheel/pam_wheel.c
```diff
@@ -130,25 +131,35 @@ perform_check (pam_handle_t *pamh, int ctrl, const char *use_group)
...
-	}
+        fromsu = pam_modutil_getlogin(pamh);
+
+        /* if getlogin fails try a fallback to PAM_RUSER */
+        if (fromsu == NULL) {
+            retval = pam_get_item(pamh, PAM_RHOST, (const void **)&rhostname);
+            if (rhostname == NULL) {
+                retval = pam_get_item(pamh, PAM_RUSER, (const void **)&fromsu);
+            }
+        }
```
ldv-alt, line 151: Please check retval. 
Also, the scope of rhostname variable could be narrower than the whole function.

ldv-alt (changes requested):

ldv-alt (approved):

t8m:
My approval holds as well.

PR #269, opened 2020-09-15
pam_motd: filter motd by user and group

ikerexxe:
modules/pam_motd/pam_motd.c: filter motd by user and group owning the
proper files. This is achieved by changing the ids of the process
reading the files from root to the target user.

Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=1861640

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,8 @@
 Linux-PAM NEWS -- history of user-visible changes.
 
+Release 1.5.0
+* pam_motd: read motd files with target user credentials skipping unreadable ones.
+
 Release 1.4.0
 * Multiple minor bug fixes and documentation improvements
 * Fixed grammar of messages printed via pam_prompt
--- a/modules/pam_motd/pam_motd.8.xml
+++ b/modules/pam_motd/pam_motd.8.xml
@@ -64,8 +64,9 @@
       override files with the same name in <filename>/usr/lib/motd.d/</filename>.
     </para>
     <para>
-      Files the in the directories listed above are displayed in
-      lexicographic order by name.
+      Files in the directories listed above are displayed in lexicographic
+      order by name. Moreover, the files are filtered by reading them with the
+      credentials of the target user authenticating on the system.
     </para>
     <para>
       To silence a message,
--- a/modules/pam_motd/pam_motd.c
+++ b/modules/pam_motd/pam_motd.c
@@ -282,6 +282,72 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
     _pam_drop(dirscans);
 }
 
+static int drop_privileges(pam_handle_t *pamh, struct pam_modutil_privs *privs)
+{
+    struct passwd *pw;
+    const char *username;
+    int retval;
+
+    retval = pam_get_user(pamh, &username, "key user");
+
+    if (retval == PAM_SUCCESS) {
+        pw = pam_modutil_getpwnam (pamh, username);
+    } else {
+        return PAM_SESSION_ERR;
+    }
+
+    if (pw == NULL || pam_modutil_drop_priv(pamh, privs, pw)) {
+        return PAM_SESSION_ERR;
+    }
+
+    return PAM_SUCCESS;
+}
+
+static int try_to_display(pam_handle_t *pamh, char **motd_path_split,
+                          unsigned int num_motd_paths,
+                          char **motd_dir_path_split,
+                          unsigned int num_motd_dir_paths, int report_missing)
+{
+    PAM_MODUTIL_DEF_PRIVS(privs);
+
+    if (drop_privileges(pamh, &privs) != PAM_SUCCESS) {
+        pam_syslog(pamh, LOG_ERR, "Unable to drop privileges");
+        return PAM_SESSION_ERR;
+    }
+
+    if (motd_path_split != NULL) {
+        unsigned int i;
+
+        for (i = 0; i < num_motd_paths; i++) {
+            int fd = open(motd_path_split[i], O_RDONLY, 0);
+
+            if (fd >= 0) {
+                try_to_display_fd(pamh, fd);
+                close(fd);
+
+                /* We found and displayed a file,
+                    * move onto next filename.
+                    */
+                break;
+            }
+        }
+    }
+
+    if (motd_dir_path_split != NULL) {
+        try_to_display_directories_with_overrides(pamh,
+                                                    motd_dir_path_split,
+                                                    num_motd_dir_paths,
+                                                    report_missing);
+    }
+
+    if (pam_modutil_regain_priv(pamh, &privs)) {
+        pam_syslog(pamh, LOG_ERR, "Unable to regain privileges");
+        return PAM_SESSION_ERR;
+    }
+
+    return PAM_SUCCESS;
+}
+
 int pam_sm_open_session(pam_handle_t *pamh, int flags,
 			int argc, const char **argv)
 {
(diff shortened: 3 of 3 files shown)


modules/pam_motd/pam_motd.c
```diff
@@ -282,6 +283,97 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
...
+    if (setuid(pw->pw_uid) < 0) {
+        pam_syslog(pamh, LOG_ERR, "Unable to change UID to %d\n", pw->pw_uid);
+
+        retval = PAM_SESSION_ERR;
+        goto out;
+    }
+
+out:
+    return retval;
+}
```
ldv-alt, line 316: Please consider using pam_modutil_drop_priv/pam_modutil_regain_priv interface.

ldv-alt (changes requested):

modules/pam_motd/pam_motd.c
ikerexxe: Thanks! Now the code is neater.

ikerexxe:
@ldv-alt do you have any feedback now that pam_modutil_*_priv is being used?

ldv-alt:
On Wed, Sep 23, 2020 at 01:28:20AM -0700, Iker Pedrosa wrote:
> @ldv-alt do you have any feedback now that `pam_modutil_*_priv` is being used?

Sorry, I was distracted by other things.

I have three questions/concerns:

- If motd_path_split == NULL, pam_modutil_regain_priv() will be invoked
without a prior pam_modutil_drop_priv() call; I don't think
pam_modutil_regain_priv() would be happy about it.

- If motd_path_split == NULL and motd_dir_path_split != NULL, then
try_to_display_directories_with_overrides() will be invoked without a prior
drop_privileges() call; is this intended?

- In a very unlikely case when pam_modutil_regain_priv returns an error,
the process will be left in an inconsistent state; the module should fail
with an error in that case, shouldn't it?

modules/pam_motd/pam_motd.c
```diff
@@ -358,35 +424,23 @@ int pam_sm_open_session(pam_handle_t *pamh, int flags,
...
     _pam_drop(motd_dir_path_split);
 
-    retval = pam_putenv(pamh, "MOTD_SHOWN=pam");
-
-    return retval == PAM_SUCCESS ? PAM_IGNORE : retval;
+    if (retval == PAM_SUCCESS) {
+        retval = pam_putenv(pamh, "MOTD_SHOWN=pam");
+        return retval == PAM_SUCCESS ? PAM_IGNORE : retval;
+    } else {
+        pam_putenv(pamh, "MOTD_SHOWN=pam");
```
ikerexxe, line 441: I'm not sure if this should be set in case of failure. What's your opinion?

ikerexxe:
> On Wed, Sep 23, 2020 at 01:28:20AM -0700, Iker Pedrosa wrote: @ldv-alt do you have any feedback now that `pam_modutil_*_priv` is being used?
> Sorry, I was distracted by other things. I have three questions/concerns: - If motd_path_split == NULL, pam_modutil_regain_priv() will be invoked without a prior pam_modutil_drop_priv() call; I don't think pam_modutil_regain_priv() would be happy about it. - If motd_path_split == NULL and motd_dir_path_split != NULL, then try_to_display_directories_with_overrides() will be invoked without a prior drop_privileges() call; is this intended? - In a very unlikely case when pam_modutil_regain_priv returns an error, the process will be left in an inconsistent state; the module should fail with an error in that case, shouldn't it?

I agree with all of your points. Thank you for the feedback.

ldv-alt:
On Thu, Sep 24, 2020 at 01:53:12AM -0700, Iker Pedrosa wrote:
> @ikerexxe commented on this pull request.
> >  
>    out:
>      _pam_drop(motd_path_copy);
>      _pam_drop(motd_path_split);
>      _pam_drop(motd_dir_path_copy);
>      _pam_drop(motd_dir_path_split);
>  
> -    retval = pam_putenv(pamh, "MOTD_SHOWN=pam");
> -
> -    return retval == PAM_SUCCESS ? PAM_IGNORE : retval;
> +    if (retval == PAM_SUCCESS) {
> +        retval = pam_putenv(pamh, "MOTD_SHOWN=pam");
> +        return retval == PAM_SUCCESS ? PAM_IGNORE : retval;
> +    } else {
> +        pam_putenv(pamh, "MOTD_SHOWN=pam");
> 
> I'm not sure if this should be set in case of failure. What's your opinion?

I'd rather be on the safe side and wouldn't set MOTD_SHOWN if the module
is going to return an error.

ikerexxe:
> On Thu, Sep 24, 2020 at 01:53:12AM -0700, Iker Pedrosa wrote: @ikerexxe commented on this pull request. > out: _pam_drop(motd_path_copy); _pam_drop(motd_path_split); _pam_drop(motd_dir_path_copy); _pam_drop(motd_dir_path_split); - retval = pam_putenv(pamh, "MOTD_SHOWN=pam"); - - return retval == PAM_SUCCESS ? PAM_IGNORE : retval; + if (retval == PAM_SUCCESS) { + retval = pam_putenv(pamh, "MOTD_SHOWN=pam"); + return retval == PAM_SUCCESS ? PAM_IGNORE : retval; + } else { + pam_putenv(pamh, "MOTD_SHOWN=pam"); I'm not sure if this should be set in case of failure. What's your opinion?
> I'd rather be on the safe side and wouldn't set MOTD_SHOWN if the module is going to return an error.

Done!

modules/pam_motd/pam_motd.c
```diff
@@ -282,6 +282,72 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
...
+    const char *username;
+    int retval;
+
+    retval = pam_get_user(pamh, &username, "key user");
+
+    if (retval == PAM_SUCCESS) {
+        pw = pam_modutil_getpwnam (pamh, username);
+    } else {
+        return PAM_SESSION_ERR;
+    }
```
ldv-alt, line 297: Please check the value returned by pam_modutil_getpwnam(); if it's NULL, then pam_modutil_drop_priv() shouldn't be called.

ldv-alt (changes requested):

modules/pam_motd/pam_motd.c
ikerexxe: Done

ldv-alt (approved):

ldv-alt:
I think it's OK now, thanks.

ikerexxe:
Do you think this is good enough to merge or do you want another pair of eyes to review it?

t8m:
The code is good. However I strongly believe this change should be reflected in pam_motd manual page and in the NEWS file. So please add that documentation as a separate patch and I'll merge this.

ikerexxe:
I've created a new release section in the NEWS file, but I'm not sure if it should be 1.5.0 (as written) or 1.4.1

modules/pam_motd/pam_motd.8.xml
```diff
@@ -64,8 +64,9 @@
       override files with the same name in <filename>/usr/lib/motd.d/</filename>.
     </para>
     <para>
-      Files the in the directories listed above are displayed in
-      lexicographic order by name.
+      Files in the directories listed above are displayed in lexicographic
+      order by name. Moreover, the files are filtered by checking the ownership
```
t8m, line 68: I'd use something like: "files are filtered by reading them with the credentials of the target user authenticating on the system."

NEWS
```diff
@@ -1,5 +1,8 @@
 Linux-PAM NEWS -- history of user-visible changes.
 
+Release 1.5.0
+* pam_motd: filter motd by user and group owning the files.
```
t8m, line 4: I prefer something like: "Read motd files with target user credentials skipping unreadable ones."

t8m:
> I've created a new release section in the NEWS file, but I'm not sure if it should be 1.5.0 (as written) or 1.4.1

That does not matter and can be changed later. But yeah 1.5.0 is more likely.

t8m (approved):
LGTM

t8m:
Merged, thank you for the contribution.

modules/pam_motd/pam_motd.c
```diff
@@ -282,6 +282,72 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
     _pam_drop(dirscans);
 }
 
+static int drop_privileges(pam_handle_t *pamh, struct pam_modutil_privs *privs)
+{
+    struct passwd *pw;
+    const char *username;
+    int retval;
+
+    retval = pam_get_user(pamh, &username, "key user");
```
ldv-alt, line 291: Apparently, I missed this "key user" argument.
I suppose NULL should be passed to pam_get_user instead.

t8m: Ah, good catch. Will you submit a patch?

ikerexxe: Thank you! I think I didn't use the correct example to drop/regain privileges.

ikerexxe: Created PR: https://github.com/linux-pam/linux-pam/pull/282

PR #271, opened 2020-09-29
Prevent SEGFAULT for unknown UID

ya-mouse:
When running systemd service with DynamicUser being set, the dynamic UID
might be not mapped to user name (/etc/nsswitch.conf is not configured
with systemd nss module).

The getuidname() routine might return NULL and this is not checked by callee.

Signed-off-by: Anton D. Kachalov <rnouse@google.com>

### DIFF
--- a/modules/pam_unix/unix_chkpwd.c
+++ b/modules/pam_unix/unix_chkpwd.c
@@ -137,7 +137,7 @@ int main(int argc, char *argv[])
 	  user = getuidname(getuid());
 	  /* if the caller specifies the username, verify that user
 	     matches it */
-	  if (strcmp(user, argv[1])) {
+	  if (user == NULL || strcmp(user, argv[1])) {
 	    user = argv[1];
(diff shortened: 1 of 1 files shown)


modules/pam_unix/unix_chkpwd.c
```diff
@@ -137,7 +137,7 @@ int main(int argc, char *argv[])
 	  user = getuidname(getuid());
 	  /* if the caller specifies the username, verify that user
 	     matches it */
-	  if (strcmp(user, argv[1])) {
+	  if (!user || strcmp(user, argv[1])) {
```
t8m, line 140: Could you please change that to `user == NULL`

t8m (changes requested):

modules/pam_unix/unix_chkpwd.c
ya-mouse: Done

t8m (approved):

t8m:
Thank you for the contribution.

PR #273, opened 2020-10-01
Add format function attributes and enable -Wmissing-format-attribute

cgzones:
This enables compilers to find format specifier mismatches, like:

   foo_print("Hello %d", "world")


conf/pam_conv1/pam_conv_y.y
```diff
@@ -20,6 +20,8 @@
 #include <ctype.h>
 #include <sys/stat.h>
 
+#include <libpam/include/security/_pam_types.h>
```
t8m, line 23: This is wrong path. Use just `security/_pam_types.h`

ldv-alt:
This fails to build with the following diagnostics:
pam_conv_y.y:23:10: fatal error: security/_pam_types.h: No such file or directory

t8m (approved):
LGTM

PR #278, opened 2020-10-12
pam_faillock: fix invalid error message

shanechko:
args_parse function pass "conf=" argument to set_conf_opt() after handling by self.
set_conf_opt is not able to handle "conf" argument and write error:
sddm-helper[415]: pam_faillock(sddm:auth): Unknown option: conf

### DIFF
--- a/modules/pam_faillock/pam_faillock.c
+++ b/modules/pam_faillock/pam_faillock.c
@@ -109,6 +109,7 @@ args_parse(pam_handle_t *pamh, int argc, const char **argv,
 		int flags, struct options *opts)
 {
 	int i;
+	int config_arg_index = -1;
 	int rv;
 	const char *conf = default_faillock_conf;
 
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
I also ran into this and can confirm this PR fixes the problem of spamming my syslog.

My fix previously was
```
                else if (strcmp(argv[i], "authsucc") == 0) {
                        opts->action = FAILLOCK_ACTION_AUTHSUCC;
                }
+               else if (strncmp(argv[i], "conf=", 5) == 0) {
+                       // ignore
+               }
                else {
                        char buf[FAILLOCK_CONF_MAX_LINELEN + 1];
                        char *val;
```
but this PR looks even better.

t8m (approved):
LGTM

t8m:
Merged, thank you for the PR.

PR #279, opened 2020-10-16
pam_wheel: improve coding logic

whzhe51:
Signed-off-by: whzhe <wanghongzhe@huawei.com>


t8m:
I am sorry but I do not see any improvement here. Could you be please more specific in what you're trying to achieve by the change?

whzhe51:
> I am sorry but I do not see any improvement here. Could you be please more specific in what you're trying to achieve by the change?

There are two times if (fromsu == NULL), i think it's redundant.And i just merge into one

t8m:
But you are duplicating other stuff due to that. That does not look like an improvement to me, I am sorry.

t8m:
Also you're really breaking the logic as fromsu can be assigned there.

PR #282, opened 2020-10-21
pam_motd: unset prompt value to drop privileges

ikerexxe:
modules/pam_motd/pam_motd.c: set NULL value instead of "key user" for the
prompt when dropping privileges.


t8m (approved):

t8m:
Merged, thank you.

PR #286, opened 2020-10-31
Prepare for the next release

ldv-alt:
The latest release was made almost 5 months ago, let's prepare for the next one.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,7 +1,15 @@
 Linux-PAM NEWS -- history of user-visible changes.
 
 Release 1.5.0
+* Multiple minor bug fixes, portability fixes, and documentation improvements.
+* Extended libpam API with pam_modutil_check_user_in_passwd function.
+* configure: added --disable-unix option to disable build of pam_unix module.
+* pam_faillock: changed /run/faillock/$USER permissions from 0600 to 0660.
+* pam_limits: added support for nonewprivs item.
 * pam_motd: read motd files with target user credentials skipping unreadable ones.
+* pam_pwhistory: added a SELinux helper executable.
+* pam_unix, pam_usertype: implemented avoidance of certain timing attacks.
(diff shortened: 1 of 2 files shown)


t8m:
Hmm, if we are already removing the modules, the new release should be 1.5.0. Or we can do 1.4.1 still with the modules kept and then immediately remove the modules from master which make it clear that they won't be in the next release.

ldv-alt:
On Mon, Nov 02, 2020 at 02:50:33AM -0800, Tomáš Mráz wrote:
> Hmm, if we are already removing the modules, the new release should be 1.5.0. Or we can do 1.4.1 still with the modules kept and then immediately remove the modules from master which make it clear that they won't be in the next release.

I don't mind keeping these modules a bit longer, but there was a
deprecation warning in the NEWS entry for 1.4.0 saying that these modules
are no longer built by default and will be removed in the next release,
so keeping them in the next release would send a misleading message
to the users.

t8m:
OK, then we have to call the release 1.5.0 as removing a module is not a minor update.

ldv-alt:
We have a reference to 1.4.1:
```
$ git grep -F 1.4.1
libpam/libpam.map:LIBPAM_MODUTIL_1.4.1 {
```
I suggest to leave it unchanged since users of this LIBPAM_MODUTIL_1.4.1 interface already exist.

t8m:
that's IMO no problem to keep that reference.

t8m (approved):

ldv-alt:
OK, master is ready for the next release then.

t8m:
@ldv-alt will you do the release?

ldv-alt:
On Tue, Nov 03, 2020 at 08:09:03AM -0800, Tomáš Mráz wrote:
> @ldv-alt will you do the release?

I could do the release tomorrow.

t8m:
Are we ready for the release now?

ldv-alt:
Done.

PR #287, opened 2020-11-02
3 covscan issues

ikerexxe:
libpam/pam_dispatch: remove store statement since the value is never
read.

modules/pam_faillock/main.c: remove store statement since the value is
only read in the enclosing expression.

modules/pam_ftp/pam_ftp.c: free anon_user before returning as it may be
still in use.

### DIFF
--- a/libpam/pam_dispatch.c
+++ b/libpam/pam_dispatch.c
@@ -424,7 +424,6 @@ int _pam_dispatch(pam_handle_t *pamh, int flags, int choice)
     /* call the list of module functions */
     pamh->choice = choice;
     retval = _pam_dispatch_aux(pamh, flags, h, resumed, use_cached_chain);
-    resumed = PAM_FALSE;
 
     __PAM_TO_APP(pamh);
 
--- a/modules/pam_ftp/pam_ftp.c
+++ b/modules/pam_ftp/pam_ftp.c
@@ -133,6 +133,8 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
 	retval = pam_set_item(pamh, PAM_USER, (const void *)anon_user);
 	if (retval != PAM_SUCCESS || anon_user == NULL) {
 	    pam_syslog(pamh, LOG_ERR, "user resetting failed");
+	    free(anon_user);
+
 	    return PAM_USER_UNKNOWN;
 	}
 	free(anon_user);
--- a/modules/pam_faillock/main.c
+++ b/modules/pam_faillock/main.c
@@ -157,7 +157,7 @@ do_user(struct options *opts, const char *user)
 		unsigned int i;
 
 		memset(&tallies, 0, sizeof(tallies));
-		if ((rv=read_tally(fd, &tallies)) == -1) {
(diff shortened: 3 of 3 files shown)


modules/pam_ftp/pam_ftp.c
```diff
@@ -133,6 +133,11 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
 	retval = pam_set_item(pamh, PAM_USER, (const void *)anon_user);
 	if (retval != PAM_SUCCESS || anon_user == NULL) {
 	    pam_syslog(pamh, LOG_ERR, "user resetting failed");
+
+        if (anon_user != NULL) {
+            free(anon_user);
+        }
```
ldv-alt, line 139: There is no need to check for NULL before free().

modules/pam_usertype/pam_usertype.c
```diff
@@ -121,6 +121,7 @@ pam_usertype_get_uid(struct pam_usertype_opts *opts,
         }
 
         *_uid = pwd->pw_uid;
+        free(pwd);
```
ldv-alt, line 124: I think this change introduces a double-free.
Why do you think the memory returned by pam_modutil_getpw* functions should be freed explicitly?

ldv-alt (changes requested):

modules/pam_ftp/pam_ftp.c
t8m: Also the indentation is off.

libpam/pam_dispatch.c
```diff
@@ -424,7 +424,6 @@ int _pam_dispatch(pam_handle_t *pamh, int flags, int choice)
     /* call the list of module functions */
     pamh->choice = choice;
     retval = _pam_dispatch_aux(pamh, flags, h, resumed, use_cached_chain);
-    resumed = PAM_FALSE;
```
ldv-alt, line 427: This might be a relic of defensive programming style when variables that are no longer expected to be used are assigned some safe defaults.  For example, ```close(fd), fd = -1```.
The best way to achieve this goal is not assigning safe defaults, though, but limiting the scope of variables, shifting the burden to the compiler.

modules/pam_usertype/pam_usertype.c
ikerexxe: I misunderstood the documentation. So, I've checked it with a simple example and yes, this would mean a double free. I'll remove it when I update the patches.

libpam/pam_dispatch.c
ikerexxe: This sounds like some defensive programming, but it makes no sense to keep it. I've checked and it also doesn't make any sense to limit the scope of the variable, since this would mean the creation of a new scope just to limit it. I still think that the best approach is to remote the statement.

modules/pam_ftp/pam_ftp.c
ikerexxe: Updated

t8m (approved):
LGTM now.

t8m:
Merged, thank you.

PR #289, opened 2020-11-04
pam_env: allow environment files without EOL at EOF

### LINKED ISSUE #263: pam_env: .pam_environment behaviour change in pam 1.4
.pam_environment containing the following single line with no end of line marker:
```
SSH_AUTH_SOCK DEFAULT="${XDG_RUNTIME_DIR}/ssh-agent.socket"
```
Previous to pam 1.4 was accepted.  With pam 1.4 it is rejected.  Debug from pam 1.4
```
[pam_env.c:_parse_config_file(138)] Config file name is: /home/testuser/.pam_environment
[pam_env.c:_assemble_line(294)] called.
[pam_env.c:_assemble_line(315)] _assemble_line: line too long
[pam_env.c:_parse_config_file(179)] Exit. 
[pam_env.c:handle_env(847)] Exit.
```
This change was introduced by 563d21d6dbb6d64613919ccb1cc939bae546baab.
There is no mention of this change in the NEWS file.  Is this intended?

Downstream bug report https://bugs.archlinux.org/task/67636

t8m:
Fixes #263

* modules/pam_env/pam_env.c (_assemble_line): Do not error out if at feof()

### DIFF
--- a/modules/pam_env/pam_env.c
+++ b/modules/pam_env/pam_env.c
@@ -311,7 +311,7 @@ static int _assemble_line(FILE *f, char *buffer, int buf_len)
 	    D(("_assemble_line: corrupted or binary file"));
 	    return -1;
 	}
-	if (p[strlen(p)-1] != '\n') {
+	if (p[strlen(p)-1] != '\n' && !feof(f)) {
 	    D(("_assemble_line: line too long"));
 	    return -1;
 	}


t8m:
@ldv-alt Please consider merging this before the release.

modules/pam_env/pam_env.c
```diff
@@ -311,7 +311,7 @@ static int _assemble_line(FILE *f, char *buffer, int buf_len)
 	    D(("_assemble_line: corrupted or binary file"));
 	    return -1;
 	}
-	if (p[strlen(p)-1] != '\n') {
+	if (p[strlen(p)-1] != '\n' && !feof(f)) {
```
ldv-alt, line 314: I don't think this would work, feof wouldn't return nonzero until fgets returned NULL.
I suppose you need a second fgets call.

t8m: No, at least according to my testing if fgets() returns because there is no EOL at the EOF, the feof() will return 1 immediately. There is no need for another fgets().

ldv-alt: OK then.

t8m:
Merged

PR #290, opened 2020-11-04
pam_env: deprecation notice of reading the user environment

t8m:
* modules/pam_env/pam_env.8.xml: Add the notice to the manual.
* modules/pam_env/pam_env.c (_pam_parse): Log deprecation warning
  if debug enabled.

Or should the deprecation warning be there also without the debug arg?

### DIFF
--- a/modules/pam_env/pam_env.8.xml
+++ b/modules/pam_env/pam_env.8.xml
@@ -163,6 +163,11 @@
             behavior of subsequent modules in the stack without the consent
             of the system administrator.
           </para>
+          <para>
+            Due to problematic security this functionality is deprecated
+            since the 1.5.0 version and will be removed completely at some
+            point in the future.
+          </para>
         </listitem>
       </varlistentry>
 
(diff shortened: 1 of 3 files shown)


t8m:
@ldv-alt ?

thkukuk:
The deprecation warning should be printed every time we read an user environment file, else nobody will notice it.

t8m:
OK, I've changed it accordingly.

modules/pam_env/pam_env.8.xml
```diff
@@ -163,6 +163,11 @@
             behavior of subsequent modules in the stack without the consent
             of the system administrator.
           </para>
+          <para>
+            Due to problematic security this functionality is deprecated
+            since the 1.5.0 version and will be removed completely at some
+            point in future.
```
ldv-alt, line 169: I'd suggest using "in the future" wording as in the NEWS entry.
Besides that, LGTM.

t8m: Done

septatrix:
I stumbled upon this in the man page and wonder what the recommended approach to setting user-specific environment variables is? Simply put them back into `~/.profile`?

t8m:
> Simply put them back into `~/.profile`?

I would recommend that. For systemd based user sessions, it could be also probably done via systemd configuration.

PR #292, opened 2020-11-05
libpam: add supplementary groups on priv drop

allisonkarlitskaya:
Replace the setgroups(0, NULL) call in pam_modutil_drop_priv() with a
call to initgroups().  This makes sure that the user's supplementary
groups are also configured.

This fixes the permission check in pam_motd: this feature was intended
to allow setting permissions on a motd file to prevent it from being
shown to users who are not a member of a particular group (for example,
wheel).

### DIFF
--- a/libpam/pam_modutil_priv.c
+++ b/libpam/pam_modutil_priv.c
@@ -107,11 +107,20 @@ int pam_modutil_drop_priv(pam_handle_t *pamh,
 	 * We should care to leave process credentials in consistent state.
 	 * That is, e.g. if change_gid() succeeded but change_uid() failed,
 	 * we should try to restore old gid.
+	 *
+	 * We try to add the supplementary groups on a best-effort
+	 * basis.  If it fails, it's not fatal: we fall back to using an
+	 * empty list.
 	 */
-	if (setgroups(0, NULL)) {
-		pam_syslog(pamh, LOG_ERR,
-			   "pam_modutil_drop_priv: setgroups failed: %m");
-		return cleanup(p);
+	if (initgroups(pw->pw_name, pw->pw_gid)) {
+		pam_syslog(pamh, LOG_WARNING,
+			   "pam_modutil_drop_priv: initgroups failed: %m");
+
+		if (setgroups(0, NULL)) {
+			pam_syslog(pamh, LOG_ERR,
+				   "pam_modutil_drop_priv: setgroups failed: %m");
+			return cleanup(p);
+		}
 	}
 	if (change_gid(pw->pw_gid, &p->old_gid)) {
 		pam_syslog(pamh, LOG_ERR,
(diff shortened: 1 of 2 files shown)


allisonkarlitskaya:
Note: this is strictly an API break.  It's hard to imagine this feature presenting problems for any user, but I'd also be happy to write a patch introducing a new function with the `initgroups()` functionality while maintaining the existing behaviour with the old API.

ldv-alt:
On Thu, Nov 05, 2020 at 05:24:38AM -0800, Allison Karlitskaya wrote:
> Note: this is strictly an API break.  It's hard to imagine this feature presenting problems for any user, but I'd also be happy to write a patch introducing a new function with the `initgroups()` functionality while maintaining the existing behaviour with the old API.

No, I wouldn't call it an API break: this change doesn't affect the
programming interface, it also doesn't affect the binary interface.

Yes, this is a change in behavior, but do we have to maintain the current
behaviour in any form?  I'm not sure why we would have to.

The side effect of initgroups call is that its implementation has to
access the group database.  I've checked all pam_modutil_drop_priv users
in the linux-pam tree, all of them invoke it either shortly after
pam_modutil_getpwnam or in other contexts where side effects of group
database access is not an issue.

I doubt that out-of-tree users of pam_modutil_drop_priv would have any
issues with the proposed change.

Maybe I'm missing something, but so far I don't see any reason why we
shouldn't just replace setgroups with initgroups as your patch does.

t8m:
The only reason not to do the replacement is potential for failure and timing effects of the group database access. I am not sure it is serious that much.

t8m:
Maybe instead of erroring out just fallback to setgroups call. You can keep the logging of the error.

allisonkarlitskaya:
> Maybe instead of erroring out just fallback to setgroups call. You can keep the logging of the error.

Sure.  I'll do another patch :)

Thanks for the input, everyone!

ldv-alt:
On Fri, Nov 06, 2020 at 12:48:58AM -0800, Allison Karlitskaya wrote:
> > Maybe instead of erroring out just fallback to setgroups call. You can keep the logging of the error.
> 
> Sure.  I'll do another patch :)

Thanks!  Would you suggest to include this patch in the upcoming 1.5.0 release?

ldv-alt:
Looks good, thanks.
Could you add a line to the NEWS file mentioning the use of initgroups in pam_modutil_drop_priv, please?

libpam/pam_modutil_priv.c
```diff
@@ -107,11 +107,20 @@ int pam_modutil_drop_priv(pam_handle_t *pamh,
...
 	 * That is, e.g. if change_gid() succeeded but change_uid() failed,
 	 * we should try to restore old gid.
+	 *
+	 * We try to add the supplementary groups on a best-effort
+	 * basis.  If it fails, it's not fatal: we fall back to using an
+	 * empty list.
 	 */
-	if (setgroups(0, NULL)) {
+	if (initgroups(pw->pw_name, pw->pw_gid)) {
 		pam_syslog(pamh, LOG_ERR,
```
t8m, line 116: Also, could you, please, change the log level to LOG_WARNING here?

t8m (approved):

ldv-alt:
Merged, thanks!

PR #298, opened 2020-11-19
pam_unix: Don't check for a blank password against root if the real user doesn't exist

### LINKED ISSUE #284: pam_unix(dovecot:auth): user [...] has blank password; authenticated without it
Hi,

since yesterday my mail server is sending massive spam and I try to figure out if my account has been hacked or there is an configuration failure with my server (postfix+dovecot running on arch). I found the following message in the logs even after changing the password for the user "USER":

pam_unix(dovecot:auth): user [USER@domain.com] has blank password; authenticated without it

and it seems that after this postfix allows to relay the incoming "spam" mails. 

btw. postfix is using dovecot as SASL provider and dovecot is configured to use PAM. PAM settings are default, i.e. 

using swaks-tool I was not able to generate same message on server side - even If I keep password empty. Any help is much appreciated. First question is: in what circumstances will the server generate this message and how can I reproduce it?

kyrias:
Otherwise this lets non-existent users authenticate without a password
if root doesn't have a password set while the nullok option is used.

This is is normally not a problem, but becomes a problem when the PAM
config also contains the pam_permit module.

Fixes #284.

### DIFF
--- a/modules/pam_unix/support.c
+++ b/modules/pam_unix/support.c
@@ -602,7 +602,8 @@ _unix_blankpasswd (pam_handle_t *pamh, unsigned long long ctrl, const char *name
 	int daysleft;
 	int retval;
 	int execloop = 1;
-	int nonexistent = 1;
+	int pwd_is_null;
+	int user_exists = 1;
 
 	D(("called"));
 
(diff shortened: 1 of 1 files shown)


t8m:
OH, good catch although in that situation the root account is wide open anyway. But it is a security issue anyway as root account might be blocked anyway.

t8m:
I have alternative fix that should make the code more clear in #300.
I have mentioned you in the commit message to give you the credits for finding the cause.

t8m:
I have merged the alternative fix.

PR #300, opened 2020-11-20
Second blank check with root for non-existent users must never return 1

### LINKED ISSUE #284: pam_unix(dovecot:auth): user [...] has blank password; authenticated without it

t8m:
The blank check could return 1 if root had empty password
because in the second case the password hash of root was used. We now
always return 0 in this case.

Fixes #284

* modules/pam_unix/support.c (_unix_blankpasswd): Make the loop
to cover the complete blank check so both existing and non existing
cases are identical except for the possible return value.

### DIFF
--- a/modules/pam_unix/support.c
+++ b/modules/pam_unix/support.c
@@ -601,8 +601,9 @@ _unix_blankpasswd (pam_handle_t *pamh, unsigned long long ctrl, const char *name
 	char *salt = NULL;
 	int daysleft;
 	int retval;
-	int execloop = 1;
-	int nonexistent = 1;
+	int blank = 0;
+	int execloop;
+	int nonexistent_check = 1;
 
 	D(("called"));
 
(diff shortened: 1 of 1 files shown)


t8m:
@ldv-alt although the security aspect of the issue is relatively minor - affecting systems that are already seriously weak by having blank root password I suppose we will need 1.5.1 release.

ldv-alt:
The loop was introduced by commit af0faf666c5008e54dfe43684f210e3581ff1bca
a.k.a. v1.5.0~80, I suppose it should be mentioned in the commit message, e.g.

```
Fixes: af0faf66 ("pam_unix: avoid determining if user exists")
```

ldv-alt:
Could we just avoid assigning anything to ```blank``` on the second iteration?

t8m:
> Could we just avoid assigning anything to `blank` on the second iteration?

I am a little bit concerned about compiler optimization to remove some parts of the second iteration then. Although the compiler probably cannot know if the get_pwd_hash and/or _unix_run_helper_binary has any side effects. So I prefer keeping it like this.

t8m:
> The loop was introduced by commit [af0faf6](https://github.com/linux-pam/linux-pam/commit/af0faf666c5008e54dfe43684f210e3581ff1bca)
> a.k.a. v1.5.0~80, I suppose it should be mentioned in the commit message, e.g.
> 
> ```
> Fixes: af0faf66 ("pam_unix: avoid determining if user exists")
> ```

I've amended the commit message to clarify.

ldv-alt:
On Fri, Nov 20, 2020 at 05:47:05AM -0800, Tomáš Mráz wrote:
> > Could we just avoid assigning anything to `blank` on the second iteration?
> 
> I am a little bit concerned about compiler optimization to remove some parts of the second iteration then. Although the compiler probably cannot know if the get_pwd_hash and/or _unix_run_helper_binary has any side effects. So I prefer keeping it like this.

Well, if the compiler can optimize out such calls as get_pwd_hash
or _unix_run_helper_binary, then this blank check would be the least
of our troubles.

t8m:
Sure, but it could at least optimize out the strlen in the second iteration - imo that is already possible.

t8m:
@ldv-alt so is it OK to merge?

ldv-alt (approved):
Although the new code is not an easy read, I think it should work.

t8m:
Merged, thank you for the review

PR #302, opened 2020-11-23
pam_wheel: Use pam_modutil_user_in_group_uid_gid instead of reimplementation

### LINKED ISSUE #297: OPT pam_wheel for checking the user in AD(Active Directory)
I login linux via AD authentication.

For example, a user named "jin", group named "xts", both are in AD:

```
# id jin    <==== "jin" is a AD user
uid=30000(jin) gid=30000(domain users) groups=30000(domain users),30001(xts),30005(yjs)
```

a local user named "test", local group named "xts":

```
# id test
uid=1000(test) gid=1000(test) groups=1000(test),1003(xts)
```

If i config `/etc/pam/d/su` like:

```
...
# only permit the users which is member of xts can "su root"
auth            required        pam_wheel.so use_uid group=xts root_only
...
```

For local user who is member of xts, `su - root` is OK

But for AD user who is member of xts, `su - root` got permission denied.

So, why not check a user which group it belong to like the `id` command, but not like this:

```
...
    if (is_on_list(grp->gr_mem, fromsu) || (tpwd->pw_gid == grp->gr_gid)) {
...
```

because when get the group's info, if the group is in AD, can not get the members of the group~

Thanks~

t8m:
The pam_modutil_user_in_group... functions use getgrouplist to check
the membership so they work also in setups with remote services which do
not provide group members in struct group.

Fixes #297

* modules/pam_wheel/pam_wheel.c (perform_check): Call pam_modutil_user_in_group_uid_gid
  to do the group check.

### DIFF
--- a/modules/pam_wheel/pam_wheel.c
+++ b/modules/pam_wheel/pam_wheel.c
@@ -44,17 +44,6 @@
 #include <security/pam_ext.h>
 #include "pam_inline.h"
 
-/* checks if a user is on a list of members of the GID 0 group */
-static int is_on_list(char * const *list, const char *member)
-{
-    while (list && *list) {
-        if (strcmp(*list, member) == 0)
-            return 1;
-        list++;
-    }
-    return 0;
-}
-
 /* argument parsing */
 
 #define PAM_DEBUG_ARG       0x0001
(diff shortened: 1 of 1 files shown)


ldv-alt (approved):
This change looks good, but
I think getgrnam/getgrnam_r should return the list of group members,
and those network services that don't provide this information are wrong
because one cannot patch all getgrnam users to use getgrouplist.

t8m:
Yep such setups will always be problematic, but there is no point in rejecting to add the getgrouplist call when we have it already conveniently available via the pam_modutil_user_in_group...

t8m:
Merged manually.

PR #303, opened 2020-11-24
Fix memory leak that occurs in a rare situations

TomMD:
If crypt_r() fails do not leak cdata.

Bug found with Muse.dev CI tool (specifically, the Infer static analyzer).  Muse's full results are [available](https://console.muse.dev/result/TomMD/linux-pam/01EQY5Q0JNQS8QPHRM5Z6EH7BK?search=bigc&tab=results) but if you are interested I could make a PR adding [the Muse CI configuration](https://github.com/TomMD/linux-pam/tree/master/.muse) to run Infer on each pull request and produce a github comment if there are any new bugs.

### DIFF
--- a/modules/pam_unix/bigcrypt.c
+++ b/modules/pam_unix/bigcrypt.c
@@ -111,6 +111,9 @@ char *bigcrypt(const char *key, const char *salt)
 #endif
 	if (tmp_ptr == NULL) {
 		free(dec_c2_cryptbuf);
+#ifdef HAVE_CRYPT_R
+		free(cdata);
+#endif
 		return NULL;
 	}
 	/* and place in the static area */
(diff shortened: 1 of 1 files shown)


ldv-alt (approved):
The patch looks good, thanks.
I've changed the commit message a bit.

As a side note, I wonder why in passverify.c ```cdata``` is cleared before being freed, but here in bigcrypt.c it isn't.

TomMD:
Good question. Not sure about passverify.c's reasons but it is using `memset` and not `memset_s` so the chances are that the assembly doesn't actually do any memory clearing anyway and the behavior matches between the two.

Would you be interested in installing [muse](https://github.com/apps/muse-dev)?  The configuration files I made aren't quite right for merge because the difference in ci platform, but it isn't hard to fix up.

t8m (approved):

ldv-alt:
> Good question. Not sure about passverify.c's reasons but it is using `memset` and not `memset_s` so the chances are that the assembly doesn't actually do any memory clearing anyway and the behavior matches between the two.

Yes, all these traditional memory clearing techniques need a review.

> Would you be interested in installing [muse](https://github.com/apps/muse-dev)? The configuration files I made aren't quite right for merge because the difference in ci platform, but it isn't hard to fix up.

Sorry, I'm not familiar with muse.  Could you open a separate PR for that?

PR #308, opened 2020-12-02
Clarify the effect of 'done' in documentation

### LINKED ISSUE #307: Stack processing continues through done if impression is negative
The documented definition of action `done` is
>   equivalent to ok with the side effect of terminating the module stack and PAM immediately returning to the application.

However, in certain circumstances the stack will continue beyond `done`.

Example service configuration, "pambug":
```
# pam_deny returns PAM_AUTH_ERR
account  required                      pam_deny.so

# pam_permit returns PAM_SUCCESS
account  [success=done default=die]    pam_permit.so
# Q: why don't we stop here?
# the done / die options in the line above are both "decisive"

# if the pam_deny line above is commented out, this pam_echo is not invoked
account  optional                      pam_echo.so Unreachable
````
What happens?
The `pam_echo` line is executed.

What is expected?
The `pam_permit` line should stop execution.
The presence or absence of the `pam_deny` line should not affect whether `pam_echo` is called.

### PR DESCRIPTION
**Pull request by t8m**
The done action does not terminate the stack processing in case
there is a failing module with bad action up in the stack.

* doc/man/pam.conf-syntax.xml: Clarify the effect of 'done'.

Fixes #307

### DIFF
--- a/doc/man/pam.conf-syntax.xml
+++ b/doc/man/pam.conf-syntax.xml
@@ -273,8 +273,9 @@
         <term>die</term>
         <listitem>
            <para>
-             equivalent to bad with the side effect of terminating the
-             module stack and PAM immediately returning to the application.
+             equivalent to <emphasis>bad</emphasis> with the side effect of
+             terminating the module stack and PAM immediately returning to
+             the application.
           </para>
         </listitem>
       </varlistentry>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: doc/man/pam.conf-syntax.xml [hunk 1]
```diff
@@ -297,8 +298,10 @@
         <term>done</term>
         <listitem>
            <para>
-             equivalent to ok with the side effect of terminating the module
-             stack and PAM immediately returning to the application.
+             equivalent to <emphasis>ok</emphasis> with the side effect of
+             terminating the module stack and PAM immediately returning to the
+             application in case there was no non-ignored module failure
+             before.
```
ldv-alt, line 304: I suggest a slightly different wording here:
```unless there was a non-ignored module failure before```
instead of
```in case there was no non-ignored module failure before```

t8m: Fixed

ldv-alt (approved):

PR #313, opened 2020-12-23
pam_umask: fix handling of umask parameter

ldv-alt:
Potential failures of strdup(3) were ignored, fix this by not using
strdup(3) at all.

* modules/pam_umask/pam_umask.c (struct options_t): Add const to umask
field, add login_umask field.
(parse_option): Do not use strdup.
(get_options): Assign pam_modutil_search_key return values
to options->login_umask.
(pam_sm_open_session): Free options.login_umask instead of
options.umask.


t8m (approved):
LGTM, please merge.

PR #314, opened 2020-12-23
pam_mkhomedir: use UMASK from /etc/login.defs

dshein-alt:
Follow the example of useradd(8) and set the file mode creation mask to
the value of UMASK configuration item from /etc/login.defs when umask=
option is not specified.

### DIFF
--- a/modules/pam_mkhomedir/mkhomedir_helper.8.xml
+++ b/modules/pam_mkhomedir/mkhomedir_helper.8.xml
@@ -47,7 +47,7 @@
       The default value of <replaceable>umask</replaceable> is 0022 and the
       default value of <replaceable>path-to-skel</replaceable> is
       <emphasis>/etc/skel</emphasis>. The default value of
-      <replaceable>home_mode</replaceable> is computed form the value of
+      <replaceable>home_mode</replaceable> is computed from the value of
       <replaceable>umask</replaceable>.
     </para>
 
--- a/modules/pam_mkhomedir/mkhomedir_helper.8.xml
+++ b/modules/pam_mkhomedir/mkhomedir_helper.8.xml
@@ -25,6 +25,9 @@
         <replaceable>umask</replaceable>
       <arg choice="opt">
         <replaceable>path-to-skel</replaceable>
+      <arg choice="opt">
+        <replaceable>home_mode</replaceable>
+      </arg>
       </arg>
       </arg>
     </cmdsynopsis>
@@ -43,7 +46,9 @@
     <para>
       The default value of <replaceable>umask</replaceable> is 0022 and the
       default value of <replaceable>path-to-skel</replaceable> is
-      <emphasis>/etc/skel</emphasis>.
+      <emphasis>/etc/skel</emphasis>. The default value of
+      <replaceable>home_mode</replaceable> is computed form the value of
+      <replaceable>umask</replaceable>.
     </para>
 
     <para>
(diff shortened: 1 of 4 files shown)


ldv-alt (approved):

t8m:
Please note that recently the shadow utilities added a new option HOME_MODE to login.defs which overrides the UMASK for the home directory creation. I think it should be implemented here as well.

t8m:
There is one caveat - the shadow utilities do not apply the HOME_MODE and the UMASK to the files copied from /etc/skel. It is questionable whether the pam_mkhomedir should continue doing it or not. The HOME_MODE should definitely be applied only to the mode of the home directory itself.

ldv-alt:
> Please note that recently the shadow utilities added a new option HOME_MODE to login.defs which overrides the UMASK for the home directory creation. I think it should be implemented here as well.

I think it would make sense to add HOME_MODE support but it should rather be done in a separate follow-up patch.

dshein-alt:
I've updated code to follow shadow/useradd behavior in case of umask option was not specified for pam_umask, so permissions for created user home directory is set in accordance to value of HOME_MODE and preempts UMASK if both are defined if /etc/login.defs. At the same time permissons for files and folders that copied fron /etc/skel is masked with default umask value.

Now home directory for users that loged in to system using pam creates with same permissions as for local one.

t8m:
We do not allow merge commits. Could you please just cleanup them and rebase your changes on top of the current master?

dshein-alt:
Got it.
I'll clean it up

modules/pam_mkhomedir/pam_mkhomedir.8.xml
```diff
@@ -91,8 +91,14 @@
         <listitem>
           <para>
             The file mode creation mask is set to
-            <replaceable>mask</replaceable>. The default value of mask is
-            0022.
+            <replaceable>mask</replaceable>. The default value of mask
```
t8m, line 94: Please use something like:
```
           The default value of mask
            is 0022. If this option is not specified, then the permissions
            of the created user home directory is set to the value of
            <option>HOME_MODE</option> configuration item from
            <filename>/etc/login.defs</filename>. If there is no such
            configuration item then the value is computed from the
            value of <option>UMASK</option> in the same file. If
            there is no such configuration option either the default
            value of 0755 is used for the mode.
```

t8m (changes requested):
I am sorry for neglecting your PR for so long time. Can you please adjust the documentation? Otherwise it looks good to me.

modules/pam_mkhomedir/pam_mkhomedir.8.xml
dshein-alt: Updated documentation for pam_mkhomedir and mkhomedir_helper

modules/pam_mkhomedir/mkhomedir_helper.8.xml
```diff
@@ -43,7 +46,9 @@
     <para>
       The default value of <replaceable>umask</replaceable> is 0022 and the
       default value of <replaceable>path-to-skel</replaceable> is
-      <emphasis>/etc/skel</emphasis>.
+      <emphasis>/etc/skel</emphasis>. The default value of
+      <replaceable>home_mode</replaceable> is computed form the value of
```
t8m, line 50: Typo: from

t8m (changes requested):
One last nit

modules/pam_mkhomedir/mkhomedir_helper.8.xml
dshein-alt: Fixed.
Thank you!

t8m (approved):

t8m:
Squashed and merged. Thank you for the contribution!

ldv-alt:
Sorry, I missed the second iteration completely,
I really don't like the way the author name and email were mangled as result of squashing.
If it was github that failed to do a proper squashing, we should use this github feature with care.

dshein-alt:
I suppose that name and email  wasn't fetched during squashing due to my Github profile settings.
I've probably missed some steps during registration.

PR #319, opened 2021-01-12
faillock: create dir before create tally file

sandy-lcq:
the default tallydir is "/var/run/faillock", and usually,
/var/run is mounted on tempfs. so faillock maybe not existed.
When we set flag to O_CREAT, and hope to create tallyfile,
but since the dir is not existed, open function will failed,
and faillock not works well. so fixed by when we want to create
tallyfile, and the dir is not existed, create it first.

Signed-off-by: Changqing Li <changqing.li@windriver.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -613,6 +613,16 @@ test -n "$opt_kerneloverflowuid" ||
           opt_kerneloverflowuid=65534
 AC_DEFINE_UNQUOTED(PAM_USERTYPE_OVERFLOW_UID, $opt_kerneloverflowuid, [Kernel overflow uid.])
 
+AC_ARG_WITH([systemdunitdir],
+AS_HELP_STRING([--with-systemdunitdir=DIR], [path to systemd service directory]),
+    [],
+    [
+    PKG_CHECK_EXISTS([systemd],
+      [with_systemdunitdir=$($PKG_CONFIG --variable=systemdunitdir systemd)],
+      [with_systemdunitdir='${prefix}/lib/systemd/system'])
+    ])
+AC_SUBST([systemdunitdir], [$with_systemdunitdir])
+
 AC_ARG_ENABLE([unix],
               [AS_HELP_STRING([--disable-unix],
                               [do not build pam_unix module])],
--- a/modules/pam_faillock/faillock.c
+++ b/modules/pam_faillock/faillock.c
@@ -74,6 +74,9 @@ open_tally (const char *dir, const char *user, uid_t uid, int create)
 
 	if (create) {
 		flags |= O_CREAT;
+		if (access(dir, F_OK) != 0) {
+			mkdir(dir, 0755);
+		}
 	}
 
 	fd = open(path, flags, 0660);
--- a/modules/pam_namespace/Makefile.am
+++ b/modules/pam_namespace/Makefile.am
@@ -18,7 +18,7 @@ TESTS = $(dist_check_SCRIPTS)
 securelibdir = $(SECUREDIR)
 secureconfdir = $(SCONFIGDIR)
 namespaceddir = $(SCONFIGDIR)/namespace.d
-servicedir = $(prefix)/lib/systemd/system
+servicedir = $(systemdunitdir)
 
 AM_CFLAGS = -I$(top_srcdir)/libpam/include -I$(top_srcdir)/libpamc/include \
         -DSECURECONF_DIR=\"$(SCONFIGDIR)/\" $(WARN_CFLAGS)


configure.ac
```diff
@@ -613,6 +613,16 @@ test -n "$opt_kerneloverflowuid" ||
           opt_kerneloverflowuid=65534
 AC_DEFINE_UNQUOTED(PAM_USERTYPE_OVERFLOW_UID, $opt_kerneloverflowuid, [Kernel overflow uid.])
 
+AC_ARG_WITH([systemduserunitdir],
+AS_HELP_STRING([--with-systemduserunitdir=DIR], [Directory for systemd user service files]),
+    [],
+    [
+    PKG_CHECK_EXISTS([systemd],
+      [with_systemduserunitdir=$($PKG_CONFIG --variable=systemduserunitdir systemd)],
+      [with_systemduserunitdir='${libdir}/systemd/user'])
```
t8m, line 622: The libdir can be /usr/lib64 for example. So I do not think it is right to be the default here.

sandy-lcq: Thanks. I updated it to ${prefix}/lib/systemd/user.

configure.ac
```diff
@@ -613,6 +613,16 @@ test -n "$opt_kerneloverflowuid" ||
           opt_kerneloverflowuid=65534
 AC_DEFINE_UNQUOTED(PAM_USERTYPE_OVERFLOW_UID, $opt_kerneloverflowuid, [Kernel overflow uid.])
 
+AC_ARG_WITH([systemduserunitdir],
```
t8m, line 616: Actually this is still wrong - this should not be a systemd user unit dir but systemd system unit dir. ${prefix}/lib/systemd/system
And everything needs to be renamed accordingly.

sandy-lcq: Thanks.  Updated according to the comments.

t8m (approved):
LGTM now

modules/pam_faillock/faillock.c
```diff
@@ -74,6 +74,9 @@ open_tally (const char *dir, const char *user, uid_t uid, int create)
 
 	if (create) {
 		flags |= O_CREAT;
+		if (access(dir, F_OK) != 0) {
+			mkdir(dir, 0755);
+		}
```
ldv-alt, line 79: Do we really have to test the path before mkdir invocation?

ldv-alt (changes requested):
The commit message of the first commit is hard to read because of multiple grammar mistakes, please fix.

modules/pam_faillock/faillock.c
t8m: I prefer keeping the test there. The code is racy anyway even without it and the added access call should not cause any performance issue.

sandy-lcq:
Hi, all

Thanks. Updated the commits and fixed all the comments.

t8m (approved):

ldv-alt:
On Thu, Jan 14, 2021 at 12:46:30AM -0800, Sandy wrote:
> Hi, all
> 
> Thanks. Updated the commits and fixed all the comments.

OK, let's go through these two commit messages.

> faillock: create tallydir before create tallyfile

This is not OK because "before create" is not a grammatically correct
phrase.

> The default tallydir is "/var/run/faillock", and this default
> tallydir may not exist.

This is OK.

> When create tallyfile, function open may fail since tallydir not
> exist. Therefore faillock not works well.

This is not OK because "When create", "tallydir not exist",
and "not works" are not grammatically correct phrases.

> fixed by creating the tallydir before create tallyfile when the
> tallydir not exist.

This is not OK because "before create" and "tallydir not exist" are not
grammatically correct phrases.  Also, "fixed" places the whole sentence
in a grammatical tense that's different from all the rest.

> add option for setting systemd service directory

Please prefix the line with configure: or configure.ac:
Consider mentioning the option name right away, e.g.
add --with-systemdunitdir option

> Previously, servicedir is set as '$(prefix)/lib/systemd/system'.

This is not OK because "Previously ... is" and "set as '...'" are not
grammatically correct phrases.

> It cannot meet the requirements, such as:

It's not "the requirements", it's configuration or setup or something else,
please use a different wording.

> So add an option for setting systemd service directory.
> User can set its own servicedir by giving this option.

The second sentence looks redundant to me.

> Besides, the default behavior is changed, if this option is not
> given, the servicedir will first try to get value from systemd
> pkg-config file, then the default value '$(prefix)/lib/systemd/system'.

Sorry, "the servicedir" is a variable, it's not capable of trying,
please use a different wording.

sandy-lcq:
@ldv-alt  Thanks for your detail comments. I have update the commit messages. Please help to check if it is ok now.

ldv-alt:
On Mon, Jan 18, 2021 at 08:09:12PM -0800, Sandy wrote:
> @ldv-alt  Thanks for your detail comments. I have update the commit messages. Please help to check if it is ok now. 

> faillock: create tallydir before creating tallyfile

OK

> The default tallydir is "/var/run/faillock", and this default
> tallydir may not exist.

OK

> Function open may fail as tallydir is not exist when creating
> the tallyfile. Therefore, faillock will not work well.

"as tallydir is not exist" is not grammatically correct.

> Fix this problem by creating tallydir before creating tallyfile
> when the tallydir is not exist.

"is not exist" is not grammatically correct.

> configure.ac: add --with-systemdunitdir option

OK

> * Add this option to support below scenario:

"below scenario" is not grammatically correct, try "the following
scenario" instead.

>   prefix = '/usr'
>   servicedir = '/lib/systemd/system'
> 
> * The default behavior is changed
>   If this option is not given, servicedir will set to value that is

"will set to value that is" is not grammatically correct, did you mean
"will be set to the value"?

> getted from systemd pkg-config file. If we cannot get value from
> systemd pkg-config file, servicedir will set to the default value
> '$(prefix)/lib/systemd/system'

"getted" is not grammatically correct, did you mean "obtained"?
"will set to" is not grammatically correct, did you mean
"will be set to"?

"If we cannot get value from systemd pkg-config file" - who are we?
Try a different wording, for example,
"If the value cannot be obtained".

sandy-lcq:
Fix the commit message according to the comments.  Great thanks to @ldv-alt.

ldv-alt:
On Tue, Jan 19, 2021 at 06:05:26PM -0800, Sandy wrote:
> Fix the commit message according to the comments.  Great thanks to @ldv-alt.

OK, we are making progress.

> * Add this option to support the following scenario:
>   prefix = '/usr'
>   servicedir = '/lib/systemd/system'

OK (it's probably better not to indent, though).

> * The default behavior is changed

Please add the trailing colon as you did above.

>   If this option is not given, servicedir will be set to value that is

"to value" -> "to the value".

It's probably better not to indent the first line of this paragraph.

> obtained from systemd pkg-config file. If the value cannot be obtained
> from systemd pkg-config file, servicedir will be set to the default value

there is no need to repeat "from systemd pkg-config file"

> '$(prefix)/lib/systemd/system'

Please add the trailing dot.

sandy-lcq:
Fix comments from @ldv-alt, thanks.

ldv-alt (approved):

t8m (approved):

t8m:

PR #320, opened 2021-01-12
man: fix build if man pages cannot be re-generated

eremin-ev:
Build stops with error if man pages cannot be regenerated from XML
sources (e.g. DocBook stylesheets check fails) or if the following
options have been passed to configure:
```
./autogen.sh &&
./configure --enable-doc --disable-regenerate-docu &&
make -j8
...
make[3]: *** No rule to make target 'pam_unix.8', needed by 'all-am'.  Stop.
make[3]: Leaving directory '/tmp/m/bp/modules/pam_unix'
make[2]: *** [Makefile:490: all-recursive] Error 1
make[2]: Leaving directory '/tmp/m/bp/modules'
make[1]: *** [Makefile:489: all-recursive] Error 1
make[1]: Leaving directory '/tmp/m/bp'
make: *** [Makefile:419: all] Error 2
```
The patch fixes this issue and seems doesn't break other combinations:
```
enable-doc   enable-regenerate-docu  OK
enable-doc  disable-regenerate-docu  OK (fails without the patch)
disable-doc  enable-regenerate-docu  OK
disable-doc disable-regenerate-docu  OK
```
Please review.

### DIFF
--- a/doc/man/Makefile.am
+++ b/doc/man/Makefile.am
@@ -7,24 +7,6 @@ MAINTAINERCLEANFILES = $(MANS)
 
 EXTRA_DIST = $(MANS) $(XMLS)
 
-man_MANS = pam.3 PAM.8 pam.8 pam.conf.5 pam.d.5 \
-	pam_acct_mgmt.3 pam_authenticate.3 \
-	pam_chauthtok.3 pam_close_session.3 pam_conv.3 \
-	pam_end.3 pam_error.3 \
-	pam_fail_delay.3 pam_xauth_data.3 \
-	pam_get_authtok.3 pam_get_authtok_noverify.3 pam_get_authtok_verify.3 \
-	pam_get_data.3 pam_get_item.3 pam_get_user.3 \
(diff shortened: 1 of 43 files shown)


t8m:
I do not think this is correct patch. This will also disable installing the pre-built manpages with --enable-doc unless you also have --enable-regenerate-docu.

What you're running the commands on? Is it an unpacked release tarball or just a git checkout?

eremin-ev:
Ah, I see.  I'm running this on a checked out git repo, where no prebuilt man pages available:
```
ls modules/pam_unix/pam_unix.8 
ls: cannot access 'modules/pam_unix/pam_unix.8': No such file or directory
```
But I didn't take into account a tarball with prebuilt man pages.  Will think about this case.  Thanks!

t8m:
If you do not want to build the documentation on a checkout, you need to use --disable-doc.
If you do not want to build the documentation but you want it installed, you need to build and install a released tarball.
If you want to install the documentation and you want to do it on a checkout, you need to have the tools to build the documentation.

There is no other way.

PR #328, opened 2021-01-26
pam_securetty: don't complain about missing config

lnussel:
Not shipping a config file should be perfectly valid for distros while
still having eg login pre-configured to honor securetty when present.
PAM itself doesn't ship any template either. So avoid spamming the log
file if /etc/securetty wasn't found.

### DIFF
--- a/modules/pam_securetty/pam_securetty.c
+++ b/modules/pam_securetty/pam_securetty.c
@@ -111,7 +111,8 @@ securetty_perform_check (pam_handle_t *pamh, int ctrl,
 #ifdef VENDORDIR
       if (errno == ENOENT) {
 	if (stat(SECURETTY2_FILE, &ttyfileinfo)) {
-	  pam_syslog(pamh, LOG_NOTICE,
+	  if (ctrl & PAM_DEBUG_ARG)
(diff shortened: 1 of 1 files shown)


t8m (changes requested):
Could you please keep the message conditional on `if (ctrl & PAM_DEBUG_ARG)` possibly with downgrading the priority to LOG_DEBUG?

lnussel:
Ah, right. Makes sense. Will do!

t8m (approved):

ldv-alt (approved):

t8m:

zeha:
#158 was an earlier variant of this, which got shot down. Are the original arguments still valid or not?

t8m:
IMO the old behavior of returning success but logging an error is weird. That helps nobody. Also sometimes it is better to not manipulate the pam.d stacks if you have some preconfigured with a distro or a package. So yeah, you could remove pam_securetty from preconfigured configs to silence the syslog warnings but I do not think we should encourage innocent sysadmins doing that.

Also this change still allows for debugging messages by the debug argument.

PR #338, opened 2021-03-17
pam_faillock:

qcq:
convert spaces to tab for code style
refactor fragment of code to remove one macro

### DIFF
--- a/modules/pam_faillock/main.c
+++ b/modules/pam_faillock/main.c
@@ -70,14 +70,13 @@ args_parse(int argc, char **argv, struct options *opts)
 	opts->progname = argv[0];
 
 	for (i = 1; i < argc; ++i) {
-
 		if (strcmp(argv[i], "--dir") == 0) {
 			++i;
 			if (i >= argc || strlen(argv[i]) == 0) {
 				fprintf(stderr, "%s: No directory supplied.\n", argv[0]);
 				return -1;
 			}
-		        opts->dir = argv[i];
+			opts->dir = argv[i];
 		}
 		else if (strcmp(argv[i], "--user") == 0) {
 			++i;
@@ -85,7 +84,7 @@ args_parse(int argc, char **argv, struct options *opts)
 				fprintf(stderr, "%s: No user name supplied.\n", argv[0]);
 				return -1;
 			}
-		        opts->user = argv[i];
+			opts->user = argv[i];
 		}
 		else if (strcmp(argv[i], "--reset") == 0) {
 			opts->reset = 1;


modules/pam_faillock/main.c
```diff
@@ -129,17 +128,14 @@ do_user(struct options *opts, const char *user)
...
-#endif
-
 		while ((rv=ftruncate(fd, 0)) == -1 && errno == EINTR);
 		if (rv == -1) {
 			fprintf(stderr, "%s: Error clearing the tally file for %s:",
 				opts->progname, user);
 			perror(NULL);
 #ifdef HAVE_LIBAUDIT
 		}
+		int audit_fd;
```
t8m, line 138: We do not allow variable declarations in the middle of block. This has to stay as it was.

modules/pam_faillock/main.c
```diff
@@ -201,8 +197,9 @@ do_allusers(struct options *opts)
 	for (i = 0; i < rv; i++) {
 		if (userlist[i]->d_name[0] == '.') {
 			if ((userlist[i]->d_name[1] == '.' && userlist[i]->d_name[2] == '\0') ||
-			    userlist[i]->d_name[1] == '\0')
-				continue;
+				userlist[i]->d_name[1] == '\0') {
+					continue;
```
t8m, line 201: I do not really see this as an improvement.

t8m (changes requested):

t8m:
Also the subject line of the commit should contain some meaningful short description. Not just `pam_faillock:`

modules/pam_faillock/main.c
qcq: ok, however in my view can remove one macro

modules/pam_faillock/main.c
qcq: convert the spaces to tab, also add pair {} to continue statement.

t8m (approved):
OK now.

modules/pam_faillock/main.c
```diff
@@ -201,8 +200,9 @@ do_allusers(struct options *opts)
...
+				}
```
ldv-alt, line 205: I don't mind adding curly brackets if you think it could help, but the new indentation is misleading.

ldv-alt (changes requested):

modules/pam_faillock/main.c
qcq: the original indentation is space, just convert that to tab, which can be told by the line 204.

ldv-alt: As you can see, `userlist[i]->d_name[1] == '\0'` is indented under `(userlist[i]->d_name[1] == '.' && userlist[i]->d_name[2] == '\0')`, so this conversion from spaces to tab breaks it.
Also, this conversion changes the indentation of `continue` from 4 tabs to 5 tabs, and indents the closing `}` one tab more that the matching `if` statement.  Both of these changes are misleading.
Could you leave this hunk unchanged, please?

qcq: sure

t8m (approved):

ldv-alt (approved):

t8m:
Merged. Thank you for the PR.

PR #342, opened 2021-03-29
pam_timestamp: replace hmac implementation

ikerexxe:
sha1 is no longer recommended as a cryptographic algorithm for
authentication. Thus, the idea of this change is to replace the
implementation provided by hmacsha1 included in pam_timestamp module by
the one in the openssl library. This way, there's no need to maintain
the cryptographic algorithm implementation and it can be easily changed
with a single configuration change.

modules/pam_timestamp/hmac_openssl_wrapper.c: implement wrapper
functions around openssl's hmac implementation. Moreover, manage the key
generation and its read and write in a file. Include an option to
configure the cryptographic algorithm in login.defs file.
modules/pam_timestamp/hmac_openssl_wrapper.h: likewise.
modules/pam_timestamp/pam_timestamp.c: replace calls to functions
provided by hmacsha1 by functions provided by openssl's wrapper.
configure.ac: include openssl dependecy if it is enabled.
modules/pam_timestamp/Makefile.am: include new files and openssl library
to compilation.
ci/install-dependencies.sh: include openssl library to dependencies.
NEWS: add new item to next release.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -504,6 +504,22 @@ else
 fi
 AC_SUBST([STRINGPARAM_VENDORDIR])
 
+AC_ARG_ENABLE([openssl],
+  AS_HELP_STRING([--enable-openssl],[use OpenSSL crypto libraries]),
+  [OPENSSL_ENABLED=$enableval], OPENSSL_ENABLED=no)
+if test "$OPENSSL_ENABLED" = "yes" ; then
+  AC_CHECK_LIB([crypto], [EVP_MAC_CTX_new],
+  [CRYPTO_LIBS="-lcrypto"
+   use_openssl=yes
+   AC_DEFINE([WITH_OPENSSL], 1, [OpenSSL provides crypto algorithm for hmac])
+   STRINGPARAM_HMAC="--stringparam profile.condition 'openssl_hmac'"],
+  [CRYPTO_LIBS=""
+   STRINGPARAM_HMAC="--stringparam profile.condition 'no_openssl_hmac'"])
+fi
+AC_SUBST([CRYPTO_LIBS])
+AC_SUBST([STRINGPARAM_HMAC])
+AM_CONDITIONAL([COND_USE_OPENSSL], [test "x$use_openssl" = "xyes"])
+
 dnl Checks for header files.
 AC_HEADER_DIRENT
 AC_HEADER_STDC
--- a/modules/pam_timestamp/Makefile.am
+++ b/modules/pam_timestamp/Makefile.am
@@ -18,12 +18,12 @@ TESTS = $(dist_check_SCRIPTS) $(check_PROGRAMS)
 securelibdir = $(SECUREDIR)
 secureconfdir = $(SCONFIGDIR)
 
-noinst_HEADERS = hmacsha1.h sha1.h
+noinst_HEADERS = hmacsha1.h sha1.h hmac_openssl_wrapper.h
 
 AM_CFLAGS = -I$(top_srcdir)/libpam/include -I$(top_srcdir)/libpamc/include \
 	$(WARN_CFLAGS)
 
-pam_timestamp_la_LDFLAGS = -no-undefined -avoid-version -module $(AM_LDFLAGS)
+pam_timestamp_la_LDFLAGS = -no-undefined -avoid-version -module $(AM_LDFLAGS) $(CRYPTO_LIBS)
 pam_timestamp_la_LIBADD = $(top_builddir)/libpam/libpam.la
 if HAVE_VERSIONING
   pam_timestamp_la_LDFLAGS += -Wl,--version-script=$(srcdir)/../modules.map
@@ -32,7 +32,12 @@ endif
 securelib_LTLIBRARIES = pam_timestamp.la
 sbin_PROGRAMS = pam_timestamp_check
 
-pam_timestamp_la_SOURCES = pam_timestamp.c hmacsha1.c sha1.c
+pam_timestamp_la_SOURCES = pam_timestamp.c
+if COND_USE_OPENSSL
+pam_timestamp_la_SOURCES += hmac_openssl_wrapper.c
+else
+pam_timestamp_la_SOURCES += hmacsha1.c sha1.c
+endif
 pam_timestamp_la_CFLAGS = $(AM_CFLAGS)
 
 pam_timestamp_check_SOURCES = pam_timestamp_check.c
@@ -40,7 +45,11 @@ pam_timestamp_check_CFLAGS = $(AM_CFLAGS) @EXE_CFLAGS@
 pam_timestamp_check_LDADD = $(top_builddir)/libpam/libpam.la
 pam_timestamp_check_LDFLAGS = @EXE_LDFLAGS@
 
+if COND_USE_OPENSSL
+hmacfile_SOURCES = hmac_openssl_wrapper.c
+else
 hmacfile_SOURCES = hmacfile.c hmacsha1.c sha1.c
+endif
 hmacfile_LDADD = $(top_builddir)/libpam/libpam.la
 
 check_PROGRAMS = hmacfile
--- a/modules/pam_timestamp/hmac_openssl_wrapper.c
+++ b/modules/pam_timestamp/hmac_openssl_wrapper.c
@@ -1,0 +1,381 @@
+/* Wrapper for hmac openssl implementation.
+ *
+ * Copyright (c) 2021 Red Hat, Inc.
+ * Written by Iker Pedrosa <ipedrosa@redhat.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, and the entire permission notice in its entirety,
+ *    including the disclaimer of warranties.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. The name of the author may not be used to endorse or promote
+ *    products derived from this software without specific prior
+ *    written permission.
+ *
+ * ALTERNATIVELY, this product may be distributed under the terms of
+ * the GNU Public License, in which case the provisions of the GPL are
+ * required INSTEAD OF the above restrictions.  (This clause is
+ * necessary due to a potential bad interaction between the GPL and
+ * the restrictions contained in a BSD-style copyright.)
+ *
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ * WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#include "config.h"
+
+#ifdef WITH_OPENSSL
+
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <syslog.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+#include <openssl/evp.h>
+#include <openssl/params.h>
+#include <openssl/core_names.h>
+
+#include <security/pam_ext.h>
+#include <security/pam_modutil.h>
+
+#include "hmac_openssl_wrapper.h"
+
+#define LOGIN_DEFS          "/etc/login.defs"
+#define CRYPTO_KEY          "HMAC_CRYPTO_ALGO"
+#define DEFAULT_ALGORITHM   "SHA512"
+#define MAX_HMAC_LENGTH     512
+#define MAX_KEY_LENGTH      EVP_MAX_KEY_LENGTH
+
+static char *
+get_crypto_algorithm(pam_handle_t *pamh, int debug){
+    char *config_value = NULL;
+
+    config_value = pam_modutil_search_key(pamh, LOGIN_DEFS, CRYPTO_KEY);
+
+    if (config_value == NULL) {
+        config_value = strdup(DEFAULT_ALGORITHM);
+        if (debug) {
+            pam_syslog(pamh, LOG_DEBUG,
+                   "Key [%s] not found, falling back to default algorithm [%s]\n",
+                   CRYPTO_KEY, DEFAULT_ALGORITHM);
+        }
+    }
+
+    return config_value;
+}
+
+static int
+generate_key(pam_handle_t *pamh, char **key, size_t key_size)
+{
+    int fd = 0;
+    size_t bytes_read = 0;
+    char * tmp = NULL;
+
+    fd = open("/dev/urandom", O_RDONLY);
+    if (fd == -1) {
+        pam_syslog(pamh, LOG_ERR, "Cannot open /dev/urandom: %m");
+        return PAM_AUTH_ERR;
+    }
+
+    tmp = malloc(key_size);
+    if (!tmp) {
+        pam_syslog(pamh, LOG_CRIT, "Not enough memory");
+        close(fd);
+        return PAM_AUTH_ERR;
+    }
+
+    bytes_read = pam_modutil_read(fd, tmp, key_size);
+    close(fd);
+
+    if (bytes_read < key_size) {
+        pam_syslog(pamh, LOG_ERR, "Short read on random device");
+        free(tmp);
+        return PAM_AUTH_ERR;
+    }
+
+    *key = tmp;
+
+    return PAM_SUCCESS;
+}
+
+static int
+read_file(pam_handle_t *pamh, int fd, char **text, size_t *text_length)
+{
+    struct stat st;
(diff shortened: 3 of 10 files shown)


modules/pam_timestamp/Makefile.am
```diff
@@ -32,15 +32,15 @@ endif
 securelib_LTLIBRARIES = pam_timestamp.la
 sbin_PROGRAMS = pam_timestamp_check
 
-pam_timestamp_la_SOURCES = pam_timestamp.c hmacsha1.c sha1.c
+pam_timestamp_la_SOURCES = pam_timestamp.c hmacsha1.c sha1.c hmac_openssl_wrapper.c
```
t8m, line 35: I do not think you want to simultaneously build and link both the wrapper and the original implementation. It should be made conditional based on whether use of OpenSSL is enabled or not. And similarly below.

ikerexxe: I've changed this part, please review it when possible.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,405 @@
...
+#include <errno.h>
+#include <openssl/hmac.h>
+
+#include <security/pam_ext.h>
+#include <security/pam_modutil.h>
+
+#include "hmac_openssl_wrapper.h"
+
+#define LOGIN_DEFS          "/etc/login.defs"
+#define CRYPTO_KEY          "HMAC_CRYPTO_ALGO"
```
t8m, line 57: I am not really sure all this hmac algorithm configurability is needed here. If SHA2 based algorithms are broken then you do not have replacement among those supported here anyway. And if you really want to have algorithm agility, instead of hardcoding a table of algorithm names, you should use something like EVP_getdigestbyname() function to get the EVP_MD.

Also did you agree with shadow utils maintainer on addition of the new key value in login.defs?

t8m (changes requested):

modules/pam_timestamp/hmac_openssl_wrapper.c
ikerexxe: > I am not really sure all this hmac algorithm configurability is needed here. If SHA2 based algorithms are broken then you do not have replacement among those supported here anyway. And if you really want to have algorithm agility, instead of hardcoding a table of algorithm names, you should use something like EVP_getdigestbyname() function to get the EVP_MD.

Awesome! I didn't know about this function. Thanks for pointing it out to me.

> Also did you agree with shadow utils maintainer on addition of the new key value in login.defs?

No, and this is something that I wanted to discuss but somehow I forgot about it. I doubt what's the best approach to this configuration. On the one hand there's the current approach. On the other hand it would be possible to create a `pam_timestamp.conf` file and include it there. Finally, it would be possible to reuse current `ENCRYPT_METHOD` in `login.defs`. What do you think is the best solution?

ikerexxe:
I have another doubt, this time concerning `hmac_size()` implementation. The current approach is agile from the development point of view as it leaves the actual details in the hands of openssl. The problem is that it takes more time to execute. On the other hand if I somehow hardcode the values then each time that an algorithm is changed the implementation would have to change. Do you think that the actual implementation is worth the execution overhead?

t8m:
I would not worry about the execution overhead.

configure.ac
```diff
@@ -642,6 +654,7 @@ AM_CONDITIONAL([COND_BUILD_PAM_SETQUOTA], [test "$ac_cv_func_quotactl" = yes])
 AM_CONDITIONAL([COND_BUILD_PAM_TTY_AUDIT], [test "$HAVE_AUDIT_TTY_STATUS" = yes])
 AM_CONDITIONAL([COND_BUILD_PAM_UNIX], [test "$enable_unix" = yes])
 AM_CONDITIONAL([COND_BUILD_PAM_USERDB], [test -n "$LIBDB"])
+AM_CONDITIONAL([COND_BUILD_OPENSSL], [test "x$use_openssl" = "xyes"])
```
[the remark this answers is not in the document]
ikerexxe: Fixed

modules/pam_timestamp/hmac_openssl_wrapper.c
ikerexxe: Can I get some feedback for this point? Every time I think about it I'm more willing to reuse `ENCRYPT_METHOD`as this  would mean changing only a single setting to improve the overall security of the system.

t8m: The problem with ENCRYPT_METHOD potentially might be that there might be other values possible for ENCRYPT_METHOD in future that are not valid hash algorithm names. So basically you'd have to have some sensible default and if the ENCRYPT_METHOD fails to return valid hash algorithm then you'd have to fallback to the default.

ikerexxe: Thank you for the analysis. Then I think I'll continue with the current approach, so I'll contact shadow-utils maintainers to get their opinion.

ikerexxe: The key is available in shadow-utils master branch as https://github.com/shadow-maint/shadow/pull/333 got merged.

ikerexxe:
Hi @t8m,

Can I have a second round of reviews for this PR?

t8m:
> Can I have a second round of reviews for this PR?

I am sorry but I am unfortunately very busy with other things currently. I'll try to look at this in a week or so.

t8m (changes requested):

ikerexxe:
Thanks for the feedback @t8m, I wasn't aware that there was a new API that provided the HMAC functionality and that the old one was being deprecated. I've updated all the calls to use only the new API, and removed the old one as from my point of view it doesn't make any sense to add new code that will be removed soon.

By the way, I've created a COPR repository to test the change: https://copr.fedorainfracloud.org/coprs/ipedrosa/openssl_pam_timestamp/

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+
+    return PAM_SUCCESS;
+}
+
+static int
+hmac_management(pam_handle_t *pamh, void **out, size_t *out_length,
+                char *key, size_t key_length,
+                OSSL_PARAM subalg_param[], const void *text, size_t text_length)
+{
+    int ret = 0;
```
t8m, line 250: Please use `ret = PAM_AUTH_ERR` here and remove all these assignments in the error cases.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+    EVP_MAC *evp_mac = NULL;
+    EVP_MAC_CTX *ctx = NULL;
+    unsigned char* hmac_message = NULL;
```
t8m, line 253: Nit: `unsigned char *hmac_message...`

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+    if (evp_mac != NULL) {
+        EVP_MAC_free(evp_mac);
+    }
+    return ret;
+}
+
+int
+hmac_size(pam_handle_t *pamh, size_t *hmac_length)
+{
+    int ret = 0;
```
t8m, line 329: As above.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+
+static void
+get_crypto_algorithm(pam_handle_t *pamh, OSSL_PARAM *subalg_param){
+    char *config_value = NULL;
+
+    config_value = pam_modutil_search_key(pamh, LOGIN_DEFS, CRYPTO_KEY);
+
+    if (config_value != NULL) {
+        subalg_param[0] = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST,
+                                                           (char *)config_value,
```
t8m, line 72: The config_value will be leaked as it is allocated and you're not freeing it after the subalg_param is no longer used. I would probably not try to construct the params in get_crypto_algorithm() but just return the value from here and call the construct (and later free) where the algorithm string is used. For the DEFAULT_ALGORITHM I'd just return strdup().

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+                                                           0);
+    } else {
+        subalg_param[0] = OSSL_PARAM_construct_utf8_string(OSSL_MAC_PARAM_DIGEST,
+                                                           (char *)DEFAULT_ALGORITHM,
+                                                           0);
+        pam_syslog(pamh, LOG_WARNING,
```
t8m, line 78: This should be LOG_DEBUG at most and it should be called only if the debug parameter of the PAM module is set.

t8m (changes requested):

modules/pam_timestamp/pam_timestamp.c
```diff
@@ -449,14 +456,22 @@ pam_sm_authenticate(pam_handle_t *pamh, int flags, int argc, const char **argv)
 			return PAM_AUTH_ERR;
 		}
 
+#ifdef WITH_OPENSSL
+		if (hmac_size(pamh, debug, &maclen)) {
+			pam_syslog(pamh, LOG_CRIT, "Unable to get hmac size");
```
t8m, line 461: I would drop this log as there will be already a log message from the hmac_size() call. Too much logging is not good.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+    ret = EVP_MAC_final(ctx, hmac_message, &hmac_length, MAX_HMAC_LENGTH);
+    if (ret == 0) {
+        pam_syslog(pamh, LOG_ERR, "Unable to calculate hmac message");
+        goto done;
+    }
+
+    *out_length = hmac_length;
+    *out = malloc(*out_length);
+    if (*out == NULL) {
+        pam_syslog(pamh, LOG_ERR, "Not enough memory");
```
t8m, line 299: LOG_CRIT here.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+
+    ret = EVP_MAC_update(ctx, (const unsigned char *)text, text_length);
+    if (ret == 0) {
+        pam_syslog(pamh, LOG_ERR, "Unable to update hmac context");
+        goto done;
+    }
+
+    hmac_message = (unsigned char*)malloc(sizeof(unsigned char) * MAX_HMAC_LENGTH);
+    if (!hmac_message) {
+        pam_syslog(pamh, LOG_ERR, "Not enough memory");
```
t8m, line 286: LOG_CRIT here.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+
+    if (fstat(fd, &st) == -1) {
+        pam_syslog(pamh, LOG_ERR, "Unable to stat file: %m");
+        close(fd);
+        return PAM_AUTH_ERR;
+    }
+
+    tmp = malloc(st.st_size);
+    if (!tmp) {
+        pam_syslog(pamh, LOG_ERR, "Not enough memory");
```
t8m, line 142: LOG_CRIT here

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+
+    fd = open("/dev/urandom", O_RDONLY);
+    if (fd == -1) {
+        pam_syslog(pamh, LOG_ERR, "Cannot open /dev/urandom: %m");
+        return PAM_AUTH_ERR;
+    }
+
+    tmp = malloc(key_size);
+    if (!tmp) {
+        pam_syslog(pamh, LOG_ERR, "Not enough memory");
```
t8m, line 98: LOG_CRIT here

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,395 @@
...
+            break;
+        }
+        count += i;
+    }
+
+    close(fd);
+
+    if (count < key_size) {
+        pam_syslog(pamh, LOG_ERR, "Short read on random device");
+        memset(tmp, 0, key_size);
```
t8m, line 116: This is probably useless as these are just random bytes.

modules/pam_timestamp/pam_timestamp.c
```diff
@@ -487,8 +502,18 @@ pam_sm_authenticate(pam_handle_t *pamh, int flags, int argc, const char **argv)
 		message_end = message + strlen(path) + 1 + sizeof(then);
 
 		/* Regenerate the MAC. */
-		hmac_sha1_generate_file(pamh, &mac, &maclen, TIMESTAMPKEY, 0, 0,
-					message, message_end - message);
+#ifdef WITH_OPENSSL
+		if (hmac_generate(pamh, debug, &mac, &maclen, TIMESTAMPKEY,
+				  ROOT_USER, ROOT_GROUP, message, message_end - message)) {
+			pam_syslog(pamh, LOG_ERR, "Error generating hmac");
```
t8m, line 508: There should be already a syslog message from hmac_generate, shouldn't it? If so I'd drop this to not spam the syslog too much.

modules/pam_timestamp/pam_timestamp.c
```diff
@@ -605,8 +630,17 @@ pam_sm_open_session(pam_handle_t *pamh, int flags UNUSED, int argc, const char *
 		}
 	}
 
+#ifdef WITH_OPENSSL
+	if (hmac_size(pamh, debug, &maclen)) {
+		pam_syslog(pamh, LOG_CRIT, "Unable to get hmac size");
```
t8m, line 635: As above.

modules/pam_timestamp/pam_timestamp.c
```diff
@@ -621,15 +655,22 @@ pam_sm_open_session(pam_handle_t *pamh, int flags UNUSED, int argc, const char *
...
 
 	/* Generate the MAC and append it to the plaintext. */
-	hmac_sha1_generate_file(pamh, &mac, &maclen,
-				TIMESTAMPKEY,
-				0, 0,
-				text, p - text);
+#ifdef WITH_OPENSSL
+	if (hmac_generate(pamh, debug, &mac, &maclen, TIMESTAMPKEY,
+			  ROOT_USER, ROOT_GROUP, text, p - text)) {
+		pam_syslog(pamh, LOG_ERR, "Error calculating hmac");
```
t8m, line 661: As above.

t8m (changes requested):
A few final nits. @ldv-alt Could you please also look?

t8m (approved):

t8m:
Approved. @ldv-alt OK to merge this?

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    count = 0;
+    while (count < key_size) {
+        i = read(fd, tmp + count, key_size - count);
+        if ((i == 0) || (i == -1)) {
+            break;
+        }
+        count += i;
+    }
```
ldv-alt, line 110: Consider using pam_modutil_read.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,394 @@
...
+
+    count = 0;
+    while (count < (size_t)st.st_size) {
+        i = read(fd, tmp + count, st.st_size - count);
+        if ((i == 0) || (i == -1)) {
+            break;
+        }
+        count += i;
+    }
```
ldv-alt, line 153: Consider using pam_modutil_read.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,394 @@
...
+
+    count = 0;
+    while (count < text_length) {
+        i = write(fd, text + count, text_length - count);
+        if ((i == 0) || (i == -1)) {
+            break;
+        }
+        count += i;
+    }
```
ldv-alt, line 201: Consider using pam_modutil_write.

ldv-alt (commented):
Consider using functions from pam_modutil_ioloop.c

ikerexxe:
> Consider using functions from pam_modutil_ioloop.c

Done. Thank you for your feedback.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,374 @@
...
+        free(text);
+        close(fd);
+        return PAM_AUTH_ERR;
+    }
+
+    bytes_written = pam_modutil_write(fd, text, text_length);
+    close(fd);
+
+    if (bytes_written < text_length) {
+        pam_syslog(pamh, LOG_ERR, "Short write on random device");
```
ldv-alt, line 180: Is `file_name` really a random device?

ikerexxe: No, not really. I changed it for the `file_name`.

t8m (approved):

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    bytes_read = pam_modutil_read(fd, tmp, st.st_size);
+    close(fd);
+
+    if (bytes_read < (size_t)st.st_size) {
+        memset(tmp, 0, st.st_size);
+        free(tmp);
+        return PAM_AUTH_ERR;
+    }
```
ldv-alt, line 143: Should some diagnostics be printed in case of short read, too?

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -0,0 +1,374 @@
...
+    char *tmp = NULL;
+
+    if (fstat(fd, &st) == -1) {
+        pam_syslog(pamh, LOG_ERR, "Unable to stat file: %m");
+        close(fd);
+        return PAM_AUTH_ERR;
+    }
+
+    tmp = malloc(st.st_size);
+    if (!tmp) {
```
t8m, line 130: This might produce misleading syslog if st.st_size is 0 actually. Please check that.

modules/pam_timestamp/hmac_openssl_wrapper.c
ikerexxe: Yes, I've just added them.

modules/pam_timestamp/hmac_openssl_wrapper.c
ikerexxe: I've added a check in the previous lines. Please review it.

t8m (approved):

ldv-alt (approved):
OK, I suppose this could be merged now.

t8m:

PR #226, opened 2020-05-12
pam_access: allow symbolic hostnames in access.conf file

jmoellers:
When checking the access.conf file, allow numeric IP addresses and netmasks as well as symbolic names and netmasks.

PR #343, opened 2021-04-01
pam_access: handle the 'LOCAL' keyword more securely

jo-he:
Previously, for remote connections (e.g. SSH) it slipped through to 'network_netmask_match', where it was rejected just luckily, because it does does not mach a netmask pattern or an IP address.

This can become a security issue, when Linux vendors extend 'network_netmask_match', e.g. with hostname resolution (like SUSE did in SLES 15). So better prefer the code to be locally correct.

### DIFF
--- a/modules/pam_access/pam_access.c
+++ b/modules/pam_access/pam_access.c
@@ -623,6 +623,8 @@ from_match (pam_handle_t *pamh UNUSED, char *tok, struct login_info *item)
     } else if (item->from_remote_host == 0) {	/* local: no PAM_RHOSTS */
 	if (strcasecmp(tok, "LOCAL") == 0)
 	    return (YES);
+    } else if (strcasecmp(tok, "LOCAL") == 0) {
+        return NO;
     } else if (tok[(tok_len = strlen(tok)) - 1] == '.') {
       struct addrinfo hint;
 


jo-he:
Pull request #226 is the SUSE patch I was referring to. Without fixing the `LOCAL` check, it would fill the syslog with `cannot resolve hostname "LOCAL"` messages.

modules/pam_access/pam_access.c
```diff
@@ -623,6 +623,8 @@ from_match (pam_handle_t *pamh UNUSED, char *tok, struct login_info *item)
     } else if (item->from_remote_host == 0) {	/* local: no PAM_RHOSTS */
 	if (strcasecmp(tok, "LOCAL") == 0)
 	    return (YES);
+    } else if (strcasecmp(tok, "LOCAL") == 0) {
+        return NO;
```
t8m, line 627: This can be rewritten in a more logical way:
```
} else if (strcasecmp(tok, "LOCAL") == 0) {
   if (item->from_remote_host == 0)
       return YES;
   return NO;
} else if ....
```

jo-he: That is not the same logic. With your suggestion, local accesses would fall through, when `LOCAL` is not configured.

jmoellers: IMHO while this is not the same logic, it has the same outcome:
from_match() only returns YES if
item->from_remote_host == 0 AND strcasecmp(tok, "LOCAL") == 0
for both patches.
One could even leave out the "return NO;" because that's the default at the end of the function.

t8m: You're right. But I still prefer doing it differently so that the comparison of tok with LOCAL is not done twice. My alternate PR: #345

jo-he: Okay, that clean-up makes it more readable. I just added some minor comments.

t8m:
Closing as alternative PR was merged. But thank you for the heads up!

PR #345, opened 2021-04-06
pam_access: clean up the remote host matching code

t8m:
* modules/pam_access/pam_access.c (from_match): Split out remote_match()
  function and avoid calling it when matching against LOCAL keyword.

### DIFF
--- a/modules/pam_access/pam_access.c
+++ b/modules/pam_access/pam_access.c
@@ -160,6 +160,7 @@ static int list_match (pam_handle_t *, char *, char *, struct login_info *,
 static int user_match (pam_handle_t *, char *, struct login_info *);
 static int group_match (pam_handle_t *, const char *, const char *, int);
 static int from_match (pam_handle_t *, char *, struct login_info *);
+static int remote_match (pam_handle_t *, char *, struct login_info *);
 static int string_match (pam_handle_t *, const char *, const char *, int);
 static int network_netmask_match (pam_handle_t *, const char *, const char *, struct login_info *);
 
@@ -589,11 +590,9 @@ group_match (pam_handle_t *pamh, const char *tok, const char* usr,
 /* from_match - match a host or tty against a list of tokens */
 
 static int
-from_match (pam_handle_t *pamh UNUSED, char *tok, struct login_info *item)
+from_match (pam_handle_t *pamh, char *tok, struct login_info *item)
 {
     const char *string = item->from;
-    int        tok_len;
-    int        str_len;
     int        rv;
 
     if (item->debug)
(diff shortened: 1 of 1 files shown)


modules/pam_access/pam_access.c
```diff
@@ -620,10 +621,24 @@ from_match (pam_handle_t *pamh UNUSED, char *tok, struct login_info *item)
...
 	    return (YES);
-    } else if (item->from_remote_host == 0) {	/* local: no PAM_RHOSTS */
-	if (strcasecmp(tok, "LOCAL") == 0)
-	    return (YES);
-    } else if (tok[(tok_len = strlen(tok)) - 1] == '.') {
+    } else if (strcasecmp(tok, "LOCAL") == 0) {
+	    /* LOCAL matches only local accesses */
+	    if (item->from_remote_host == 0)
+	        return YES;
+	    return NO;
```
jo-he, line 628: This `return` is duplicate to the default below.

modules/pam_access/pam_access.c
```diff
@@ -620,10 +621,24 @@ from_match (pam_handle_t *pamh UNUSED, char *tok, struct login_info *item)
 	if ((str_len = strlen(string)) > (tok_len = strlen(tok))
 	    && strcasecmp(tok, string + str_len - tok_len) == 0)
 	    return (YES);
-    } else if (item->from_remote_host == 0) {	/* local: no PAM_RHOSTS */
-	if (strcasecmp(tok, "LOCAL") == 0)
-	    return (YES);
-    } else if (tok[(tok_len = strlen(tok)) - 1] == '.') {
+    } else if (strcasecmp(tok, "LOCAL") == 0) {
+	    /* LOCAL matches only local accesses */
+	    if (item->from_remote_host == 0)
```
jo-he, line 626: Maybe use `if (!item->from_remote_host)` to match the boolean nature of the next else branch.

modules/pam_access/pam_access.c
t8m: I prefer to keep it.

t8m:
I did one more change - moved the domain match to remote match as it makes no sense to do it against TTY or SERVICE.

t8m:
@ldv-alt Please review

modules/pam_access/pam_access.c
jo-he: Okay, but in that case
```
return item->from_remote_host ? NO : YES;
```
or at least
```
if (item->from_remote_host)
    return NO;
return YES;
```
would be a bit more readable.  ;-)

jo-he:
closes #343

modules/pam_access/pam_access.c
```diff
@@ -616,13 +615,28 @@ from_match (pam_handle_t *pamh UNUSED, char *tok, struct login_info *item)
...
+{
+    const char *string = item->from;
+    size_t tok_len = strlen(tok);
+    size_t str_len;
+
+    if (tok[0] == '.') {			/* domain: match last fields */
+      if ((str_len = strlen(string)) > tok_len
+	  && strcasecmp(tok, string + str_len - tok_len) == 0)
+	return YES;
     } else if (tok[(tok_len = strlen(tok)) - 1] == '.') {
```
ldv-alt, line 640: `tok_len` is being assigned twice in this function.

ldv-alt (changes requested):

modules/pam_access/pam_access.c
t8m: Fixed

ldv-alt (approved):

PR #348, opened 2021-04-12

jmoellers:
Specifying "unlimited" (or any of its synonyms) for "nofile" in the limits.conf file will prevent users to log in as `RLIM_INFINITY` is not a valid value for `RLIMIT_NOFILE` (see `man setrlimit`). A better way would be to use the value of `/proc/sys/fs/nr_open` in this case.


ldv-alt:
Note that this PR contains unrelated changes to pam_setquota, please consider rebasing.

ldv-alt:
duplicate of #353

PR #351, opened 2021-04-13

jmoellers:
Specifying "unlimited" (or any of its synonyms) for "nofile" in the limits.conf file will prevent users to log in as `RLIM_INFINITY` is not a valid value for `RLIMIT_NOFILE` (see `man setrlimit`). A better way would be to use the value of `/proc/sys/fs/nr_open` in this case.


t8m:
Could you please avoid these whitespace changes and Merge commits?

jmoellers:
Hm, this is strange: I actually made a fresh fork and then a fresh clone shortly before making the change.

ldv-alt:
duplicate of #353

PR #353, opened 2021-04-14

jmoellers:
pam_limits: Specifying "unlimited" (or any of its synonyms) for "nofile" in the limits.conf file will prevent users to log in as `RLIM_INFINITY` is not a valid value for `RLIMIT_NOFILE` as this exceeds the maximum defined by /proc/sys/fs/nr_open (see `man setrlimit`).
Therefore, if this is given, the contents of /proc/sys/fs/nr_open is used instead as the new limit. If this cannot be read, `RLIM_INFINITY` is used because this is what the user configured.

### DIFF
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -714,7 +714,7 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
 		pam_syslog(pamh, LOG_WARNING,
 			   "Cannot set \"nofile\" to a sensible value");
 	    else if (ctrl & PAM_DEBUG_ARG)
-		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %lu", (long unsigned) rlimit_value);
+		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %llu", (unsigned long long) rlimit_value);
 	}
 	break;
     }
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -507,14 +507,18 @@ value_from_proc_sys_fs(const char *name, rlim_t *valuep)
     if ((fp = fopen(pathname, "r")) != NULL) {
 	if (fgets(buf, sizeof(buf), fp) != NULL) {
 	    char *endptr;
+	    rlim_t value;
 
 #ifdef __USE_FILE_OFFSET64
-	    *valuep = strtoull(buf, &endptr, 10);
+	    value = strtoull(buf, &endptr, 10);
 #else
-	    *valuep = strtoul(buf, &endptr, 10);
+	    value = strtoul(buf, &endptr, 10);
 #endif
 
-	    retval = (endptr != buf);
+	    if (endptr != buf) {
+		*valuep = value;
+		retval = 1;
+	    }
 	}
 
 	fclose(fp);
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -510,7 +510,7 @@ value_from_file(const char *pathname, rlim_t *valuep)
 	    value = strtoull(buf, &endptr, 10);
 	    if (endptr != buf &&
 		(value != ULLONG_MAX || errno == 0) &&
-                (unsigned long long) (rlim_t) value == value)) {
+                (unsigned long long) (rlim_t) value == value) {
 		*valuep = (rlim_t) value;
 		retval = 1;
 	    }
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -711,7 +711,8 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
 		pam_syslog(pamh, LOG_WARNING,
 			   "Cannot set \"nofile\" to a sensible value");
 	    else if (ctrl & PAM_DEBUG_ARG)
-		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %llu", (unsigned long long) rlimit_value);
+		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %llu",
+			   (unsigned long long) rlimit_value);
 	}
 	break;
     }
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -507,16 +507,14 @@ value_from_proc_sys_fs(const char *name, rlim_t *valuep)
     if ((fp = fopen(pathname, "r")) != NULL) {
 	if (fgets(buf, sizeof(buf), fp) != NULL) {
 	    char *endptr;
-	    rlim_t value;
+	    unsigned long long value;
 
-#ifdef __USE_FILE_OFFSET64
+	    errno = 0;
 	    value = strtoull(buf, &endptr, 10);
-#else
-	    value = strtoul(buf, &endptr, 10);
-#endif
-
-	    if (endptr != buf) {
-		*valuep = value;
+	    if (endptr != buf &&
+		(value != ULLONG_MAX || errno == 0) &&
+                (unsigned long long) (rlim_t) value == value)) {
+		*valuep = (rlim_t) value;
 		retval = 1;
 	    }
 	}
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -488,21 +488,18 @@ static int init_limits(pam_handle_t *pamh, struct pam_limit_s *pl, int ctrl)
 }
 
 /*
- * Read the contents of /proc/sys/fs/<name>
+ * Read the contents of <pathname> and return it in *valuep
  * return 1 if conversion succeeds, result is in *valuep
- * return 0 if conversion fails.
+ * return 0 if conversion fails, *valuep is untouched.
  */
 static int
-value_from_proc_sys_fs(const char *name, rlim_t *valuep)
+value_from_file(const char *pathname, rlim_t *valuep)
 {
-    char pathname[128];
     char buf[128];
     FILE *fp;
     int retval;
 
     retval = 0;
-
-    snprintf(pathname, sizeof(pathname), "/proc/sys/fs/%s", name);
 
     if ((fp = fopen(pathname, "r")) != NULL) {
 	if (fgets(buf, sizeof(buf), fp) != NULL) {
@@ -710,7 +707,7 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
 	 * the value in /proc/sys/fs/nr_open instead.
 	 */
 	if (rlimit_value == RLIM_INFINITY) {
-	    if (!value_from_proc_sys_fs("nr_open", &rlimit_value))
+	    if (!value_from_file("/proc/sys/fs/nr_open", &rlimit_value))
 		pam_syslog(pamh, LOG_WARNING,
 			   "Cannot set \"nofile\" to a sensible value");
 	    else if (ctrl & PAM_DEBUG_ARG)
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -711,7 +711,7 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
 	    if (!value_from_proc_sys_fs("nr_open", &rlimit_value))
 		pam_syslog(pamh, LOG_WARNING,
 			   "Cannot set \"nofile\" to a sensible value");
-	    else
+	    else if (ctrl & PAM_DEBUG_ARG)
 		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %lu", (long unsigned) rlimit_value);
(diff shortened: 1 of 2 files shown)


modules/pam_limits/pam_limits.c
```diff
@@ -666,6 +702,19 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
...
+	case RLIMIT_NOFILE:
+	/*
+	 * If nofile is to be set to "unlimited", try to set it to
+	 * the value in /proc/sys/fs/nr_open instead.
+	 */
+	if (rlimit_value == RLIM_INFINITY) {
+	    if (!value_from_proc_sys_fs("nr_open", &rlimit_value))
+		pam_syslog(pamh, LOG_WARNING,
+			   "Cannot set \"nofile\" to a sensible value");
+	    else
```
t8m, line 714: Please add `if (ctrl & PAM_DEBUG_ARG)`

t8m (approved):

modules/pam_limits/pam_limits.c
```diff
@@ -487,6 +487,42 @@ static int init_limits(pam_handle_t *pamh, struct pam_limit_s *pl, int ctrl)
...
+	if (fgets(buf, sizeof(buf), fp) != NULL) {
+	    char *endptr;
+
+#ifdef __USE_FILE_OFFSET64
+	    *valuep = strtoull(buf, &endptr, 10);
+#else
+	    *valuep = strtoul(buf, &endptr, 10);
+#endif
+
+	    retval = (endptr != buf);
```
ldv-alt, line 517: Here `*valuep` is assigned regardless of `retval`, is it expected?

modules/pam_limits/pam_limits.c
```diff
@@ -666,6 +702,19 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
...
+	 * If nofile is to be set to "unlimited", try to set it to
+	 * the value in /proc/sys/fs/nr_open instead.
+	 */
+	if (rlimit_value == RLIM_INFINITY) {
+	    if (!value_from_proc_sys_fs("nr_open", &rlimit_value))
+		pam_syslog(pamh, LOG_WARNING,
+			   "Cannot set \"nofile\" to a sensible value");
+	    else if (ctrl & PAM_DEBUG_ARG)
+		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %lu", (long unsigned) rlimit_value);
+	}
```
ldv-alt, line 716: Starting from this point, the value assigned to `rlimit_value` by `value_from_proc_sys_fs` is used regardless of the return value of the latter, is it expected?

ldv-alt (changes requested):

jmoellers:
On 20.04.21 03:54, Dmitry V. Levin wrote:
> ***@***.**** requested changes on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In modules/pam_limits/pam_limits.c
> <https://github.com/linux-pam/linux-pam/pull/353#discussion_r616286715>:
> 
>> +#ifdef __USE_FILE_OFFSET64
> +	    *valuep = strtoull(buf, &endptr, 10);
> +#else
> +	    *valuep = strtoul(buf, &endptr, 10);
> +#endif
> +
> +	    retval = (endptr != buf);
> 
> Here |*valuep| is assigned regardless of |retval|, is it expected?
> 
> ------------------------------------------------------------------------
> 
> In modules/pam_limits/pam_limits.c
> <https://github.com/linux-pam/linux-pam/pull/353#discussion_r616287999>:
> 
>> +	if (rlimit_value == RLIM_INFINITY) {
> +	    if (!value_from_proc_sys_fs("nr_open", &rlimit_value))
> +		pam_syslog(pamh, LOG_WARNING,
> +			   "Cannot set \"nofile\" to a sensible value");
> +	    else if (ctrl & PAM_DEBUG_ARG)
> +		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %lu", (long unsigned) rlimit_value);
> +	}
> 
> Starting from this point, the value assigned to |rlimit_value| by
> |value_from_proc_sys_fs| is used regardless of the return value of the
> latter, is it expected?

No, you're absolutely right: *valuep/rlimit_value should only be
modified if a valid number was found in the buffer.
I have changed the code in my fork. Do I need to create a new pull
request from there?

Josef
-- 
SUSE Software Solutions Germany GmbH
Maxfeldstr. 5
90409 Nürnberg
Germany

(HRB 36809, AG Nürnberg)
Geschäftsführer: Felix Imendörffer

t8m:
You can just `git commit -a --amend` and force push. Or you can do `git commit -a --fixup HEAD`.

jmoellers:
On 20.04.21 09:40, Tomáš Mráz wrote:
> You can just |git commit -a --amend| and force push. Or you can do |git
> commit -a --fixup HEAD|.

Hm, I made the change directly in my fork on GH, so I'm not sure how I
can get this to work on my local clone.


-- 
SUSE Software Solutions Germany GmbH
Maxfeldstr. 5
90409 Nürnberg
Germany

(HRB 36809, AG Nürnberg)
Geschäftsführer: Felix Imendörffer

t8m:
Ah, OK. It's fine then. No need for another PR as the changes in the branch in the fork are automatically included in the PR.

modules/pam_limits/pam_limits.c
t8m: Maybe it would make some sense to default to some reasonably large value? So failure of value_from_proc_sys_fs does not prevent login?

jmoellers: The question is: which one?
No matter which one you choose, there will always be cases where the value which **should** be in `/proc/sys/fs/nr_open` is smaller. So I would really go for: "if we have no other value, take what the administrator has configured".

t8m: But that value is RLIM_INFINITY which is unusable and will fail the setrlimit() call. And that effectively makes using infinity for RLIMIT_NOFILE unusable because you cannot depend on it being converted to reasonable value.

On the other hand value_from_proc_sys_fs() should not normally fail, so perhaps we could right away make its failure to mean failing the pam_limit pam_open_session().

jmoellers: How about not change the value at all?
If the value is RLIM_INFINITY and value_from_proc_sys_fs() fails, set rlimit_value to pl->limits[limit_item].limit.rlim_cur or pl->limits[limit_item].limit.rlim_max, depending upon limit_type?

t8m: Hmm... that would make some sense. @ldv-alt what's your opinion?

ldv-alt: If the contents of `/proc/sys/fs/nr_open` cannot be obtained, there is no easy way to find out the maximal value for `RLIMIT_NOFILE`. It could be probed, but the range supported by Linux kernel is too large to probe. How likely is the chance of failing to obtain the contents of `/proc/sys/fs/nr_open`?
- `/proc` is not mounted in the mount namespace: not likely, but possible;
- `/proc/sys/fs/nr_open` is available but fopen or fgets failed: unlikely, other serious problems are expected;
- `strtoul` failed: very unlikely, other serious problems are expected.

There are indeed two possible courses of action:
- strict: use the value of `RLIM_INFINITY` which is always larger than the maximal value for `RLIMIT_NOFILE`, causing the module to fail with `PAM_PERM_DENIED` as before the proposed change;
- permissive: avoid changing the corresponding limit as if it was not configured.

I'm slightly in favor of the first variant as it seems to follow the principle of the least surprise: signal an error if the maximal limit is configured but the software failed to calculate the corresponding value.

modules/pam_limits/pam_limits.c
```diff
@@ -487,6 +487,46 @@ static int init_limits(pam_handle_t *pamh, struct pam_limit_s *pl, int ctrl)
...
+value_from_proc_sys_fs(const char *name, rlim_t *valuep)
+{
+    char pathname[128];
+    char buf[128];
+    FILE *fp;
+    int retval;
+
+    retval = 0;
+
+    snprintf(pathname, sizeof(pathname), "/proc/sys/fs/%s", name);
```
ldv-alt, line 505: I don't see anything specific to /proc/sys/fs/ here.
What if you pass the whole pathname to this function?

modules/pam_limits/pam_limits.c
```diff
@@ -487,6 +487,46 @@ static int init_limits(pam_handle_t *pamh, struct pam_limit_s *pl, int ctrl)
...
+    if ((fp = fopen(pathname, "r")) != NULL) {
+	if (fgets(buf, sizeof(buf), fp) != NULL) {
+	    char *endptr;
+	    rlim_t value;
+
+#ifdef __USE_FILE_OFFSET64
+	    value = strtoull(buf, &endptr, 10);
+#else
+	    value = strtoul(buf, &endptr, 10);
+#endif
```
ldv-alt, line 516: Yes, configure.ac uses `AC_SYS_LARGEFILE` macro, but do we really need this very glibc-specific ifdef?
We could use `strtoull`unconditionally, assign the return value to a variable of type unsigned long long, and check for an overflow.  FWiW, we shall check the return value + `errno` anyway.  It's very unlikely given that `sysctl_nr_open` is less than `INT_MAX`, but a correct check is simple enough, e.g.:
```
char *endptr;
unsigned long long value;
errno = 0;
value = strtoull(buf, &endptr, 10);
if (endptr != buf &&
    (value != ULLONG_MAX || errno == 0) &&
    (unsigned long long) (rlim_t) value == value) {
        *valuep = (rlim_t) value;
        retval = 1;
}
```

modules/pam_limits/pam_limits.c
```diff
@@ -666,6 +706,19 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
...
+	/*
+	 * If nofile is to be set to "unlimited", try to set it to
+	 * the value in /proc/sys/fs/nr_open instead.
+	 */
+	if (rlimit_value == RLIM_INFINITY) {
+	    if (!value_from_proc_sys_fs("nr_open", &rlimit_value))
+		pam_syslog(pamh, LOG_WARNING,
+			   "Cannot set \"nofile\" to a sensible value");
+	    else if (ctrl & PAM_DEBUG_ARG)
+		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %lu", (long unsigned) rlimit_value);
```
ldv-alt, line 719: Despite the fact that `nr_open` is less than `INT_MAX`, let's remember that `rlim_t` is most likely 64-bit and print it as `unsigned long long`.

ldv-alt (changes requested):

modules/pam_limits/pam_limits.c
jmoellers: No problem here. The current construct is a relic from an intermediate idea to allow filenames as "values", ie
```
* - nofile nr_open
```

modules/pam_limits/pam_limits.c
jmoellers: The same construct is used inside `process_limit()`, so I stuck with that.
But, again: no problem here.

modules/pam_limits/pam_limits.c
jmoellers: np.

modules/pam_limits/pam_limits.c
ldv-alt: Thanks.
Looks like the module would benefit from some cleaning up, but that's out of scope of this PR.

modules/pam_limits/pam_limits.c
```diff
@@ -666,6 +701,19 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
...
+	/*
+	 * If nofile is to be set to "unlimited", try to set it to
+	 * the value in /proc/sys/fs/nr_open instead.
+	 */
+	if (rlimit_value == RLIM_INFINITY) {
+	    if (!value_from_file("/proc/sys/fs/nr_open", &rlimit_value))
+		pam_syslog(pamh, LOG_WARNING,
+			   "Cannot set \"nofile\" to a sensible value");
+	    else if (ctrl & PAM_DEBUG_ARG)
+		pam_syslog(pamh, LOG_DEBUG, "Setting \"nofile\" limit to %llu", (unsigned long long) rlimit_value);
```
ldv-alt, line 714: This line is getting too long, please wrap.

ldv-alt (changes requested):
Overall looks good, but the commit message is lacking the information provided in the PR.

Could you change the commit message so its first line would start with `pam_limits: ` prefix,
and the whole message would be as descriptive as the PR, please?

t8m:
We will need to merge this manually anyway as the commits need squashing and squashing via github mangles author.
So we can adjust the commit message during the squash.

t8m (approved):

ldv-alt (approved):

ldv-alt:
> We will need to merge this manually anyway as the commits need squashing and squashing via github mangles author.
> So we can adjust the commit message during the squash.

This doesn't necessarily have to be done by us, I suppose Josef can do a rebase+squash, too.

t8m:
Merged manually to master with squasing and commit message adjustment.

PR #355, opened 2021-04-16
pam_userdb: Prevent garbage characters from db

ep69:
Recently, I debugged some problems stemming from the fact that `db_fetch` output in `.dptr` contained a non-printable character in the end and got passed directly to `crypt_r`. I tested this patch in RHEL-8, but I have no prior experience with pam. I'd be glad for a review.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1791965

### DIFF
--- a/modules/pam_userdb/pam_userdb.c
+++ b/modules/pam_userdb/pam_userdb.c
@@ -194,7 +194,7 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
     }
 
     if (data.dptr != NULL) {
-	int compare = 0;
+	int compare = -2;
 
 	if (ctrl & PAM_KEY_ONLY_ARG)
 	  {
@@ -209,36 +209,48 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
 	  char *cryptpw = NULL;
 
 	  if (data.dsize < 13) {
-	    compare = -2;
+	    /* hash is too short */
+	    pam_syslog(pamh, LOG_INFO, "password hash in database is too short");
 	  } else if (ctrl & PAM_ICASE_ARG) {
-	    compare = -2;
+	    pam_syslog(pamh, LOG_INFO,
+	       "case-insensitive comparison only works with plaintext passwords");
 	  } else {
-#ifdef HAVE_CRYPT_R
-	    struct crypt_data *cdata = NULL;
-	    cdata = malloc(sizeof(*cdata));
-	    if (cdata != NULL) {
-		cdata->initialized = 0;
-		cryptpw = crypt_r(pass, data.dptr, cdata);
-	    }
-#else
-	    cryptpw = crypt (pass, data.dptr);
-#endif
-	    if (cryptpw && strlen(cryptpw) == (size_t)data.dsize) {
-	      compare = memcmp(data.dptr, cryptpw, data.dsize);
-	    } else {
-	      compare = -2;
-	      if (ctrl & PAM_DEBUG_ARG) {
-		if (cryptpw)
-		  pam_syslog(pamh, LOG_INFO, "lengths of computed and stored hashes differ");
-		else
-		  pam_syslog(pamh, LOG_INFO, "crypt() returned NULL");
-	      }
-	    }
-#ifdef HAVE_CRYPT_R
-	    free(cdata);
-#endif
-	  }
+	    /* libdb is not guaranteed to produce null terminated strings */
+	    char *pwhash = strndup(data.dptr, data.dsize);
 
+	    if (pwhash == NULL) {
+	      pam_syslog(pamh, LOG_CRIT, "strndup failed: data.dptr");
+	    } else {
+#ifdef HAVE_CRYPT_R
+	      struct crypt_data *cdata = NULL;
+	      cdata = malloc(sizeof(*cdata));
+	      if (cdata == NULL) {
+	        pam_syslog(pamh, LOG_CRIT, "malloc failed: struct crypt_data");
+	      } else {
+	        cdata->initialized = 0;
+	        cryptpw = crypt_r(pass, pwhash, cdata);
+	      }
+#else
+	      cryptpw = crypt (pass, pwhash);
+#endif
+	      if (cryptpw && strlen(cryptpw) == (size_t)data.dsize) {
+	        compare = memcmp(data.dptr, cryptpw, data.dsize);
+	      } else {
+	        if (ctrl & PAM_DEBUG_ARG) {
+	          if (cryptpw) {
+	            pam_syslog(pamh, LOG_INFO, "lengths of computed and stored hashes differ");
+	            pam_syslog(pamh, LOG_INFO, "computed hash: %s", cryptpw);
+	          } else {
+	            pam_syslog(pamh, LOG_ERR, "crypt() returned NULL");
+	          }
+	        }
+	      }
+#ifdef HAVE_CRYPT_R
+	      free(cdata);
+#endif
+	    }
+	    free(pwhash);
+	  }
 	} else {
 
 	  /* Unknown password encryption method -
--- a/modules/pam_userdb/pam_userdb.8.xml
+++ b/modules/pam_userdb/pam_userdb.8.xml
@@ -100,7 +100,8 @@
         </term>
         <listitem>
           <para>
-            Print debug information.
+            Print debug information. Note that password hashes, both from db
+            and computed, will be printed to syslog.
           </para>
         </listitem>
       </varlistentry>


modules/pam_userdb/pam_userdb.c
```diff
@@ -213,30 +213,39 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
...
 	    if (cryptpw && strlen(cryptpw) == (size_t)data.dsize) {
 	      compare = memcmp(data.dptr, cryptpw, data.dsize);
 	    } else {
 	      compare = -2;
 	      if (ctrl & PAM_DEBUG_ARG) {
-		if (cryptpw)
+		if (cryptpw) {
 		  pam_syslog(pamh, LOG_INFO, "lengths of computed and stored hashes differ");
-		else
+		  pam_syslog(pamh, LOG_INFO, "computed hash: %s", cryptpw);
```
t8m, line 243: I do not think we ever want the password hashes to appear in the syslog.

t8m: Hmm, but we do print the hashes from the DB above with ctrl&PAM_DEBUG_ARG already. OK, then. But could you please amend the pam_userdb.8.xml so it prominently notes that with the debug option it will print password hashes from the db and computed from the entered password to the syslog?

modules/pam_userdb/pam_userdb.c
```diff
@@ -213,30 +213,39 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
...
+        } else {
 		  pam_syslog(pamh, LOG_INFO, "crypt() returned NULL");
+        }
 	      }
 	    }
-#ifdef HAVE_CRYPT_R
-	    free(cdata);
```
t8m, line 238: You cannot move the free, it needs to stay here.

t8m (changes requested):

modules/pam_userdb/pam_userdb.c
```diff
@@ -213,30 +213,39 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
 	  } else if (ctrl & PAM_ICASE_ARG) {
 	    compare = -2;
 	  } else {
+		char *setting = malloc(data.dsize+1);
+		if (setting != NULL) {
+			strncpy(setting, data.dptr, data.dsize);
+			setting[data.dsize] = '\0';
+			if (ctrl & PAM_DEBUG_ARG) {
+				pam_syslog(pamh, LOG_INFO, "crypt setting: %s", setting);
+			}
```
t8m, line 222: There are multiple problems here:
1. use strndup instead of malloc+strncpy
2. do not call it setting, that is misleading, call it salt or something
3. no need to syslog it as there is syslog of it above already
4. please correct the indentation so the whole block is aligned

ep69: Ad 2.: I used it according to crypt_r documentation in man page and source code. I wasn't sure because `salt` might be misleading for some hashes, but I can rename it, you surely know way more about it.
Ad 4.: The whole file is a mess with respect to indentation, a lot of mixing of tabs and spaces. I wanted to make the change as minimal as possible so it is easy to review, basically using the same (mis-)indentation as in the lines around. I will be happy to fix it if you provide me some guidance - Should I do it withing this PR in separate commit or create a new one? Should I use tab or spaces (and how much?) - I wasn't able to find any guildeline. Should I fix it in whole file, just in the function, or just the respective block?

t8m: `setting` might be correct if you're producing a new password hash. Here it is definitely not just the setting, it is the whole password hash from the DB. Use `pwhash` or something.

For 4. I mean just adding an indentation level to the whole block if you're adding an `if`. Not just indenting properly the new code but leaving the existing code that is inside this new block unindented as you did. Of course keep the rest of the code as is.

modules/pam_userdb/pam_userdb.c
ep69: Could you tell me why? It is not obvious to me and I'd like to learn more not to make similar mistakes in future.

t8m: Just read the crypt_r manpage and what it tells about the data argument and the returned value from the crypt_r. The returned value is pointing into the cdata.

modules/pam_userdb/pam_userdb.c
```diff
@@ -217,28 +217,34 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
...
-	    }
+	      cdata->initialized = 0;
+	      /* libdb is not guaranteed to produce null terminated strings */
+	      char *pwhash = strndup(data.dptr, data.dsize);
+	      if (pwhash != NULL) {
+	        cryptpw = crypt_r(pass, pwhash, cdata);
+	      }
 #else
-	    cryptpw = crypt (pass, data.dptr);
+	      cryptpw = crypt (pass, pwhash);
```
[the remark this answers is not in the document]
ep69: Thanks for catching, @ldv-alt, it starts to be a madness with scopes and `ifdef`s. I changed the `pwhash` scope to the whole else branch. Could you have a look at updated version?

modules/pam_userdb/pam_userdb.c
```diff
@@ -213,32 +213,38 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
 	  } else if (ctrl & PAM_ICASE_ARG) {
 	    compare = -2;
 	  } else {
+	    /* libdb is not guaranteed to produce null terminated strings */
+	    char *pwhash = strndup(data.dptr, data.dsize);
+	    if (pwhash != NULL) {
```
t8m, line 218: Please add an empty line after the declaration and also add compare = -2 otherwise if the strndup fails compare == 0 and that will signalize a password match! Perhaps also some syslog when the strndup fails might be appropriate.

ep69: Thanks for spotting and your patience. I changed the code a bit to initialize `compare` to `-2` to be more safe against these kinds of overlooks. Does it make sense?

ikerexxe: I don't think `malloc` or any other memory allocation function is safe enough and they should be checked in case of failure. Take into account that pam is a fundamental software in Linux distributions and any unexplained failure can lead to hundreds of hours of work to reproduce and understand the failure.

ep69: @ikerexxe There might be a misunderstanding. Of course the return value must be checked. We are talking about adding syslog message in case allocation fails.
Are we on the same page? Are you proposing that syslog message should be added for all the cases where allocation fails?

ikerexxe: > @ikerexxe There might be a misunderstanding. Of course the return value must be checked. We are talking about adding syslog message in case allocation fails.

Sorry, I had understood something different. 

> Are we on the same page? Are you proposing that syslog message should be added for all the cases where allocation fails?

If possible I'd add a syslog message for all allocations that fail.

ep69: @ikerexxe ok, added syslog messages

modules/pam_userdb/pam_userdb.c
```diff
@@ -209,36 +209,48 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
...
-	    cdata = malloc(sizeof(*cdata));
-	    if (cdata != NULL) {
-		cdata->initialized = 0;
-		cryptpw = crypt_r(pass, data.dptr, cdata);
-	    }
+	      struct crypt_data *cdata = NULL;
+	      cdata = malloc(sizeof(*cdata));
+	      if (cdata == NULL) {
+	        pam_syslog(pamh, LOG_INFO, "malloc failed: struct crypt_data");
+          } else {
```
t8m, line 229: Wrong indentation on this line

t8m (commented):
Please just fix the nit with the indentation and this can go in.

modules/pam_userdb/pam_userdb.c
```diff
@@ -209,36 +209,48 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
...
 	  } else if (ctrl & PAM_ICASE_ARG) {
-	    compare = -2;
+	    pam_syslog(pamh, LOG_INFO,
+	       "case-insensitive comparison only works with plaintext passwords");
 	  } else {
+	    /* libdb is not guaranteed to produce null terminated strings */
+	    char *pwhash = strndup(data.dptr, data.dsize);
+
+	    if (pwhash == NULL) {
+	      pam_syslog(pamh, LOG_INFO, "strndup failed: data.dptr");
```
t8m, line 222: I'm sorry, one more thing. We use LOG_CRIT for memory allocation errors in syslog in other modules.

modules/pam_userdb/pam_userdb.c
```diff
@@ -209,36 +209,48 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
...
-	    struct crypt_data *cdata = NULL;
-	    cdata = malloc(sizeof(*cdata));
-	    if (cdata != NULL) {
-		cdata->initialized = 0;
-		cryptpw = crypt_r(pass, data.dptr, cdata);
-	    }
+	      struct crypt_data *cdata = NULL;
+	      cdata = malloc(sizeof(*cdata));
+	      if (cdata == NULL) {
+	        pam_syslog(pamh, LOG_INFO, "malloc failed: struct crypt_data");
```
t8m, line 228: LOG_CRIT here as well.

modules/pam_userdb/pam_userdb.c
```diff
@@ -209,36 +209,48 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
...
-		  pam_syslog(pamh, LOG_INFO, "crypt() returned NULL");
+	      if (cryptpw && strlen(cryptpw) == (size_t)data.dsize) {
+	        compare = memcmp(data.dptr, cryptpw, data.dsize);
+	      } else {
+	        if (ctrl & PAM_DEBUG_ARG) {
+	          if (cryptpw) {
+	            pam_syslog(pamh, LOG_INFO, "lengths of computed and stored hashes differ");
+	            pam_syslog(pamh, LOG_INFO, "computed hash: %s", cryptpw);
+	          } else {
+	            pam_syslog(pamh, LOG_INFO, "crypt() returned NULL");
```
t8m, line 244: And LOG_ERR here please, as you're already touching the code.

t8m (changes requested):

modules/pam_userdb/pam_userdb.c
ep69: Thanks, I found one more wrongly indented one and fixed. Unrelated question - would you accept a PR fixing the indentation of the whole file? If so, what are the guidelines?

t8m (approved):

modules/pam_userdb/pam_userdb.c
t8m: That's the problem. There are no guidelines for code format in Linux-PAM. So I'd recommend not doing that PR. At least not until we establish some official code format guideline for Linux-PAM.

t8m: Up until now it was something like - follow the existing code format in the file you edit.  Of course in case some file is just a mess of different code formats we would probably even without a guideline accept a PR that would unify the format within the file to the code format the majority of the file is already in.

ep69: Thanks for explaining. I won't hujer into things.

modules/pam_userdb/pam_userdb.c
```diff
@@ -209,36 +209,48 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
 	  char *cryptpw = NULL;
 
 	  if (data.dsize < 13) {
-	    compare = -2;
+	    /* hash from db too short */
+	    pam_syslog(pamh, LOG_INFO, "password hash in database too short");
```
ldv-alt, line 213: grammar: password hash in the database is too short

ldv-alt (changes requested):
Overall looks good.

t8m (approved):

ldv-alt (approved):

t8m:
Merged, thank you for the contribution!

ep69:
Thanks for your patience and guidance again, Tomasi! Highly appreciated.

PR #356, opened 2021-04-20
LOGIN_STRING: use for password prompt string

ikerexxe:
modules/pam_pwhistory/pam_pwhistory.c: use LOGIN_STRING from login.defs
file as the string for the password prompt.
modules/pam_unix/pam_unix_auth.c: likewise.
modules/pam_unix/pam_unix_passwd.c: likewise.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1884702

### DIFF
--- a/modules/pam_pwhistory/pam_pwhistory.c
+++ b/modules/pam_pwhistory/pam_pwhistory.c
@@ -64,6 +64,8 @@
 #include "opasswd.h"
 #include "pam_inline.h"
 
+#define LOGIN_DEFS "/etc/login.defs"
+
 struct options_t {
   int debug;
   int enforce_for_root;
@@ -292,6 +294,8 @@ pam_sm_chauthtok (pam_handle_t *pamh, int flags, int argc, const char **argv)
   const char *user;
     int retval, tries;
   options_t options;
+  char *login_string = NULL;
+  char *prompt = NULL;
 
   memset (&options, 0, sizeof (options));
 
@@ -331,13 +335,29 @@ pam_sm_chauthtok (pam_handle_t *pamh, int flags, int argc, const char **argv)
   if (retval != PAM_SUCCESS)
     return retval;
 
+  login_string = pam_modutil_search_key(pamh, LOGIN_DEFS, "LOGIN_STRING");
+  if (login_string != NULL)
+    {
+      retval = asprintf(&prompt, login_string, user);
+      if (retval == -1)
+        {
+          prompt = NULL;
+        }
+
+      free(login_string);
+    }
+
   newpass = NULL;
   tries = 0;
   while ((newpass == NULL) && (tries < options.tries))
     {
-      retval = pam_get_authtok (pamh, PAM_AUTHTOK, &newpass, NULL);
+      retval = pam_get_authtok (pamh, PAM_AUTHTOK, &newpass, prompt);
       if (retval != PAM_SUCCESS && retval != PAM_TRY_AGAIN)
 	{
+	  if (prompt != NULL)
+	    {
+	      free(prompt);
+	    }
 	  if (retval == PAM_CONV_AGAIN)
 	    retval = PAM_INCOMPLETE;
 	  return retval;
@@ -377,6 +397,11 @@ pam_sm_chauthtok (pam_handle_t *pamh, int flags, int argc, const char **argv)
 	    pam_info (pamh,
 		       _("Password has been already used."));
 	}
+    }
+
+  if (prompt != NULL)
+    {
+      free(prompt);
     }
 
   if (newpass == NULL && tries >= options.tries)
--- a/modules/pam_unix/pam_unix_auth.c
+++ b/modules/pam_unix/pam_unix_auth.c
@@ -53,6 +53,7 @@
 #include <security/_pam_macros.h>
 #include <security/pam_modules.h>
 #include <security/pam_ext.h>
+#include <security/pam_modutil.h>
 
 #include "support.h"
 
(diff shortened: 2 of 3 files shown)


t8m (commented):
Could you please create a helper function to avoid the repetition of the same code? Or, the question is whether this should be handled in the pam_unix module or the pam_get_authtok() function itself.

t8m:
@ldv-alt any opinion?

ldv-alt:
I'd like to see the rationale for this new feature.
Yes, I see a reference to RH bugzilla, but I'd prefer the PR to be consistent.

With regards to the feature itself, I didn't know `LOGIN_STRING` thing existed, it is **not** mentioned in the login.defs(5) manual page installed in my system, although it seems to be present in shadow.git since 2007.

The description of `LOGIN_STRING` in shadow.git tells that `If the string contains %s, this will be replaced by the user's name.` - do they really implement it, and if yes, do you suggest to implement that in pam, too?

Given that `pam_get_authtok` has different default prompts depending on the `item` argument, the single `LOGIN_STRING` doesn't look as a good design decision.

ikerexxe:
> I'd like to see the rationale for this new feature.
> Yes, I see a reference to RH bugzilla, but I'd prefer the PR to be consistent.

The main idea behind this PR is to give user's the ability to configure the prompt string that asks for the password. This prompt can be configured in such a way that it can also contain the user's name.
 
> With regards to the feature itself, I didn't know `LOGIN_STRING` thing existed, it is **not** mentioned in the login.defs(5) manual page installed in my system, although it seems to be present in shadow.git since 2007.

As you have already mentioned it's been present in the repository for several years. In my system (Fedora33) I can see the following in the man page:
```
LOGIN_STRING (string)
           The string used for prompting a password. The default is to use "Password: ", or a translation of that string. If you set this variable, the prompt will not be translated.

           If the string contains %s, this will be replaced by the user's name.
```
 
> The description of `LOGIN_STRING` in shadow.git tells that `If the string contains %s, this will be replaced by the user's name.` - do they really implement it, and if yes, do you suggest to implement that in pam, too?

Yes, this is implemented in https://github.com/shadow-maint/shadow/blob/b30e9614c3dff4823f7d0989b64ecb80e2434612/lib/pwauth.c#L163 by shadow-utils itself. And yes, I propose to follow the same approach also in pam.
 
> Given that `pam_get_authtok` has different default prompts depending on the `item` argument, the single `LOGIN_STRING` doesn't look as a good design decision.

IIUC, for the modules that I've changed (pam_unix and pam_pwhistory) the `item` argument is set to `PAM_AUTHTOK` or `PAM_OLDAUTHTOK` which means that the argument is used only to reuse a password provided by a previous module. If no password has been provided by a previous module then the prompt is used, and this is the part that I'm changing. Moreover, I've checked it and the pam stack configured in Fedora enables me to configure the prompt without any change to the stack itself.

Can you explain a little bit more why you think it's a bad design decision? I'd like to understand it better.

ldv-alt:
> > With regards to the feature itself, I didn't know `LOGIN_STRING` thing existed, it is **not** mentioned in the login.defs(5) manual page installed in my system, although it seems to be present in shadow.git since 2007.
> 
> As you have already mentioned it's been present in the repository for several years. In my system (Fedora33) I can see the following in the man page:

The reason why `LOGIN_STRING` is not mentioned at all in my login.defs(5) but mentioned in yours is simple:
this part of documentation is conditionalised on `no_pam`, my shadow package is built using `--with-libpam` configure option, while yours is apparently built using `--without-libpam` configure option, which is an unusual choice for a heavily pam-based operating system like Fedora.

> > The description of `LOGIN_STRING` in shadow.git tells that `If the string contains %s, this will be replaced by the user's name.` - do they really implement it, and if yes, do you suggest to implement that in pam, too?
> 
> Yes, this is implemented in https://github.com/shadow-maint/shadow/blob/b30e9614c3dff4823f7d0989b64ecb80e2434612/lib/pwauth.c#L163 by shadow-utils itself. And yes, I propose to follow the same approach also in pam.

They implement it by taking the string from `login.defs` and passing it unchecked as a format string directly to `snprintf`.
This approach is error-prone, potential errors are hard to debug, and `%s` doesn't have connotations with username.

> > Given that `pam_get_authtok` has different default prompts depending on the `item` argument, the single `LOGIN_STRING` doesn't look as a good design decision.
> 
> IIUC, for the modules that I've changed (pam_unix and pam_pwhistory) the `item` argument is set to `PAM_AUTHTOK` or `PAM_OLDAUTHTOK` which means that the argument is used only to reuse a password provided by a previous module. If no password has been provided by a previous module then the prompt is used,

This is true for all users of `pam_get_authtok`.

> and this is the part that I'm changing.

Yes, and in that case `pam_get_authtok` is currently capable of choosing from a variety of default translated prompts, see
`git grep '#define PROMPT'`.  The `LOGIN_STRING` approach would replace this variety with a single string.

t8m:
> The reason why `LOGIN_STRING` is not mentioned at all in my login.defs(5) but mentioned in yours is simple:
> this part of documentation is conditionalised on `no_pam`, my shadow package is built using `--with-libpam` configure option, while yours is apparently built using `--without-libpam` configure option, which is an unusual choice for a heavily pam-based operating system like Fedora.

On Fedora/RHEL the login utilities from shadow package are not used and the remaining utilities are sysadmin-only which means there is no point in linking them to libpam. The right workaround would be to just remove those variables from documentation with a downstream Fedora patch. I am almost sure we did that already at least at some point but maybe this regressed or I do not remember it correctly.

t8m:
I agree with @ldv-alt that this is a misfeature and it should be rejected.

ikerexxe:
After reviewing it and taking into account your feedback, I agree that it doesn't make much sense to implement this change. Thank your for your explanation and for pointing me in the right direction.

PR #363, opened 2021-05-07
libpam_misc:misc_conv() buffer length

jsquyres:
# Short version

The `libpam_misc:misc_conv()` password buffer is fixed at 512 bytes long.  This PR is two commits:

1. Add a `--with-misc-conv-bufsize` configure CLI option to allow setting the `misc_conv()` buffer size to an arbitrary value (e.g., 4096), but keep the default length at 512 bytes.
2. Change the default length of the `misc_conv()` buffer to 4096 bytes.

The PR is two commits just to show that the 2nd commit could be omitted if there would be problems increasing the default buffer length.

# More detail

We have an internal use case to use Linux PAM for a cryptographic challenge/response login.  It looks something like this:

```
$ ssh someuser@somehost 
Challenge: ...a very long text blob...
Response: <wait for user input>
```

The user copies the challenge blob to an internal authorization system.  If the user is authorized, a response text blob will be generated (typically ~1KB in length).  The user then pastes the response blob to the `ssh` login "Response:" prompt.  Our PAM module validates the response, and returns PAM_SUCCESS on success, or an appropriate failure code.

This works great in an ssh context, because `sshd` provides its own PAM conversation callback function with a variable-length buffer (specifically: `sshd` does not use `libpam_misc:misc_conv()`).  It can handle ~1KB text blobs, for example.

However, the Linux `login` process -- used for console logins -- uses `libpam_misc:misc_conv()`.  Our challenge-response PAM module therefore fails because `misc_conv()` truncates the conversation response buffer to 512 bytes.  It is worth noting that copy-and-paste is available in some flavors of console login, such as KVM consoles and serial-over-LAN console logins.  Hence, we're interested in making our challenge-response PAM module work properly even for console logins.

If the Linux PAM community is open to the concept, the first commit provides a configure CLI option to set the size of `libpam_misc`'s `misc_conv()` buffer to an arbitrary value, but keeps the default length at 512.  This would be a "good enough" solution for our use case: we can recompile `libpam_misc` for our systems that use this challenge-response mechanism, and possibly even convince Linux distros to also use `--with-misc-conv-bufsize=4096` when building their PAM packages.

But if possible, it would be great if the default `misc_conv()` buffer size could be increased -- e.g., to 4K, as in the second commit.  This would enable us to (eventually) use `libpam_misc` out of the box with any Linux distro that updates to a release that contains this commit.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -634,9 +634,9 @@ esac
 
 AC_ARG_WITH([misc-conv-bufsize],
 AS_HELP_STRING([--with-misc-conv-bufsize=<number>],
-    [Size of input buffer for libpam_misc's misc_conv() conversation function, default=512]),
+    [Size of input buffer for libpam_misc's misc_conv() conversation function, default=4096]),
     [],
-    [with_misc_conv_bufsize=512])
+    [with_misc_conv_bufsize=4096])
 AC_DEFINE_UNQUOTED(PAM_MISC_CONV_BUFSIZE, $with_misc_conv_bufsize, [libpam_misc misc_conv() buffer size.])
 
 AM_CONDITIONAL([COND_BUILD_PAM_KEYINIT], [test "$have_key_syscalls" = 1])
--- a/configure.ac
+++ b/configure.ac
@@ -632,6 +632,13 @@ case "$enable_unix" in
   *) AC_MSG_ERROR([bad value $enable_unix for --enable-unix option]) ;;
 esac
 
+AC_ARG_WITH([misc-conv-bufsize],
+AS_HELP_STRING([--with-misc-conv-bufsize=<number>],
+    [Size of input buffer for libpam_misc's misc_conv() conversation function, default=512]),
+    [],
+    [with_misc_conv_bufsize=512])
+AC_DEFINE_UNQUOTED(PAM_MISC_CONV_BUFSIZE, $with_misc_conv_bufsize, [libpam_misc misc_conv() buffer size.])
+
 AM_CONDITIONAL([COND_BUILD_PAM_KEYINIT], [test "$have_key_syscalls" = 1])
 AM_CONDITIONAL([COND_BUILD_PAM_LASTLOG], [test "$ac_cv_func_logwtmp" = yes])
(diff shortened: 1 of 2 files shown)


jsquyres:
Greetings Linux PAM community.  Any thoughts on this PR?

I think GitHub is also not running all CI actions for first-time contributors by default any more; I think someone needs to click a button to allow the rest of CI for this PR to run.  Thanks!

jsquyres:
Could someone check the Travis job?  It started 1.5 weeks ago, and while a bunch of the jobs completed successfully, some of them are still in the "Job received" state (i.e., they didn't even make it to the "queued" state).  As a new contributor, I don't think that I have the permissions to re-submit the job.

Thanks!

t8m:
The Travis CI is unusable, we should just drop it.

jsquyres:
Greetings all.  Any thoughts or comments on this PR?

t8m:
I am really not sure if there is any downside of this change apart from allocating 4kB on the stack. There are definitely modules that will truncate the authtok obtained from the conversation function to the PAM_MAX_RESP_SIZE and I do not think we should change that value. But that should not be a problem.

So I am tentatively thinking we could approve both commits.

@ldv-alt any other concerns?

configure.ac
```diff
@@ -632,6 +632,11 @@ case "$enable_unix" in
   *) AC_MSG_ERROR([bad value $enable_unix for --enable-unix option]) ;;
 esac
 
+AC_ARG_WITH([misc-conv-bufsize],
+AS_HELP_STRING([--with-misc-conv-bufsize=<number>],[Size of input buffer for libpam_misc's misc_conv() conversation function, default=4096]),
+[], [with_misc_conv_bufsize=4096])
```
ldv-alt, line 637: The line is too long, please wrap and indent.

ldv-alt (changes requested):
On the whole, I don't see any downsides from this change.

configure.ac
jsquyres: Done!

ldv-alt (approved):

t8m:

PR #365, opened 2021-05-21
pam.conf: clarify default action for unspecified return codes

jsquyres:
Add short blurbs explaining that if a return code is not specified in
the "[value1=action1 value2=action2 ...]" form and "default=action" is
not specified, that return code's action defaults to "bad".

### DIFF
--- a/doc/man/pam.conf-syntax.xml
+++ b/doc/man/pam.conf-syntax.xml
@@ -265,7 +265,8 @@
              this action indicates that the return code should be thought
              of as indicative of the module failing. If this module is the
              first in the stack to fail, its status value will be used for
-             that of the whole stack.
+             that of the whole stack.  This is the default action for
+             all return codes.
           </para>
         </listitem>
       </varlistentry>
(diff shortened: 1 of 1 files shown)


doc/man/pam.conf-syntax.xml
```diff
@@ -265,7 +265,10 @@
              this action indicates that the return code should be thought
              of as indicative of the module failing. If this module is the
              first in the stack to fail, its status value will be used for
-             that of the whole stack.
+             that of the whole stack.  This is the default action for
+             all return codes, unless overridden via a
+             <emphasis>valueN</emphasis> token or the
+             <emphasis>default</emphasis> token.
```
t8m, line 271: I think this is a little bit confusing given the context. It also repeats the information given below. So I'd just add `This is the default action for all return codes.` here.

t8m (changes requested):

doc/man/pam.conf-syntax.xml
jsquyres: Done!

jsquyres:
I made the requested changes; thanks!

Can someone approve the CI workflow on this PR?  I'm a first-time contributor, so Github won't automatically launch CI until a maintainer approves it.  Thank you!

t8m (approved):

t8m:
Merged. Thank you for the contribution.

Issue #367, opened 2021-06-10
pam_unix: should not treat CRYPT_SALT_METHOD_LEGACY hash as expired


msharov:
libxcrypt 4.4.22 has deprecated multiple hashes, including SHA-256, resulting in login failures for many people. The failures are uninformative:
```
PAM failed: Authentication token is no longer valid; new one required
pam_unix(login:account): expired password for user (root enforced)
pam_chauthtok: Authentication token manipulation error
```
The messages do not indicate what the real problem is and the only way I was able to figure it out was by reading the news board on my distribution's home page (Arch). The token can not be made valid if pam configuration explicitly sets the hash, as the documentation previously recommended. Until the configuration changes, logins will continue to fail. If the root partition is mounted read-only, the system will be completely unbootable.

A more informative error message is needed to tell people to change the hash setting (or, in my case, downgrade libxcrypt and blacklist its upgrades until they get over their God complex) instead of helplessly banging their heads on the keyboard.

t8m:
I have to say that I was disappointed by the libxcrypt change as well. IMO this was not well managed there. In the end it would be much better if this was handled somehow in coordination with Linux PAM project.

We could drop the crypt_checksalt() call completely in pam_unix check_shadow_expiry(). It was probably not a right thing there.

ldv-alt:
On Thu, Jun 10, 2021 at 07:02:58AM -0700, Tomáš Mráz wrote:
> I have to say that I was disappointed by the libxcrypt change as well. IMO this was not well managed there. In the end it would be much better if this was handled somehow in coordination with Linux PAM project.

The libxcrypt change itself was OK, but ...

> We could drop the crypt_checksalt() call completely in pam_unix check_shadow_expiry(). It was probably not a right thing there.

... pam_unix check_shadow_expiry() is not correct in treating
CRYPT_SALT_METHOD_LEGACY hash as expired.

zackw:
libxcrypt author here.

The intended meaning of CRYPT_SALT_LEGACY is "`passwd(1)` should not use this hashing method."  It is _not_ supposed to mean "force a password change on next login for any user with an existing stored hash using this method."  At most, `login(1)` should _silently_ re-hash the user's _existing_ password using a stronger method upon successful authentication, and if it can't update the password database (e.g. because the root filesystem is read-only) then it should just leave things alone.

ldv-alt:
[...]

PR #368, opened 2021-06-10
pam_unix: do not use crypt_checksalt when checking for password expiration

ldv-alt:
According to Zack Weinberg, the intended meaning of
CRYPT_SALT_METHOD_LEGACY is "passwd(1) should not use this hashing
method", it is not supposed to mean "force a password change on next
login for any user with an existing stored hash using this method".

This reverts commit 4da9febc39b955892a30686e8396785b96bb8ba5.

* modules/pam_unix/passverify.c (check_shadow_expiry)
[CRYPT_CHECKSALT_AVAILABLE]: Remove.

Closes: https://github.com/linux-pam/linux-pam/issues/367


t8m (approved):
LGTM.

PR #373, opened 2021-06-29
Permit unix_chkpwd & pam_unix.so to run without being setuid-root.

AndrewGMorgan:
Remove the hard-coding of the idea that the only way pam_unix.so can
read the shadow file is if it can, in some way, run setuid-root.
Linux capabilities only require cap_dac_override to read the /etc/shadow
file.

This change achieves two things: it opens a path for a linux-pam
application to run without being setuid-root; further, it allows
unix_chkpwd to run non-setuid-root if it is installed:

   sudo setcap cap_dac_override=ep unix_chkpwd

If we wanted to link against libcap, we could install this binary with
cap_dac_override=p, and use cap_set_proc() to raise the effective bit
at runtime. However, some distributions already link unix_chkpwd
against libcap-ng for some, likely spurious, reason so "ep" is fine
for now.

Signed-off-by: Andrew G. Morgan <morgan@kernel.org>

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -241,12 +241,16 @@ PAMH_ARG_DECL(int get_account_info,
 			 * ...and shadow password file entry for this user,
 			 * if shadowing is enabled
 			 */
-#ifndef HELPER_COMPILE
-			if (geteuid() || SELINUX_ENABLED)
-				return PAM_UNIX_RUN_HELPER;
-#endif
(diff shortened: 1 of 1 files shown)


t8m (approved):

t8m:
@ldv-alt This LGTM so I'll merge unless you have any comments, objections.

ldv-alt (approved):

t8m:
Merged to master. Thank you for the pull request!

Issue #379, opened 2021-07-23
Incompatibility with systemd 249


Arfrever:
When systemd >=249 is installed and `nsswitch.conf` contains:
```
shadow:     files systemd
```
And when PAM code is run by process with non-root user, authentication will fail.
See https://bugs.gentoo.org/803050 for more details.

floppym:
To summarize:

Commit f220cace205332a3dc34e7b37a85e7627e097e7d changed pam_unix so that it only executes unix_chkpwd if getspnam sets errno to EACCES.

This relies on libnss_files setting errno to EACCES when /etc/shadow cannot be opened.

If libnss_files is not the last NSS module listed for the shadow database, subsequent NSS modules (like libnss_systemd) may overwrite errno with some other value (like 0).

ldv-alt:
On Fri, Jul 23, 2021 at 01:17:00PM -0700, Mike Gilbert wrote:
> Commit f220cace205332a3dc34e7b37a85e7627e097e7d changed pam_unix so that it only executes unix_chkpwd if getspnam sets errno to EACCES.
> 
> This relies on libnss_files setting errno to EACCES when /etc/shadow cannot be opened.
> 
> If libnss_files is not the last NSS module listed for the shadow database, subsequent NSS modules (like libnss_systemd) may overwrite errno with some other value (like 0).

nss modules are not permitted to do that because such behaviour
would violate the getspnam(3) contract.  Please file a bug report
to the incompatible nss module.

floppym:
What exactly do you mean by "violate contract"?

The man page mentions EACCES if /etc/shadow cannot be opened, but that does not really consider the possibility of NSS modules that do not even attempt to open this file .

ldv-alt:
On Sun, Jul 25, 2021 at 08:19:04AM -0700, Mike Gilbert wrote:
> What exactly do you mean by "violate contract"?
> 
> The man page mentions EACCES if /etc/shadow cannot be opened, but that does not really consider the possibility of NSS modules that do not even attempt to open this file .

The manual page says that EACCES shall be set when the caller does not
have permission to access the shadow password file.

In this case getspnam/getspnam_r tried to access the shadow password file
but, apparently, the error code was lost.

ldv-alt:
[...]

PR #386, opened 2021-08-20
pam_unix: workaround the problem caused by libnss_systemd

ldv-alt:
The getspnam(3) manual page says that errno shall be set to EACCES when
the caller does not have permission to access the shadow password file.
Unfortunately, this contract is broken when libnss_systemd is used in
the nss stack.

Workaround this problem by falling back to the helper invocation when
pam_modutil_getspnam returns NULL regardless of errno.  As pam_unix
already behaves this way when selinux is enabled, it should be OK
for the case when selinux is not enabled, too.

* modules/pam_unix/passverify.c (get_account_info): When
pam_modutil_getspnam returns NULL, unconditionally fall back
to the helper invocation.

Complements: f220cace2053 ("Permit unix_chkpwd & pam_unix.so to run without being setuid-root")
Resolves: https://github.com/linux-pam/linux-pam/issues/379

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -245,8 +245,7 @@ PAMH_ARG_DECL(int get_account_info,
 			if (*spwdent == NULL) {
 #ifndef HELPER_COMPILE
 				/* still a chance the user can authenticate */
-				if (errno == EACCES || SELINUX_ENABLED)
-					return PAM_UNIX_RUN_HELPER;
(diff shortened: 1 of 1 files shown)


t8m (approved):

t8m:
This is a good idea as this makes the behavior of the pam_unix more simple and predictable.

floppym:
Thanks for adding this workaround.

PR #388, opened 2021-08-29
configure.ac: also search libcrypt through pkg-config

ffontaine:
libxcrypt provides a libcrypt.pc file so use it if available as this will allow to retrieve the library path (e.g. `-L/home/buildroot/output/host//riscv64-buildroot-linux-musl/sysroot/usr/lib`) which is useful when cross-compiling and will avoid the following build failure on buildroot:

```
/home/buildroot/autobuild/run/instance-3/output-1/host/opt/ext-toolchain/bin/../lib/gcc/riscv64-buildroot-linux-musl/10.2.0/../../../../riscv64-buildroot-linux-musl/bin/ld: .libs/passverify.o: in function `.L30':
passverify.c:(.text+0x368): undefined reference to `crypt_checksalt'
```

Fixes:
 - http://autobuild.buildroot.org/results/20b14e222b35c2d1269960075832b784ba81aa1a

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -363,8 +363,14 @@ AC_CHECK_HEADERS(crypt.h)
 
 BACKUP_LIBS=$LIBS
 LIBCRYPT=""
-AC_SEARCH_LIBS([crypt_gensalt_rn],[crypt])
-case "$ac_cv_search_crypt_gensalt_rn" in
+PKG_CHECK_MODULES([CRYPT], [libcrypt], [
+  CFLAGS="$CFLAGS $CRYPT_CFLAGS"
+  CPPFLAGS="$CPPFLAGS $CRYPT_CFLAGS"
+  LIBS="$LIBS $CRYPT_LIBS"
+  LIBCRYPT="$CRYPT_LIBS"
+], [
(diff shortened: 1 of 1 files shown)


t8m (approved):
LGTM. @ldv-alt what do you think?

configure.ac
```diff
@@ -374,7 +379,8 @@ case "$ac_cv_search_crypt_gensalt_rn" in
 			-l*) LIBCRYPT="$ac_cv_search_crypt" ;;
 		esac ;;
 	esac ;;
-esac
+  esac
+])
 AC_CHECK_FUNCS([crypt_r])
```
ldv-alt, line 384: While `AC_SEARCH_LIBS([crypt_gensalt_rn],[crypt])` sets `LIBS`, `LIBCRYPT="$CRYPT_LIBS"` doesn't, so `AC_CHECK_FUNCS([crypt_r])` fails to find `crypt_r` in -lcrypt.

ldv-alt (changes requested):

configure.ac
ffontaine: Good catch, I updated the PR.

ldv-alt (approved):

PR #396, opened 2021-10-20
pam_pwhistory: Enable alternate location for password history file

mwelchuk:
Sometimes, especially in embedded devices, the /etc directory can be
read-only and/or not saved over upgrades. In order to ensure password
policies are maintained across upgrades and the module functions on
read-only file systems, allow the location of the password history file
to be set in the PAM configuration.

Signed-off-by: Edward <jinzhou.zhu1@ge.com>
[Martyn Welch: Updated commit message and ported to latest version]
Signed-off-by: Martyn Welch <martyn.welch@collabora.com>

### DIFF
--- a/modules/pam_pwhistory/opasswd.c
+++ b/modules/pam_pwhistory/opasswd.c
@@ -44,6 +44,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <stdio.h>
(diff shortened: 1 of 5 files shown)


ldv-alt (approved):
I fixed all bugs that I found in this patch, feel free to proceed with this PR.

ldv-alt:
It would be great if somebody who uses this module tested these changes.

PR #398, opened 2021-10-27
run-xtests.sh: check the file exist or not

Mingli-Yu:
Fixes:
 \#  ./run-xtests.sh . tst-pam_access1
 mv: cannot stat '/etc/security/opasswd': No such file or directory
 PASS: tst-pam_access1
 mv: cannot stat '/etc/security/opasswd-pam-xtests': No such file or directory
 ==================
 1 tests passed
 0 tests not run
 ==================

Signed-off-by: Mingli Yu <mingli.yu@windriver.com>

### DIFF
--- a/xtests/run-xtests.sh
+++ b/xtests/run-xtests.sh
@@ -18,10 +18,10 @@ all=0
 
 mkdir -p /etc/security
 for config in access.conf group.conf time.conf limits.conf ; do
-	cp /etc/security/$config /etc/security/$config-pam-xtests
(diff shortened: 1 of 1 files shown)


ldv-alt (changes requested):
Please wrap long lines, they normally shouldn't be longer than 80 symbols.

Please correct the commit message grammar, it should say something like "check whether files exist" rather than "check the file exist".

Mingli-Yu:
Use new pull request https://github.com/linux-pam/linux-pam/pull/404 to track.

PR #401, opened 2021-10-29
Only include vendordir in manual page if set

thkukuk:
Compile in %vendordir%/pam.d section only if vendordir was enabled.


t8m (approved):
LGTM

ldv-alt (approved):
Haven't tested myself but looks good.

PR #402, opened 2021-10-29
Use vendor specific limits.conf as fallback

thkukuk:
pam_limits.so fails if it cannot read /etc/security/limits.conf. This is especially a problem if you use image based installations and updates (so image = /usr, not image = raw disk image). Use the vendor directory as fallback for a distribution provided default config.

### DIFF
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -817,21 +817,21 @@ parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
     fil = fopen(CONF_FILE, "r");
     if (fil == NULL) {
       int err = errno;
-#ifdef VENDORDIR
 
+#ifdef VENDORDIR
       /* if the specified file does not exist, and it is not provided by
-	 the user, try the vendor file as fallback. */
-      if (pl->conf_file == NULL)
-	fil = fopen(VENDORDIR"/security/limits.conf", "r");
+         the user, try the vendor file as fallback. */
+      if (pl->conf_file == NULL && err == ENOENT)
+        fil = fopen(VENDORDIR"/security/limits.conf", "r");
 
       if (fil == NULL)
 #endif
-	{
-	  pam_syslog (pamh, LOG_WARNING,
-		      "cannot read settings from %s: %s", CONF_FILE,
-		      strerror(err));
-	  return PAM_SERVICE_ERR;
-	}
+        {
+          pam_syslog (pamh, LOG_WARNING,
+                      "cannot read settings from %s: %s", CONF_FILE,
+                      strerror(err));
+          return PAM_SERVICE_ERR;
+        }
     }
 
     /* start the show */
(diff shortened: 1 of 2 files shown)


modules/pam_limits/pam_limits.c
```diff
@@ -816,9 +816,22 @@ parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
         pam_syslog(pamh, LOG_DEBUG, "reading settings from '%s'", CONF_FILE);
     fil = fopen(CONF_FILE, "r");
     if (fil == NULL) {
-        pam_syslog (pamh, LOG_WARNING,
-		    "cannot read settings from %s: %m", CONF_FILE);
-        return PAM_SERVICE_ERR;
+      int err = errno;
+#ifdef VENDORDIR
```
t8m, line 820: Please move this below the added empty line.

modules/pam_limits/pam_limits.c
```diff
@@ -816,9 +816,22 @@ parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
...
+	fil = fopen(VENDORDIR"/security/limits.conf", "r");
+
+      if (fil == NULL)
+#endif
+	{
+	  pam_syslog (pamh, LOG_WARNING,
+		      "cannot read settings from %s: %s", CONF_FILE,
+		      strerror(err));
+	  return PAM_SERVICE_ERR;
+	}
```
t8m, line 834: The indentation seems to be somehow weird here.

t8m (changes requested):

modules/pam_limits/pam_limits.c
```diff
@@ -816,9 +816,22 @@ parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
...
     fil = fopen(CONF_FILE, "r");
     if (fil == NULL) {
-        pam_syslog (pamh, LOG_WARNING,
-		    "cannot read settings from %s: %m", CONF_FILE);
-        return PAM_SERVICE_ERR;
+      int err = errno;
+#ifdef VENDORDIR
+
+      /* if the specified file does not exist, and it is not provided by
+	 the user, try the vendor file as fallback. */
```
ldv-alt, line 823: If the idea is to fall back when the file doesn't exist, then `errno` has to be checked to make sure it's `ENOENT` and not something else like `EACCES`.

ldv-alt (changes requested):

modules/pam_limits/pam_limits.c
thkukuk: If you remove the #if/#endif block, the indentation looks correct.
Since this part is outside of the if/endif block: I can let it as it is or adjust the indentation to the if/endif block. I'm fine with both.

t8m: IMO the indentation should be correct with the conditionally compiled block present.

t8m (approved):

ldv-alt (approved):
LGTM, please squash when merging.

PR #408, opened 2021-12-03
pam_env: use vendor specific pam_env.conf and environment as fallback

schubi2:
Use the vendor directory (/usr/etc) as fallback for a distribution provided default config if there is no one in /etc.

* Makefile.am: Add libeconf setting.

* pam_env.c: Take care about the fallback configuration in /usr/etc.

* pam_env.8.xml: Add /usr/etc description.

* pam_env.conf.5.xml: Add /usr/etc description.

* tst-pam_env-retval.c: General testcases for pam_env.

* tests: Contains data for testcases.


ldv-alt:
Could you make sure it passes CI, please?

modules/pam_env/pam_env.8.xml
```diff
@@ -52,13 +52,26 @@
...
       option.
     </para>
-    <para>
+    <para condition="with_vendordir">
+      Second a file (<filename>/usr/etc/environment</filename> by default) with simple
+      <emphasis>KEY=VAL</emphasis> pairs on separate lines will be read.
+      An alternate file is <filename>/etc/security/pam_env.conf</filename>
+      or can be specifed with the <emphasis>envfile</emphasis> option.
+      And with the <emphasis>readenv</emphasis> option this can be completely disabled.
+    </para>
```
thkukuk, line 73: Please also document /etc/security/pam_env.conf.d for the case we use libeconf, so that people know about all new features.

modules/pam_rootok/pam_rootok.c
```diff
@@ -53,11 +53,10 @@ static int
...
 {
-    int audit_fd;
     va_list ap;
 
 #ifdef HAVE_LIBAUDIT
-    audit_fd = audit_open();
+    int audit_fd = audit_open();
 
     if (audit_fd >= 0) {
 	char *buf;
```
thkukuk, line 62: Please make this fix an own PR, so that there is no conflict with the other PR containing the same fix.

thkukuk (changes requested):

ldv-alt:
Stefan, could you rebase and squash commits, please?

t8m:
Please do not add merge commits. Could you please rebase instead?

modules/pam_env/pam_env.8.xml
schubi2: I have updated the docu.

schubi2:
> Stefan, could you rebase and squash commits, please?

Yes, I have done it...hopefully correctly :-)

schubi2:
ups, sorry

thkukuk (approved):

modules/pam_rootok/pam_rootok.c
```diff
@@ -53,10 +53,11 @@ static int
...
 {
+    int audit_fd;
     va_list ap;
 
 #ifdef HAVE_LIBAUDIT
-    int audit_fd = audit_open();
+    audit_fd = audit_open();
 
     if (audit_fd >= 0) {
 	char *buf;
```
ldv-alt, line 63: No, this reverts #425.

ldv-alt (changes requested):

modules/pam_env/pam_env.8.xml
```diff
@@ -52,13 +52,40 @@
...
+      files <filename>/etc/security/environment</filename> and
+      <filename>/etc/security/environment.d/*</filename>.
+      If the file <filename>/etc/environment</filename> is not available the
+      settings are taken from the files <filename>/usr/etc/environment</filename>,
+      <filename>/usr/etc/environment.d/*</filename> and
+      <filename>/etc/environment.d/*</filename> in that order.
+      And last but not least with the <emphasis>readenv</emphasis> option this can
+      be completely disabled.
+    </para>
+    <para condition="without_vendordir">
```
ldv-alt, line 88: This creates an impression that all these new files are supported when linux-pam is configured using `--enable-vendordir`, but the implementation does something completely different: the only place where `VENDORDIR` is going to be used is a new function `_econf_read_file` which is naturally under `USE_ECONF` guard.

In other words, this PR is not about `--enable-vendordir` support in pam_env but rather about `--enable-econf` support in pam_env, and about new features supported only if linux-pam is configured using `--enable-econf`.

I don't see why `--enable-econf` should be a prerequisite for `--enable-vendordir`, it certainly hasn't been discussed.

ldv-alt (changes requested):

thkukuk:

ldv-alt:
Sorry, I intended to do the next review of this and other PRs, but putin's war in Ukraine disrupted everything.
Please bear with me, I'll have a look.

configure.ac
```diff
@@ -512,6 +512,11 @@ if test -n "$enable_vendordir"; then
   AC_DEFINE_UNQUOTED([VENDOR_SCONFIGDIR], ["$enable_vendordir/security"],
 		     [Directory for PAM modules distribution provided configuration files])
   STRINGPARAM_VENDORDIR="--stringparam vendordir '$enable_vendordir' --stringparam profile.condition 'with_vendordir'"
+  if test "$WITH_ECONF" = "yes" ; then
+    STRINGPARAM_VENDORDIR="--stringparam vendordir '$enable_vendordir' --stringparam profile.condition 'with_vendordir;with_vendordir_and_with_econf'"
+  else
+    STRINGPARAM_VENDORDIR="--stringparam vendordir '$enable_vendordir' --stringparam profile.condition 'with_vendordir;with_vendordir_and_without_econf"
+  fi
```
ldv-alt, line 519: The first definition of `STRINGPARAM_VENDORDIR` is redefined inside the added `if` statement.

modules/pam_env/pam_env.conf.5.xml
```diff
@@ -20,7 +20,15 @@
   <refsect1 id='pam_env.conf-description'>
     <title>DESCRIPTION</title>
 
-    <para>
+    <para condition="with_vendordir">
+      The <filename>/usr/etc/security/pam_env.conf</filename> and
+      <filename>/etc/security/pam_env.conf</filename> file specifies
+      the environment variables to be set, unset or modified by
+      <citerefentry><refentrytitle>pam_env</refentrytitle><manvolnum>8</manvolnum></citerefentry>.
+      When someone logs in, this file is read and the environment
```
ldv-alt, line 28: While one part of the statement talks about two files, another part talks about a single file, please check the grammar.

modules/pam_env/pam_env.conf.5.xml
```diff
@@ -61,7 +69,15 @@
       at front) can be used to mark this line as a comment line.
     </para>
 
-    <para>
+    <para condition="with_vendordir">
+      The <filename>/usr/etc/environment</filename> and <filename>/etc/environment</filename> file specifies
+      the environment variables to be set. The file must consist of simple
```
ldv-alt, line 74: Likewise, please check the grammar.

modules/pam_env/pam_env.c
```diff
@@ -53,7 +62,15 @@ typedef struct var {
 #define UNDEFINE_VAR 102
 #define ILLEGAL_VAR  103
 
+#ifdef USE_ECONF
+static int _isDirectory(const char *path);
+static char **_econf_read_file(const char *filename, const pam_handle_t *pamh,
+			       const char *delim, const char *name,
+			       const char *suffix, const char *subpath);
+#else
+static char **_read_file(const char*filename, const pam_handle_t *pamh);
```
ldv-alt, line 71: All identifiers starting with underscore is a reserved namespace.
The reason why we used them was static linking we used to support long time ago.
Nowadays this is just a relic from the past.

That is, you don't _have to_ introduce new identifiers from the reserved namespace.

modules/pam_env/pam_env.c
```diff
@@ -806,11 +993,8 @@ handle_env (pam_handle_t *pamh, int argc, const char **argv)
 
   retval = _parse_config_file(pamh, ctrl, conf_file);
 
-  if(readenv && retval == PAM_SUCCESS) {
+  if(readenv && retval == PAM_SUCCESS)
     retval = _parse_env_file(pamh, ctrl, env_file);
-    if (retval == PAM_IGNORE)
-      retval = PAM_SUCCESS;
```
ldv-alt, line 812: Why do you remove these two lines?  Is this a bug fix, or is it a change in behaviour?

ldv-alt (commented):
Stefan, when you are moving a lot of old code around along with new changes, a reviewer has to analyze the whole mix, which is an extra burden that I'd like to avoid. If you really want to do some refactoring, please separate these changes from those that change behaviour.

modules/pam_env/tst-pam_env-retval.c
```diff
@@ -0,0 +1,273 @@
...
+
+  setup();
+
+  /* PAM_SUCCESS -> PAM_SUCCESS, PAM_IGNORE -> PAM_PERM_DENIED */
+  ASSERT_NE(NULL, fp = fopen(service_file, "w"));
+  ASSERT_LT(0, fprintf(fp, "#%%PAM-1.0\n"
+		       "auth required %s/.libs/%s.so\n"
+		       "account required %s/.libs/%s.so\n"
+		       "password required %s/.libs/%s.so\n"
+		       "session required %s/.libs/%s.so\n",
```
ldv-alt, line 129: This is undefined behaviour: `pam_env` returns `PAM_SUCCESS` when environment variables are set.
Please have a look at [ldv/tst-pam_env-retval](https://github.com/linux-pam/linux-pam/commit/b53b9511274d2d98b2055a9b4ab7e635d3c9727a).

ldv-alt (changes requested):

modules/pam_env/pam_env.c
```diff
@@ -375,6 +353,215 @@ static int _assemble_line(FILE *f, char *buffer, int buf_len)
...
+      lines = realloc(lines, ++i * sizeof(char**));
+      if (lines == NULL) {
+	pam_syslog(pamh, LOG_ERR, "Cannot allocate memory.");
+	(void) fclose(conf);
+	return NULL;
+      }
+      lines[i-1] = strdup(buffer);
+    }
+    /* realloc one extra element for the last 0 */
+    lines = realloc(lines, ++i * sizeof(char**));
```
ldv-alt, line 381: Every memory allocation function can fail any time.

But why reading the whole file into memory in the first place?

modules/pam_env/pam_env.c
ldv-alt, line 71: In pam functions, `pamh` argument is usually the first one.

modules/pam_env/pam_env.c
```diff
@@ -375,6 +353,215 @@ static int _assemble_line(FILE *f, char *buffer, int buf_len)
...
+
+    size_t i = 0;
+    while (_assemble_line(conf, buffer, BUF_SIZE) > 0) {
+      D(("Read line: %s", buffer));
+      lines = realloc(lines, ++i * sizeof(char**));
+      if (lines == NULL) {
+	pam_syslog(pamh, LOG_ERR, "Cannot allocate memory.");
+	(void) fclose(conf);
+	return NULL;
+      }
```
ldv-alt, line 377: This is a memory leak on error path.

configure.ac
schubi2: Yes, fixed.

thkukuk:

ldv-alt:

ldv-alt (changes requested):
The main problem with this change is that you are moving big functions (`_parse_config_file` and `_parse_env_file`) and modifying them in the same commit, this makes reviewing changes impossible, please don't do that.

modules/pam_env/tst-pam_env-retval.c
```diff
@@ -17,11 +17,18 @@
...
 static const char service_file[] = TEST_NAME ".service";
 static const char missing_file[] = TEST_NAME ".missing";
-static const char my_conf[] = TEST_NAME ".conf";
-static const char my_env[] = TEST_NAME ".env";
+static const char dir[] = TEST_NAME_DIR;
+static const char dir_usr[] = TEST_NAME_DIR "/usr";
+static const char dir_usr_etc[] = TEST_NAME_DIR "/usr/etc";
+static const char dir_usr_etc_security[] = TEST_NAME_DIR "/usr/etc/security";
+static const char my_conf[] = TEST_NAME_DIR ".conf";
+static const char my_env[] = TEST_NAME_DIR ".env";
```
ldv-alt, line 29: I don't see why `my_conf` and `my_env` are changed here.

modules/pam_env/tst-pam_env-retval.c
```diff
@@ -41,13 +53,31 @@ setup(void)
...
+	ASSERT_NE(NULL, fp = fopen(usr_env, "w"));
+	ASSERT_LT(0, fprintf(fp,
+			     "var_etc_test=foo\n"
+			     "var_etc_test2=bar\n"));
+	ASSERT_EQ(0, fclose(fp));
+
+	ASSERT_NE(NULL, fp = fopen(usr_conf, "w"));
+	ASSERT_LT(0, fprintf(fp,
+			     "PAGER		DEFAULT=emacs\n"
+			     "MANPAGER	DEFAULT=less\n"));
```
ldv-alt, line 66: Please use `\t` instead of `	` in C strings.

modules/pam_env/tst-pam_env-retval.c
```diff
@@ -191,6 +221,36 @@ main(void)
 	const char *env2[] = { "test_value=foo", "test2_value=bar", NULL };
 	check_env(env2);
 
+#if defined (USE_ECONF)	&& defined (VENDORDIR)
+
+	/* envfile is a directory. So values will be read from /var/etc and /etc */
```
ldv-alt, line 226: From /var/etc?
Do we check that both directories are being used?

modules/pam_env/tst-pam_env-retval.c
```diff
@@ -191,6 +221,36 @@ main(void)
...
+			     " conffile=%s envfile=%s/%s/\n",
+			     cwd, MODULE_NAME,
+			     "/dev/null",
+			     cwd, dir));
+	ASSERT_EQ(0, fclose(fp));
+
+	const char *env3[] = {"var_etc_test=foo", "var_etc_test2=bar", NULL};
+	check_env(env3);
+
+	/* conffile is a directory. So values will be read from /var/etc and /etc */
```
ldv-alt, line 239: Likewise.

ldv-alt (changes requested):

modules/pam_env/pam_env.c
```diff
@@ -128,160 +157,132 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
-	*/
-
-	if (key[0] == '=') {
-		pam_syslog(pamh, LOG_ERR,
-		           "missing key name '%s' in %s', ignoring",
-		           key, file);
-		continue;
+    if (filename == NULL || strlen(base_dir) > 0) {
+      /* Read and merge all setting in e.g. /usr/etc and /etc */
+      char *vendor_dir, *sysconf_dir;
```
ldv-alt, line 209: Initializing `vendor_dir` to NULL would simplify the code below.

modules/pam_env/pam_env.c
```diff
@@ -128,160 +157,132 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
-       /*
-	* sanity check, the key must be alphanumeric
-	*/
-
-	if (key[0] == '=') {
-		pam_syslog(pamh, LOG_ERR,
-		           "missing key name '%s' in %s', ignoring",
-		           key, file);
-		continue;
+    if (filename == NULL || strlen(base_dir) > 0) {
```
ldv-alt, line 207: `base_dir[0] != '\0'` would be enough

modules/pam_env/pam_env.c
```diff
@@ -128,160 +157,132 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+      /* Read and merge all setting in e.g. /usr/etc and /etc */
+      char *vendor_dir, *sysconf_dir;
+      if (subpath != NULL && strlen(subpath) > 0) {
```
ldv-alt, line 210: `subpath[0] != '\0'` would be enough

modules/pam_env/pam_env.c
```diff
@@ -128,160 +157,132 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
-	    }
-	/* non-alphanumeric key, ignore this line */
-	if (key[i] != '=' && key[i] != '\0')
-	    continue;
+      D(("Read configuration from directory %s and %s", vendor_dir, sysconf_dir));
+      error = econf_readDirs (&key_file, vendor_dir, sysconf_dir, name, suffix,
+			      delim, "#");
+      free(vendor_dir);
+      free(sysconf_dir);
+      free(base_dir);
```
ldv-alt, line 243: Why do you need to allocate memory for `base_dir` in the first place?

modules/pam_env/pam_env.c
```diff
@@ -128,160 +157,132 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+      if (error != ECONF_SUCCESS) {
+	pam_syslog(pamh, LOG_ERR, "Unable to read configuration in different directories: %s",
+		   econf_errString(error));
+	return PAM_ABORT;
```
ldv-alt, line 247: I'm not sure `PAM_ABORT` is an appropriate return code here.

modules/pam_env/pam_env.c
```diff
@@ -128,160 +157,132 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+		 econf_errString(error));
+      return PAM_ABORT;
+    }
 
-	/* if this is a request to delete a variable, check that it's
-	   actually set first, so we don't get a vague error back from
-	   pam_putenv() */
-	for (i = 0; key[i] != '=' && key[i] != '\0'; i++);
+    *lines = malloc((key_number +1)* sizeof(char**));
+    (*lines)[key_number] = 0;
```
ldv-alt, line 259: Every memory allocation is allowed to fail.

modules/pam_env/pam_env.c
```diff
@@ -128,160 +157,132 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+      error = econf_getStringValue (key_file, NULL, keys[i], &val);
+      if (error != ECONF_SUCCESS) {
+	pam_syslog(pamh, LOG_ERR, "Unable to get string from key %s: %s",
+		   keys[i],
+		   econf_errString(error));
+      } else {
+        if (asprintf(&(*lines)[i],"%s%c%s", keys[i], delim[0], val) < 0) {
+	  pam_syslog(pamh, LOG_ERR, "Cannot allocate memory.");
+	  free_string_array(*lines);
+	  return PAM_BUF_ERR;
```
ldv-alt, line 273: This leaks both `keys` and `key_file`.

modules/pam_env/pam_env.c
```diff
@@ -375,6 +376,221 @@ static int _assemble_line(FILE *f, char *buffer, int buf_len)
...
+    D(("Parsed file name is: %s", filename));
+
+    if ((conf = fopen(filename,"r")) == NULL) {
+      pam_syslog(pamh, LOG_ERR, "Unable to open env file: %s", filename);
+      return PAM_IGNORE;
+    }
+
+    size_t i = 0;
+    *lines = malloc((i + 1)* sizeof(char**));
+    (*lines)[i] = 0;
```
ldv-alt, line 393: Every memory allocation is allowed to fail.

modules/pam_env/pam_env.c
```diff
@@ -375,6 +376,221 @@ static int _assemble_line(FILE *f, char *buffer, int buf_len)
     return used;
 }
 
+static int read_file(const pam_handle_t *pamh, const char*filename, char ***lines)
```
ldv-alt, line 379: Why do you want to read the whole file into memory?

ldv-alt (changes requested):

ldv-alt (changes requested):
Stefan, I asked you several times to avoid mixing code reordering with functional changes, but, unfortunately, you failed to do that.  Now I did the reordering, please rebase on top of the master branch.

schubi2:
> Stefan, I asked you several times to avoid mixing code reordering with functional changes, but, unfortunately, you failed to do that. Now I did the reordering, please rebase on top of the master branch.

Sorry for the additional work Dmitry. Somehow I have not really gotten your point. Now I see it. I have moved functions in the code and have changed them in one PR. That's really too much. Sorry again, next time I will ask earlier how to proceed. 
I have rebased the code and have not touched the position of the functions. I hope it is a little bit better for reading it. But the code could be still complex due the three code stream:
- as it is
- with vendor handling
- with libeconf (with and without vendor handling)

thkukuk:

ldv-alt:

modules/pam_env/pam_env.c
```diff
@@ -119,6 +138,154 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+    error = econf_getKeys(key_file, NULL, &key_number, &keys);
+    if (error != ECONF_SUCCESS && error != ECONF_NOKEY) {
+      pam_syslog(pamh, LOG_ERR, "Unable to read keys: %s",
+		 econf_errString(error));
+      econf_freeFile(key_file);
+      return PAM_ABORT;
+    }
+
+    *lines = malloc((key_number +1)* sizeof(char**));
+    if (lines == NULL) {
```
ldv-alt, line 253: If `*lines` has been assigned, then `lines` is definitely non-NULL, you probably meant to check `*lines` instead.

modules/pam_env/pam_env.c
```diff
@@ -212,6 +379,52 @@ _assemble_line(FILE *f, char *buffer, int buf_len)
...
+    D(("Parsed file name is: %s", filename));
+
+    if ((conf = fopen(filename,"r")) == NULL) {
+      pam_syslog(pamh, LOG_ERR, "Unable to open env file: %s", filename);
+      return PAM_IGNORE;
+    }
+
+    size_t i = 0;
+    *lines = malloc((i + 1)* sizeof(char**));
+    if (lines == NULL) {
```
ldv-alt, line 396: Likewise, if `*lines` has been assigned, then `lines` is definitely non-NULL, you probably meant to check `*lines` instead.

ldv-alt (changes requested):

modules/pam_env/pam_env.c
schubi2: Yes, you are right

modules/pam_env/pam_env.c
schubi2: Yes, you are right

modules/pam_env/pam_env.c
```diff
@@ -128,160 +157,132 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
-      return PAM_IGNORE;
+static int econf_read_file(const pam_handle_t *pamh, const char *filename, const char *delim,
+			   const char *name, const char *suffix, const char *subpath,
+			   char ***lines)
+{
+    econf_file *key_file = NULL;
+    econf_err error;
+    size_t key_number = 0;
+    char **keys = NULL;
+    char *base_dir = strdup("");
```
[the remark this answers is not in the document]
ldv-alt: I wonder why do you need to have a copy of a string in `base_dir`.
Why can't you just have
```
const char *base_dir = "";
...
base_dir = filename;
```
This way you won't have to deal with memory allocation issues.

ldv-alt (changes requested):

modules/pam_env/pam_env.c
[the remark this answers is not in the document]
schubi2: Yes, looks much more better now !

modules/pam_env/pam_env.c
```diff
@@ -119,6 +138,138 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+		   econf_errString(error));
+      } else {
+        if (asprintf(&(*lines)[i],"%s%c%s", keys[i], delim[0], val) < 0) {
+	  pam_syslog(pamh, LOG_ERR, "Cannot allocate memory.");
+          econf_free(keys);
+          econf_freeFile(key_file);
+	  free_string_array(*lines);
+	  return PAM_BUF_ERR;
+	}
+	free (val);
```
ldv-alt, line 262: Note that `val` is leaked here on error path.

ldv-alt (changes requested):

modules/pam_env/pam_env.c
schubi2: yes, good point !

ldv-alt (changes requested):

modules/pam_env/pam_env.8.xml
```diff
@@ -52,13 +52,55 @@
...
+      <filename>/usr/etc/security/pam_env.conf.d/*.conf</filename> and
+      <filename>/etc/security/pam_env.conf.d/*.conf</filename> in that order.
+    </para>
+    <para condition="with_vendordir_and_without_econf">
+      By default rules for (un)setting of variables are taken from the
+      config file <filename>/etc/security/pam_env.conf</filename>.
+      If this file does not exist <filename>%vendordir%/security/pam_env.conf</filename> is used.
+      An alternate file can be specified with the <emphasis>conffile</emphasis>
+      option, which overrules all other files.
+    </para>
```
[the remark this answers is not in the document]
schubi2: > In many places: grammar: path **to** this file
> 
> Please update the whole text from the latest editions you made for other modules.

Hm, the behaviour  differs form the other modules, so I cannot take over from them. But I have tried to improve it.

modules/pam_env/pam_env.8.xml
```diff
@@ -52,13 +52,55 @@
       variables as well as <emphasis>PAM_ITEM</emphasis>s such as
       <emphasis>PAM_RHOST</emphasis>.
     </para>
-    <para>
+    <para condition="with_vendordir_and_with_econf">
+      Rules for (un)setting of variables can be defined in an own conig
```
aplanas, line 56: Typo: conig -> config

schubi2: fixed. Thank you

ldv-alt:
Looks like the only remaining concern is that the new implementation unconditionally loads the whole file (or files) into memory.
I'm not sure it's a good idea, but I'd like to hear other opinions.

thkukuk:
> Looks like the only remaining concern is that the new implementation unconditionally loads the whole file (or files) into memory. I'm not sure it's a good idea, but I'd like to hear other opinions.

Which problems do you expect? I would assume we are speaking about some kb here, not even MB or GB...

ldv-alt:
While these files are expected to be small, we haven't relied on them to be small before, and with this change we would.
It's usually a good idea to make sure that something we rely upon really meets our expectations.

modules/pam_env/pam_env.c
```diff
@@ -682,19 +886,36 @@ static int
...
-    char buffer[BUF_SIZE], *key, *mark;
-    FILE *conf;
+    char *key, *mark;
+    char **env_list = NULL;
+
+#ifdef VENDOR_DEFAULT_ETC_ENVFILE
+#ifndef USE_ECONF
+    /* Only one file will be parsed. So, file has to be set. */
+    if (file == NULL) /* No filename has been set via argv. */
+      file = DEFAULT_ETC_ENVFILE;;
```
ldv-alt, line 896: Two semicolons in a row?

modules/pam_env/pam_env.c
```diff
@@ -625,34 +823,41 @@ static int
...
+#endif
 
-    /* _pam_assemble_line will provide a complete line from the config file,
-     * with all comments removed and any escaped newlines fixed up
-     */
+#ifdef USE_ECONF
+    /* If "file" is not NULL, only this file will be parsed. */
+    retval = econf_read_file(pamh, file, " \t", PAM_ENV, ".conf", "security", &conf_list);
+#else
+    retval = read_file(pamh, file, &conf_list);
```
ldv-alt, line 853: I think the `#ifdef VENDOR_DEFAULT_CONF_FILE` hunk could be moved right before the `read_file` invocation, this would make `#ifndef USE_ECONF` there unneeded.

modules/pam_env/pam_env.c
```diff
@@ -682,19 +886,36 @@ static int
...
+    /*
+    * Check whether DEFAULT_ETC_ENVFILE; file is available.
+    * If it does not exist, fall back to VENDOR_DEFAULT_ETC_ENVFILE; file.
+    */
+    struct stat stat_buffer;
+    if (stat(file, &stat_buffer) != 0 && errno == ENOENT) {
+      file = VENDOR_DEFAULT_ETC_ENVFILE;
+    }
+#endif
+#endif
```
ldv-alt, line 906: In `USE_ECONF` case this means that `file` can be `NULL`, and this is problematic because `file` is used later in this function for error diagnostics.

ldv-alt (dismissed):

modules/pam_env/pam_env.c
schubi2: Yes, good point

modules/pam_env/pam_env.c
schubi2: I am ensuring now that file is not NULL while reading the file via the "old style".

ldv-alt:
The problem with this workflow is that when requested changes take long time to implement, the reviewer no longer remembers what those changes were about, and has to do another review almost from scratch.

modules/pam_env/pam_env.c
```diff
@@ -119,6 +138,139 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+      }
+
+      D(("Read configuration from directory %s and %s", vendor_dir, sysconf_dir));
+      error = econf_readDirs (&key_file, vendor_dir, sysconf_dir, name, suffix,
+			      delim, "#");
+      free(vendor_dir);
+      free(sysconf_dir);
+      if (error != ECONF_SUCCESS) {
+	pam_syslog(pamh, LOG_ERR, "Unable to read configuration in different directories: %s",
+		   econf_errString(error));
```
thkukuk, line 220: For an user it is not clear which configuration file is not found: /etc/environment or any variant? Or pam_env.conf?
I think we should change this to something like:
`pam_syslog(pamh, LOG_ERR, "Unable to read configuration \"<configuration file>\": %s", econf_errString(error)); `or something like that.

thkukuk (changes requested):

modules/pam_env/pam_env.c
schubi2: I have used "econf_errLocation" in libeconf for this.

modules/pam_env/pam_env.c
```diff
@@ -119,6 +138,148 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+      }
+
+      D(("Read configuration from directory %s and %s", vendor_dir, sysconf_dir));
+      error = econf_readDirs (&key_file, vendor_dir, sysconf_dir, name, suffix,
+			      delim, "#");
+      free(vendor_dir);
+      free(sysconf_dir);
+      if (error != ECONF_SUCCESS) {
+        if (error == ECONF_NOFILE) {
+          pam_syslog(pamh, LOG_ERR, "No configuration file found");
```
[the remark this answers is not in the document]
schubi2: Yes, I have updated it.

modules/pam_env/pam_env.c
```diff
@@ -119,6 +138,148 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+      }
+
+      D(("Read configuration from directory %s and %s", vendor_dir, sysconf_dir));
+      error = econf_readDirs (&key_file, vendor_dir, sysconf_dir, name, suffix,
+			      delim, "#");
+      free(vendor_dir);
+      free(sysconf_dir);
+      if (error != ECONF_SUCCESS) {
+        if (error == ECONF_NOFILE) {
+	  pam_syslog(pamh, LOG_ERR, "Configuration file not found: %s.%s", name, suffix);
```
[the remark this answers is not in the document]
schubi2: Yes, I have reverted it.

thkukuk (approved):

configure.ac
```diff
@@ -523,6 +523,7 @@ if test "$WITH_ECONF" = "yes" ; then
   PKG_CHECK_MODULES([ECONF], [libeconf], [],
   [AC_CHECK_LIB([econf],[econf_readDirs],[ECONF_LIBS="-leconf"],[ECONF_LIBS=""])])
   if test -n "$ECONF_LIBS" ; then
+    AC_CHECK_LIB([econf],[econf_errLocation], [], [AC_MSG_ERROR([Please update libeconf])])
```
ldv-alt, line 526: This configure snippet requires that `libeconf` either doesn't exist or is fresh enough, which is somewhat unusual.
I suppose it would be surprising for many to see the build fail by default in case` libeconf` is installed but not fresh enough.

schubi2: I have changed it to check the version of libeconf only.

configure.ac
```diff
@@ -520,8 +520,7 @@ AC_ARG_ENABLE([econf],
   AS_HELP_STRING([--disable-econf], [do not use libeconf]),
   [WITH_ECONF=$enableval], WITH_ECONF=yes)
 if test "$WITH_ECONF" = "yes" ; then
-  PKG_CHECK_MODULES([ECONF], [libeconf], [],
-  [AC_CHECK_LIB([econf],[econf_readDirs],[ECONF_LIBS="-leconf"],[ECONF_LIBS=""])])
+  PKG_CHECK_MODULES([ECONF], [libeconf >= 0.5.0], [ECONF_LIBS="-leconf"], [ECONF_LIBS=""] )
```
ldv-alt, line 523: I wonder wouldn't a simpler variant be enough, e.g.
```
PKG_CHECK_MODULES([ECONF], [libeconf >= 0.5.0])
```

schubi2: Yes, that should be done in a more simple way. Fixed.
There is another PR which checks this too. I will fix it too.

schubi2: Hm, the "problem" is the buid PAM environment. It does not contain libeconf but configure will be called without the option --disable-econf. So it is is so "complicate" because libeconf will be disabled for the build if it is not there. So I have reverted it in order to get a running build again:
+++ b/configure.ac
@@ -520,9 +520,10 @@ AC_ARG_ENABLE([econf],
   AS_HELP_STRING([--disable-econf], [do not use libeconf]),
   [WITH_ECONF=$enableval], WITH_ECONF=yes)
 if test "$WITH_ECONF" = "yes" ; then
-  PKG_CHECK_MODULES([ECONF], [libeconf >= 0.5.0])
-  ECONF_LIBS="-leconf"
-  ECONF_CFLAGS="-DUSE_ECONF=1 $ECONF_CFLAGS"
+  PKG_CHECK_MODULES([ECONF], [libeconf >= 0.5.0], [ECONF_LIBS="-leconf"],[ECONF_LIBS=""])
+  if test -n "$ECONF_LIBS" ; then
+    ECONF_CFLAGS="-DUSE_ECONF=1 $ECONF_CFLAGS"
+  fi
 fi

Either we put libeconf in the PAM build environment, or we disabling libeconf via --disable-econf in the configure calls or we disable libeconf (almost) silently if it is not there (current state in the PR again).
So, what is would be prefered ?

thkukuk: > Sorry, I don't understand what was the problem with `PKG_CHECK_MODULES([ECONF], [libeconf >= 0.5.0])`.

With your suggestion, configure will fail if libeconf is not installed and you can only build Linux-PAM if you use --disable-econf.
Because this statement makes libeconf >= 0.5.0 a hard requirement.
With Stefans solution, the test will not fail, but libeconf support get's disabled if libeconf is not installed.

ldv-alt: I see.  I suppose this would work:
```
ECONF_CFLAGS=
ECONF_LIBS=
AC_ARG_ENABLE([econf],
  AS_HELP_STRING([--disable-econf], [do not use libeconf]),
  [WITH_ECONF=$enableval], [WITH_ECONF=yes])
if test "$WITH_ECONF" = "yes"; then
  PKG_CHECK_MODULES([ECONF], [libeconf >= 0.5.0], [], [])
  if test -n "$ECONF_LIBS"; then
    ECONF_CFLAGS="-DUSE_ECONF=1 $ECONF_CFLAGS"
  fi
fi
AC_SUBST([ECONF_CFLAGS])
AC_SUBST([ECONF_LIBS])
```

By the way, it should have been called --with-econf/--without-econf rather than --enable-econf/--disable-econf.

ldv-alt: Or even a shorter form:
```
ECONF_CFLAGS=
ECONF_LIBS=
AC_ARG_ENABLE([econf],
  AS_HELP_STRING([--disable-econf], [do not use libeconf]),
  [WITH_ECONF=$enableval], [WITH_ECONF=yes])
if test "$WITH_ECONF" = "yes"; then
  PKG_CHECK_MODULES([ECONF], [libeconf >= 0.5.0], [ECONF_CFLAGS="-DUSE_ECONF=1 $ECONF_CFLAGS"], [:])
fi
AC_SUBST([ECONF_CFLAGS])
AC_SUBST([ECONF_LIBS])
```

schubi2: Thank you !!!!

ldv-alt:
Stefan, in the commit message you mix phrases in the present tense (e.g. `Add libeconf setting`) and in the past tense (e.g. `Added tests for libeconf`), please follow the same style in all sentences, preferably the present tense.

ldv-alt (approved):

lnussel:
Is it really necessary to add `environment.d` to pam when there's `pam_env.conf.d`? The former is already used by systemd, see `environment.d(5)`

ldv-alt:
> Is it really necessary to add `environment.d` to pam when there's `pam_env.conf.d`? The former is already used by systemd, see `environment.d(5)`

`/etc/security/pam_env.conf` and `/etc/environment` are not to be confused, they have different format, see pam_env(8).

PR #409, opened 2021-12-03
pam_time: use vendor specific time.conf as fallback

schubi2:
Use the vendor directory as fallback for a distribution provided default config if there is no configuration in /etc.

* Makefile.am: compile testcases

* pam_time.8.xml: add description for vendor directory.

* pam_time.c: Take care about the fallback configuration in vendor directory.

* tst-pam_time-retval.c: General testcases for pam_time.

### DIFF
--- a/modules/pam_time/pam_time.c
+++ b/modules/pam_time/pam_time.c
@@ -34,6 +34,9 @@
 #endif
 
 #define PAM_TIME_CONF	(SCONFIGDIR "/time.conf")
+#ifdef VENDOR_SCONFIGDIR
+#define VENDOR_PAM_TIME_CONF (VENDOR_SCONFIGDIR "/time.conf")
+#endif
 
 #define PAM_TIME_BUFLEN        1000
 #define FIELD_SEPARATOR        ';'   /* this is new as of .02 */
@@ -78,6 +81,19 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv, const char **
 	    pam_syslog(pamh, LOG_ERR, "unknown option: %s", *argv);
 	}
     }
+
+#ifdef VENDOR_PAM_TIME_CONF
+    if (*conffile == PAM_TIME_CONF) {
+	/*
+	 * Check whether PAM_TIME_CONF file is available.
+	 * If it does not exist, fall back to VENDOR_PAM_TIME_CONF file.
+	 */
+	struct stat buffer;
+	if (stat(*conffile, &buffer) != 0 && errno == ENOENT) {
+	    *conffile = VENDOR_PAM_TIME_CONF;
+	}
+    }
+#endif
 
     return ctrl;
 }
--- a/modules/pam_time/Makefile.am
+++ b/modules/pam_time/Makefile.am
@@ -12,7 +12,7 @@ dist_man_MANS = time.conf.5 pam_time.8
 endif
 XMLS = README.xml time.conf.5.xml pam_time.8.xml
 dist_check_SCRIPTS = tst-pam_time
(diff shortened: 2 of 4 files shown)


ldv-alt:
Could you make sure it passes CI, please?

modules/pam_time/pam_time.c
```diff
@@ -54,6 +54,18 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv, const char **
     int ctrl = 0;
 
     *conffile = PAM_TIME_CONF;
+
+#ifdef VENDORDIR
+    /* If the specified file does not exist, try the vendor file as fallback. */
+    /* This can be overruled again by a user provided file definition (conffile=...) */
+    struct stat buffer;
+    if (stat(*conffile, &buffer) != 0) {
+      *conffile = VENDORDIR"/security/time.conf";
```
thkukuk, line 63: The order looks wrong. I think you want to change first conffile before checking that the file exists? Else the error message about a missing file will point to the wrong file.

thkukuk (changes requested):

modules/pam_time/pam_time.c
schubi2: Hm, I think it is correct. The check will be done on PAM_TIME_CONF which should be in this case /etc/security/time.conf. If this does not exists we are switching to /usr/etc/security/time.conf.
The error message should then be:
"cannot read settings from /etc/security/time.conf.. Trying /usr/etc/security/time.conf."
Or does I have overseen something ?

thkukuk: Seems I got confused as the next message (LOG_WARNING) sounded like an error message.
Since the fallback is wanted, I don't think we should log a warning, if at all an info message.

ldv-alt, line 63: Please check errno like other VENDORDIR fallbacks do.

ldv-alt:
Stefan, _please_ make sure your changes do not contain merge commits.
In other words, please _rebase_ on top of the master branch.

schubi2:
> Stefan, _please_ make sure your changes do not contain merge commits. In other words, please _rebase_ on top of the master branch.

OK, thanks !

modules/pam_time/pam_time.c
```diff
@@ -79,6 +82,19 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv, const char **
...
 
+#ifdef VENDOR_PAM_TIME_CONF
+    if (*conffile == PAM_TIME_CONF) {
+      /* Check whether PAM_TIME_CONF file is available.
+       * If it does not exist, fall back to VENDOR_PAM_TIME_CONF file. */
+      struct stat buffer;
+      if (stat(*conffile, &buffer) != 0 && errno == ENOENT) {
+	*conffile = VENDOR_PAM_TIME_CONF;
+	pam_syslog (pamh, LOG_INFO,
+		    "cannot read settings from %s. Trying %s", PAM_TIME_CONF, *conffile);
```
ldv-alt, line 93: Note that `pam_sm_acct_mgmt` already does some logging:
```
   ctrl = _pam_parse(pamh, argc, argv, &conf_file);

    if (ctrl & PAM_DEBUG_ARG) {
        pam_syslog(pamh, LOG_DEBUG, "conffile=%s", conf_file);
    }
```

Why do you suggest to log this one more time in `_pam_parse` with a higher level of importance?

ldv-alt (changes requested):

ldv-alt (approved):

thkukuk (approved):

PR #418, opened 2021-12-29
pam_limits: don't fail on missing config files

lnussel:
A config with only comments or an empty one is completely fine for
pam_limits. So don't complain about missing config files either.


thkukuk:
@ldv-alt @t8m Ok to merge? Fine with me.

t8m (approved):

ldv-alt (approved):

PR #419, opened 2021-12-31
pam_faillock: increase the limit of failed attempts before lockout

AnilRedshift:
Right now, any 3 failed passwords within a 15 minute window lock you out of the computer.
This seems very excessive, even for a something like a 4 digit pin password.

As a simple change to balance being a human with security, this proposed change bumps the default to 20 within the window

### DIFF
--- a/modules/pam_faillock/faillock.conf
+++ b/modules/pam_faillock/faillock.conf
@@ -28,8 +28,8 @@
 #
 # Deny access if the number of consecutive authentication failures
 # for this user during the recent interval exceeds n tries.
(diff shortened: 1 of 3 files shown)


ldv-alt:
If the default is ever going to change, the documentation would have to be updated, too.

AnilRedshift:
Thanks for taking a look. I've updated faillock.conf.5.xml, please let me know if there's anywhere else that needs updating

ldv-alt:
Sorry for the extra work, but this is not going to be changed.

PR #427, opened 2022-01-17
faillock: Hide unlocked users when showing locked users

BugBaron:
The pam_tally[2] program would only show users who had failed login
attempts. The equivalent faillock program also shows users who have
no failed login attempts.

Added a new option to the faillock program which will show users without
failed login attempts. By default this option is disabled, so the
command will not show those users.

### DIFF
--- a/modules/pam_faillock/main.c
+++ b/modules/pam_faillock/main.c
@@ -54,6 +54,7 @@
 #include "faillock.h"
 
 struct options {
+	unsigned int show_unlocked;
 	unsigned int reset;
 	const char *dir;
 	const char *user;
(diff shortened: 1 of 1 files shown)


ldv-alt:
Since the word _unlocked_ implies that something was locked but is no longer locked,
I'm not sure it is the right word to describe users with zero number of tally records.

BugBaron:
Would _users without failed logins_ be fine instead of _unlocked users_? Its a bit long, but its more correct.

ldv-alt:
A bit longer commit message is fine if it's more correct, but
there is the same issue with the name of new option --show-unlocked.

BugBaron:
Maybe --show-all-users would work, since the default functionality can be explained in the help message. Im struggling to come up with anything more succinct to use for the option.

PR #430, opened 2022-01-24
Introduce SCONFIGDIR and VENDOR_SCONFIGDIR macros

ldv-alt:
Dmitry V. Levin (6):
      Introduce SCONFIGDIR macro
      modules: use SCONFIGDIR macro
      pam_limits: make LIMITS_FILE_DIR macro consistent
      modules: move SCONFIGDIR-based macro definitions from Makefile.am to the source code
      Introduce VENDOR_SCONFIGDIR macro
      pam_limits: use VENDOR_SCONFIGDIR macro

Assuming that we will have all PAM modules updated to support VENDORDIR,
introduce two macros to make the code more consistent.

Similar to VENDORDIR, SCONFIGDIR is a new macro defined to the
argument of --enable-sconfigdir option.  Unlike --enable-vendordir,
--enable-sconfigdir has a default value, so when --enable-sconfigdir
is not used for build, SCONFIGDIR will be defined to that default value.

Use SCONFIGDIR macro instead of open-coding "/etc/security",
the latter is not correct when configured using --enable-sconfigdir
with an argument different from /etc/security.

VENDOR_SCONFIGDIR macro is a VENDORDIR version of SCONFIGDIR macro,
defined to VENDORDIR"/security" when --enable-vendordir is used for build.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -260,7 +260,7 @@ AC_ARG_ENABLE(sconfigdir,
 	AS_HELP_STRING([--enable-sconfigdir=DIR],[path to module conf files @<:@default=$sysconfdir/security@:>@]),
 	SCONFIGDIR=$enableval, SCONFIGDIR=$sysconfdir/security)
 AC_DEFINE_UNQUOTED([SCONFIGDIR], ["$SCONFIGDIR"],
-		   [Directory for system PAM modules configuration files])
+		   [Directory for PAM modules system configuration files])
 AC_SUBST(SCONFIGDIR)
 
 AC_ARG_ENABLE(pamlocking,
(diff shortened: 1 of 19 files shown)


ldv-alt:
@schubi2 Stefan, this might be of some interest to you, too.

thkukuk (approved):
I like it.

schubi2:
> I like it.

Me too :-)

ldv-alt:
Thorsten, you somehow squashed my 6 carefully crafted commits into a single one. :)
This is certainly not what I intended for this series, so I force-pushed my original 6 commits instead.
Sorry for the inconvenience.

PR #432, opened 2022-02-01
pam_keyinit: thread-safe implementation

ikerexxe:
modules/pam_keyinit/pam_keyinit.c: bypass setre*id() C library calls
with kernel calls and change global variables definitions to be
thread-safe.
configure.ac: macro definition for 32 bit system calls.

I also developed two test cases, but I decided to temporarily host them in [another repository](https://github.com/ikerexxe/test_pam_keyinit/blob/main/test_pam_keyinit.c). The tests would add new dependencies (cmoka, pamtest and pam_wrapper) and I think it's best if we agree on that first.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1997969

### DIFF
--- a/modules/pam_keyinit/pam_keyinit.c
+++ b/modules/pam_keyinit/pam_keyinit.c
@@ -21,6 +21,7 @@
 #include <security/pam_modutil.h>
 #include <security/pam_ext.h>
 #include <sys/syscall.h>
+#include <stdatomic.h>
 
 #define KEY_SPEC_SESSION_KEYRING	-3 /* ID for session keyring */
 #define KEY_SPEC_USER_KEYRING		-4 /* ID for UID-specific keyring */
@@ -31,12 +32,12 @@
 #define KEYCTL_REVOKE			3 /* revoke a key */
 #define KEYCTL_LINK			8 /* link a key into a keyring */
 
-static int my_session_keyring = 0;
-static int session_counter = 0;
-static int do_revoke = 0;
-static uid_t revoke_as_uid;
-static gid_t revoke_as_gid;
-static int xdebug = 0;
+static _Thread_local int my_session_keyring = 0;
+static _Atomic int session_counter = 0;
+static _Thread_local int do_revoke = 0;
+static _Thread_local uid_t revoke_as_uid;
+static _Thread_local gid_t revoke_as_gid;
+static _Thread_local int xdebug = 0;
 
 static void debug(pam_handle_t *pamh, const char *fmt, ...)
 	__attribute__((format(printf, 2, 3)));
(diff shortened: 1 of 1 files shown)


modules/pam_keyinit/pam_keyinit.c
```diff
@@ -140,14 +159,14 @@ static int kill_keyrings(pam_handle_t *pamh, int error_ret)
 
 		/* switch to the real UID and GID so that we have permission to
 		 * revoke the key */
-		if (revoke_as_gid != old_gid && setregid(-1, revoke_as_gid) < 0) {
+		if (revoke_as_gid != old_gid && pam_setregid(-1, revoke_as_gid) < 0) {
 			error(pamh, "Unable to change GID to %d temporarily\n", revoke_as_gid);
 			return error_ret;
 		}
 
 		if (revoke_as_uid != old_uid && setresuid(-1, revoke_as_uid, old_uid) < 0) {
```
ldv-alt, line 167: It's inconsistent to wrap setre[ug]id but leave setresuid as is.

ldv-alt (changes requested):

modules/pam_keyinit/pam_keyinit.c
ikerexxe: Thanks for pointing that out. I've corrected it.

configure.ac
```diff
@@ -533,6 +533,10 @@ AC_SUBST([CRYPTO_LIBS])
 AC_SUBST([STRINGPARAM_HMAC])
 AM_CONDITIONAL([COND_USE_OPENSSL], [test "x$use_openssl" = "xyes"])
 
+AC_CHECK_FUNCS([setreuid setregid],,
+  AC_CHECK_FUNCS([setreuid32 setregid32],
+    AC_DEFINE([HAVE_LINUX_32BIT_SYSCALLS], 1, [32 bit system calls provided])))
+
```
ldv-alt, line 539: Imagine that setreuid() and setregid() are provided by libc but implemented using appropriate *32 syscalls,
this is what glibc does on old 32-bit architectures.
I don't think the check for functions provided by libc is a reliable way to decide which syscalls to invoke.
Wouldn't simple ifdef's for SYS_setresuid32/SYS_setreuid32/SYS_setregid32 work better?

ldv-alt (changes requested):

ldv-alt (approved):

ikerexxe:
Before you merge this PR I'd like to discuss something. As mentioned in the first comment, I developed two test cases, but I decided to temporarily host them in [another repository](https://github.com/ikerexxe/test_pam_keyinit/blob/main/test_pam_keyinit.c). The tests would include new dependencies (cmoka, pamtest and pam_wrapper) to this project and that's why I didn't include those changes in this PR. What I'd like to know is if you'd be interested in having those tests upstream and enabling the aforementioned testing framework. If so, I'd open a new PR.

thkukuk (approved):

ldv-alt:
On Tue, Feb 08, 2022 at 12:22:05AM -0800, Iker Pedrosa wrote:
> Before you merge this PR I'd like to discuss something. As mentioned in the first comment, I developed two test cases, but I decided to temporarily host them in [another repository](https://github.com/ikerexxe/test_pam_keyinit/blob/main/test_pam_keyinit.c). The tests would include new dependencies (cmoka, pamtest and pam_wrapper) to this project and that's why I didn't include those changes in this PR. What I'd like to know is if you'd be interested in having those tests upstream and enabling the aforementioned testing framework. If so, I'd open a new PR.

I'm not familiar with these tools, but this pam_keyinit change requires
two different uids for testing, so these tools must be privileged, which
is scary.

Meanwhile, we have a dozen of modules/pam_*/tst-pam_*-retval.c tests
of pam modules.  All of them use the same simple testing framework
that doesn't need any extra dependencies.

ikerexxe:
> I'm not familiar with these tools, but this pam_keyinit change requires two different uids for testing, so these tools must be privileged, which is scary. Meanwhile, we have a dozen of modules/pam_*/tst-pam_*-retval.c tests of pam modules. All of them use the same simple testing framework that doesn't need any extra dependencies.

Even if I port the test to the existing infrastructure in `linux-pam` it would need privileges. I think I'll keep the test in its actual repository just for that.

Thanks for the reviews! I think the PR is ready for merging.

ikerexxe:
@ldv-alt can we get the PR merged?

PR #434, opened 2022-02-04
module/pam_access: use vendor specific access.conf as fallback

schubi2:
Use the vendor directory (/usr/etc) as fallback for a distribution provided default config if there is no configuration in /etc.

* pam_access.c: Take care about the fallback configuration in /usr/etc.

* pam_access.8.xml: added /usr/etc description


ldv-alt (changes requested):
Assuming that I understand the whole idea correctly, the following has to be implemented:
* if `accessfile=` option is specified, then no default `access.conf` or `access.d/*.conf` files are used;
* if `/etc/security/access.conf` doesn't exist, then `%vendordir%/security/access.conf` should not be used;
* if `etc/security/access.d/@filename@.conf` exists, then `%vendordir%/security/access.d/@filename@.conf` should not be used;
* All `access.d/*.conf` files are sorted by their `@filename@.conf` in lexicographic order regardless of which of the directories they reside in.

schubi2:
> Assuming that I understand the whole idea correctly, the following has to be implemented:
> 
> * if `accessfile=` option is specified, then no default `access.conf` or `access.d/*.conf` files are used;
> * if `/etc/security/access.conf` doesn't exist, then `%vendordir%/security/access.conf` should not be used;
> * if `etc/security/access.d/@filename@.conf` exists, then `%vendordir%/security/access.d/@filename@.conf` should not be used;
> * All `access.d/*.conf` files are sorted by their `@filename@.conf` in lexicographic order regardless of which of the directories they reside in.

I have adapted the code to these requirements. Currently without libeconfs until we have come to a decision how we are proceeding with the VENDORDIR and USE_ECONF definition.

thkukuk:

modules/pam_access/pam_access.c
```diff
@@ -34,6 +34,7 @@
 #include <stdarg.h>
 #include <syslog.h>
 #include <string.h>
+#include <libgen.h>
```
ldv-alt, line 37: Please don't include <libgen.h> and don't use functions declared there unless you really have to.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +159,73 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
     return 1;  /* OK */
 }
 
+/* --- evaluting all files in VENDORDIR/security/access.d and /etc/security/access.d --- */
+static int
+compare_filename(const void *a, const void *b)
+{
+	return strcmp(basename(* (char * const *) a), basename(* (char * const *) b));
```
ldv-alt, line 166: This function shall not modify data, so please don't use the XPG version of `basename`.
If you cannot use the GNU version of `basename`, don't use basename at all.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +159,73 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+	char **file_list = {NULL};
+	int file_list_size = glob_rv == 0 ? globbuf.gl_pathc : 0;
+
+#ifdef VENDOR_ACCESS_CONF_GLOB
+	glob_t globbuf_vendor;
+	int glob_rv_vendor = glob(VENDOR_ACCESS_CONF_GLOB, GLOB_ERR | GLOB_NOSORT, NULL, &globbuf_vendor);
+	if (glob_rv_vendor == 0)
+	    file_list_size += globbuf_vendor.gl_pathc;
+#endif
+	file_list = malloc((file_list_size + 1) * sizeof(char*));
```
ldv-alt, line 189: Memory allocation functions can fail any time.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +159,73 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+
+	if (!glob_rv) {
+	    for (i = 0; i < globbuf.gl_pathc; i++) {
+	        file_list[i] = strdup(globbuf.gl_pathv[i]);
+		if (file_list[i] == NULL)
+		    pam_syslog(pamh, LOG_ERR, "strdup failed: %m");
```
ldv-alt, line 195: This is not enough: if `file_list[i] == NULL`, then `compare_filename` will try to dereference NULL.

modules/pam_access/pam_access.c
```diff
@@ -919,28 +1002,18 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
...
+	    if (rv != NOMATCH)
+		break;
 	}
+	for (int i = 0; filename_list[i] != NULL; i++)
+	    free(filename_list[i]);
+	free(filename_list);
     }
 
-    if (loginfo.gai_rv == 0 && loginfo.res)
-	freeaddrinfo(loginfo.res);
```
ldv-alt, line 942: Why these lines are removed?  Is it a bug fix or a memory leak?

ldv-alt (changes requested):

thkukuk:

ldv-alt:

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,88 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
     return 1;  /* OK */
 }
 
+/* --- evaluting all files in VENDORDIR/security/access.d and /etc/security/access.d --- */
+static char *base_name(char *path)
+{
+    char *base = strrchr(path, '/');
+    return base ? base+1 : path;
+}
```
ldv-alt, line 166: Please change `char *` to `const char *` everywhere in this function.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,88 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+ * - If etc/security/access.d/@filename@.conf exists, then
+ *   %vendordir%/security/access.d/@filename@.conf should not be used.
+ * - All files in both access.d directories are sorted by their @filename@.conf in
+ *   lexicographic order regardless of which of the directories they reside in. */
+static char **read_access_dir(pam_handle_t *pamh)
+{
+	glob_t globbuf;
+	size_t i=0;
+	int glob_rv = glob(ACCESS_CONF_GLOB, GLOB_ERR | GLOB_NOSORT, NULL, &globbuf);
+	char **file_list = {NULL};
```
ldv-alt, line 185: Why `file_list` is initialized this way?

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,88 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+	if (glob_rv_vendor == 0)
+	    file_list_size += globbuf_vendor.gl_pathc;
+#endif
+	file_list = malloc((file_list_size + 1) * sizeof(char*));
+	if (file_list == NULL) {
+	    pam_syslog(pamh, LOG_ERR, "Cannot allocate memory for file list: %m");
+	    return NULL;
+	}
+
+	if (!glob_rv) {
```
ldv-alt, line 200: Please use `glob_rv == 0` instead of `!glob_rv`.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,88 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+	    for (i = 0; i < globbuf.gl_pathc; ) {
+	        file_list[i] = strdup(globbuf.gl_pathv[i]);
+		if (file_list[i] == NULL) {
+		    pam_syslog(pamh, LOG_ERR, "strdup failed: %m");
+		    break;
+		}
+		i++;
```
ldv-alt, line 207: Please move `i++` into the `for` statement.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,88 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+	    }
+	}
+#ifdef VENDOR_ACCESS_CONF_GLOB
+	if (!glob_rv_vendor) {
```
ldv-alt, line 211: Please use `glob_rv_vendor == 0` instead of `!glob_rv_vendor`.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,88 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+	    for (size_t j = 0; j < globbuf_vendor.gl_pathc; j++) {
+		if (globbuf.gl_pathc > 0) {
```
ldv-alt, line 213: `globbuf` can be used only if `glob_rv == 0`.

modules/pam_access/pam_access.c
```diff
@@ -976,22 +1073,17 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
...
-	/* We do not manipulate locale as setlocale() is not
-	 * thread safe. We could use uselocale() in future.
-	 */
-	glob_rv = glob(ACCESS_CONF_GLOB, GLOB_ERR, NULL, &globbuf);
-	if (!glob_rv) {
-	    /* Parse the *.conf files. */
-	    for (i = 0; globbuf.gl_pathv[i] != NULL; i++) {
-		loginfo.config_file = globbuf.gl_pathv[i];
+        char **filename_list = read_access_dir(pamh);
+	if (filename_list != NULL) {
```
ldv-alt, line 1077: Something went wrong with the indentation here.

ldv-alt (changes requested):

modules/pam_access/pam_access.c
schubi2: Where do you see the issue ? 
read_access_dir can return NULL

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,87 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+static const char *base_name(const char *path)
+{
+    const char *base = strrchr(path, '/');
+    return base ? base+1 : path;
+}
+
+static int
+compare_filename(const void *a, const void *b)
+{
+	return strcmp(base_name(* (char * const *) a), base_name(* (char * const *) b));
```
ldv-alt, line 171: Please cast to `const char **` instead of `char * const *`.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,87 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+#ifdef VENDOR_ACCESS_CONF_GLOB
+	glob_t globbuf_vendor;
+	int glob_rv_vendor = glob(VENDOR_ACCESS_CONF_GLOB, GLOB_ERR | GLOB_NOSORT, NULL, &globbuf_vendor);
+	if (glob_rv_vendor == 0)
+	    file_list_size += globbuf_vendor.gl_pathc;
+#endif
+	file_list = malloc((file_list_size + 1) * sizeof(char*));
+	if (file_list == NULL) {
+	    pam_syslog(pamh, LOG_ERR, "Cannot allocate memory for file list: %m");
+	    return NULL;
```
ldv-alt, line 197: Note that `globbuf` and `globbuf_vendor` are potentially leaked on this error path.

ldv-alt (changes requested):

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,91 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+	if (glob_rv_vendor == 0)
+	    file_list_size += globbuf_vendor.gl_pathc;
+#endif
+	file_list = malloc((file_list_size + 1) * sizeof(char*));
+	if (file_list == NULL) {
+	    pam_syslog(pamh, LOG_ERR, "Cannot allocate memory for file list: %m");
+#ifdef VENDOR_ACCESS_CONF_GLOB
+	    globfree(&globbuf_vendor);
+#endif
+	    globfree(&globbuf);
```
ldv-alt, line 200: `globfree` should be called only if the corresponding `glob` was successful.

ldv-alt (changes requested):

modules/pam_access/pam_access.c
ldv-alt: Yes.

ldv-alt:

thkukuk:

ldv-alt:

modules/pam_access/pam_access.8.xml
```diff
@@ -66,6 +66,26 @@
...
     </para>
+    <para condition="with_vendordir">
+      By default rules for access management are taken from config file
+      <filename>/etc/security/access.conf</filename> or, if that one is not
+      present, the file <filename>%vendordir%/security/access.conf</filename>.
+      These settings can be overruled by setting in a config file explicitly
+      specified with the <option>accessfile</option> option.
+      Then individual <filename>*.conf</filename> files from the
+      <filename>/etc/security/access.d/</filename> and
+      <filename>%vendordir%/security/access.d</filename> directory are read.
```
ldv-alt, line 77: Since you are talking about two directories here, I suppose `directories are read` would be more correct.

modules/pam_access/pam_access.8.xml
```diff
@@ -66,6 +66,26 @@
...
+      If <filename>/etc/security/access.d/@filename@.conf</filename> exists, then
+      <filename>%vendordir%/security/access.d/@filename@.conf</filename> will not be used.
+      All <filename>access.d/*.conf</filename> files are sorted by their
+      <filename>@filename@.conf</filename> in lexicographic order regardless of which
+      of the directories they reside in.
+      The effect of the individual files is the same as if all the files were
+      concatenated together in the order of parsing. This means that once
+      a pattern is matched in some file no further files are parsed.
+      If a config file is explicitly specified with the <option>accessfile</option>
+      option the files in the above directory are not parsed.
```
ldv-alt, line 87: Likewise, since you are talking about two directories here, I suppose `directories are not parsed` would be more correct.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,95 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+static const char *base_name(const char *path)
+{
+    const char *base = strrchr(path, '/');
+    return base ? base+1 : path;
+}
+
+static int
+compare_filename(const void *a, const void *b)
+{
+	return strcmp(base_name(* (const char * const *) a), base_name(* (const char * const *) b));
```
ldv-alt, line 171: This line is too long, please wrap, e.g.
```
```

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,95 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+ *   %vendordir%/security/access.d/@filename@.conf should not be used.
+ * - All files in both access.d directories are sorted by their @filename@.conf in
+ *   lexicographic order regardless of which of the directories they reside in. */
+static char **read_access_dir(pam_handle_t *pamh)
+{
+	glob_t globbuf;
+	size_t i=0;
+	int glob_rv = glob(ACCESS_CONF_GLOB, GLOB_ERR | GLOB_NOSORT, NULL, &globbuf);
+	char **file_list;
+	int file_list_size = glob_rv == 0 ? globbuf.gl_pathc : 0;
```
ldv-alt, line 186: `file_list_size` should be of the same type as `globbuf.gl_pathc`, which is `size_t`.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,95 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+	    }
+	}
+#ifdef VENDOR_ACCESS_CONF_GLOB
+	if (glob_rv_vendor == 0) {
+	    for (size_t j = 0; j < globbuf_vendor.gl_pathc; j++) {
+		if (glob_rv == 0 && globbuf.gl_pathc > 0) {
+		    int double_found = 0;
+		    for (size_t k = 0; k < globbuf.gl_pathc; k++) {
+		        if (strcmp(basename(globbuf.gl_pathv[k]),
+				   basename(globbuf_vendor.gl_pathv[j])) == 0) {
```
ldv-alt, line 222: Please consider using `base_name` instead of `basename` here as well.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,95 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+		    break;
+		}
+		i++;
+	    }
+	    globfree(&globbuf_vendor);
+	}
+#endif
+	file_list[i] = NULL;
+	if (i>0)
+	    qsort(file_list, i, sizeof(char *), compare_filename);
```
ldv-alt, line 242: The `i>0` check is redundant.

modules/pam_access/pam_access.c
```diff
@@ -154,6 +158,95 @@ parse_args(pam_handle_t *pamh, struct login_info *loginfo,
...
+
+        if (glob_rv == 0)
+            globfree(&globbuf);
```
ldv-alt, line 245: Please indent these lines the same way as the neighboring lines.

ldv-alt (changes requested):
Looks almost ready, please make these minor fixes.

ldv-alt:
Also, in the commit message, it should be just `pam_access`, not `module/pam_access`. By the way the directory where pam_access is located is called `modules`.

In the changelog part of the commit message, there is no need to insert an empty line between `* pam_access.c` and `* pam_access.8.xml`.

modules/pam_access/pam_access.c
schubi2: Hopefully I have hit the right one.

schubi2:
I hope, I have done all requested changes.

ldv-alt (approved):

PR #435, opened 2022-02-08
pam_limits: use vendor specific content in limits.d directory as fallback

schubi2:
Use the vendor directory as fallback for a distribution provided default config if there is no configuration in /etc.

- pam_limits.c: Take care about the fallback configuration in vendor directory.
- pam_limits.8.xml: Added description for vendor directory.

### DIFF
--- a/modules/pam_limits/pam_limits.8.xml
+++ b/modules/pam_limits/pam_limits.8.xml
@@ -48,7 +48,7 @@
       obtained in a user-session. Users of <emphasis>uid=0</emphasis> are affected
       by this limits, too.
     </para>
-    <para>
+    <para condition="without_vendordir">
       By default limits are taken from the <filename>/etc/security/limits.conf</filename>
       config file. Then individual *.conf files from the <filename>/etc/security/limits.d/</filename>
       directory are read. The files are parsed one after another in the order of "C" locale.
@@ -58,9 +58,21 @@
       files in the above directory are not parsed.
     </para>
     <para condition="with_vendordir">
-      If there is no explicitly specified configuration file and
-      <filename>/etc/security/limits.conf</filename> does not exist,
-      <filename>%vendordir%/security/limits.conf</filename> is used.
+      By default limits are taken from the <filename>/etc/security/limits.conf</filename>
+      config file or, if that one is not present, the file
+      <filename>%vendordir%/security/limits.conf</filename>.
+      Then individual <filename>*.conf</filename> files from the
+      <filename>/etc/security/limits.d/</filename> and
+      <filename>%vendordir%/security/limits.d</filename> directories are read.
+      If <filename>/etc/security/limits.d/@filename@.conf</filename> exists, then
+      <filename>%vendordir%/security/limits.d/@filename@.conf</filename> will not be used.
+      All <filename>limits.d/*.conf</filename> files are sorted by their
+      <filename>@filename@.conf</filename> in lexicographic order regardless of which
+      of the directories they reside in.
+      The effect of the individual files is the same as if all the files were
+      concatenated together in the order of parsing.
+      If a config file is explicitly specified with the <option>config</option>
+      option the files in the above directories are not parsed.
     </para>
     <para>
       The module must not be called by a multithreaded application.
@@ -214,6 +226,13 @@
         <term><filename>/etc/security/limits.conf</filename></term>
         <listitem>
           <para>Default configuration file</para>
+        </listitem>
+      </varlistentry>
+      <varlistentry condition="with_vendordir">
+        <term><filename>%vendordir%/security/limits.conf</filename></term>
+        <listitem>
+          <para>Default configuration file if
+	  <filename>/etc/security/limits.conf</filename> does not exist.</para>
         </listitem>
       </varlistentry>
     </variablelist>
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -126,7 +126,11 @@ struct pam_limit_s {
 #define LIMITS_CONF_GLOB	(LIMITS_FILE_DIR "/*.conf")
 
 #define LIMITS_FILE	(SCONFIGDIR "/limits.conf")
-#define CONF_FILE	((pl->conf_file != NULL) ? pl->conf_file : LIMITS_FILE)
+
+#ifdef VENDOR_SCONFIGDIR
+#define VENDOR_LIMITS_FILE (VENDOR_SCONFIGDIR "/limits.conf")
+#define VENDOR_LIMITS_CONF_GLOB  (VENDOR_SCONFIGDIR "/limits.d/*.conf")
+#endif
 
 static int
 _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
@@ -811,35 +815,23 @@ parse_uid_range(pam_handle_t *pamh, const char *domain,
 
 static int
 parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
-			     int ctrl, struct pam_limit_s *pl)
+		  int ctrl, struct pam_limit_s *pl, const int conf_file_set_by_user)
 {
     FILE *fil;
     char buf[LINE_LENGTH];
 
-    /* check for the CONF_FILE */
+    /* check for the conf_file */
     if (ctrl & PAM_DEBUG_ARG)
-        pam_syslog(pamh, LOG_DEBUG, "reading settings from '%s'", CONF_FILE);
-    fil = fopen(CONF_FILE, "r");
+        pam_syslog(pamh, LOG_DEBUG, "reading settings from '%s'", pl->conf_file);
+    fil = fopen(pl->conf_file, "r");
     if (fil == NULL) {
-      int err = errno;
+        if (errno == ENOENT && !conf_file_set_by_user)
+            return PAM_SUCCESS; /* file is not there and it has not been set by the conf= argument */
 
-#ifdef VENDOR_SCONFIGDIR
-      /* if the specified file does not exist, and it is not provided by
-         the user, try the vendor file as fallback. */
-      if (pl->conf_file == NULL && err == ENOENT)
-        fil = fopen(VENDOR_SCONFIGDIR "/limits.conf", "r");
-
-      if (fil == NULL)
-#endif
-        {
-          if (err == ENOENT)
-            return PAM_SUCCESS;
-
-          pam_syslog (pamh, LOG_WARNING,
-                      "cannot read settings from %s: %s", CONF_FILE,
-                      strerror(err));
-          return PAM_SERVICE_ERR;
-        }
+        pam_syslog(pamh, LOG_WARNING,
+                   "cannot read settings from %s: %s", pl->conf_file,
+                   strerror(errno));
+        return PAM_SERVICE_ERR;
     }
 
     /* start the show */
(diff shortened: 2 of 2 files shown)


ldv-alt (dismissed):
Assuming that I understand the whole idea correctly, the following has to be implemented:
* if `conf=` option is specified, then no default `limits.conf` or `limits.d/*.conf` files are used;
* if `/etc/security/limits.conf` doesn't exist, then `%vendordir%/security/limits.conf` should not be used;
* if `etc/security/limits.d/@filename@.conf` exists, then `%vendordir%/security/limits.d/@filename@.conf` should not be used;
* all `limits.d/*.conf` files are sorted by their `@filename@.conf` in lexicographic order regardless of which of the directories they reside in.

thkukuk:
@ldv-alt about your requests:

>     * if `conf=` option is specified, then no default `limits.conf` or `limits.d/*.conf` files are used;

While I agree to this from a logical point of view: this would break 15 year old behavior. Couldn't this create unexpected behavior? Or can we safely assume, that nobody is using conf= with limits.d/  during the last 15 years?
 
>     * if `/etc/security/limits.conf` doesn't exist, then `%vendordir%/security/limits.conf` should not be used;

I don't understand, if  `/etc/security/limits.conf` does not exisst then `%vendordir%/security/limits.conf`  is the fallback and should be used. Why shouldn't it be used?
 
>     * if `etc/security/limits.d/@filename@.conf` exists, then `%vendordir%/security/limits.d/@filename@.conf` should not be used;

I fully agree.

>     * all `limits.d/*.conf` files are sorted by their `@filename@.conf` in lexicographic order regardless of which of the directories they reside in.

Again, this would break 15 year old behavior, but I agree that this is a good idea.

This all sounds like a complete rewrite of pam_limits.
I'm not sure if we can enhance libeconf to read the limits.conf format, but that would simplify this a lot, as libeconf is designed exactly for this case.
@schubi2 do you see a chance to handle limits.conf with the libeconf interface? Or can we enhance it accordingly? If yes, then we could even keep pam_limits as it is without breaking anything if compiled without libeconf support, else delegate this all to the libeconf library.

ldv-alt:
On Thu, Feb 10, 2022 at 05:18:05AM -0800, Thorsten Kukuk wrote:
> @ldv-alt about your requests:
> 
> >     * if `conf=` option is specified, then no default `limits.conf` or `limits.d/*.conf` files are used;
> 
> While I agree to this from a logical point of view: this would break 15 year old behavior. Couldn't this create unexpected behavior? Or can we safely assume, that nobody is using conf= with limits.d/  during the last 15 years?

The manual page state that "If a config file is explicitly specified with
a module option then the files in the above directory are not parsed."

> >     * if `/etc/security/limits.conf` doesn't exist, then `%vendordir%/security/limits.conf` should not be used;
> 
> I don't understand, if  `/etc/security/limits.conf` does not exisst then `%vendordir%/security/limits.conf`  is the fallback and should be used. Why shouldn't it be used?

Sorry, it was a typo, of course I meant that if /etc/security/limits.conf
exists, then %vendordir%/security/limits.conf should not be used.

schubi2:
> On Thu, Feb 10, 2022 at 05:18:05AM -0800, Thorsten Kukuk wrote: @ldv-alt about your requests: > * if `conf=` option is specified, then no default `limits.conf` or `limits.d/*.conf` files are used; While I agree to this from a logical point of view: this would break 15 year old behavior. Couldn't this create unexpected behavior? Or can we safely assume, that nobody is using conf= with limits.d/ during the last 15 years?
> The manual page state that "If a config file is explicitly specified with a module option then the files in the above directory are not parsed."

Yes, the current code and this PR are taking care about it.

schubi2:
> This all sounds like a complete rewrite of pam_limits. I'm not sure if we can enhance libeconf to read the limits.conf format, but that would simplify this a lot, as libeconf is designed exactly for this case. @schubi2 do you see a chance to handle limits.conf with the libeconf interface? Or can we enhance it accordingly? If yes, then we could even keep pam_limits as it is without breaking anything if compiled without libeconf support, else delegate this all to the libeconf library.

I will have a look on it. I am currently struggling a little bit over the definitions in PAM. Dimitry has already pointed this in my pam_env PR where I am also not so really happy with it:

We have two defines: VENDORDIR and USE_ECONF. In the pam_env PR (  https://github.com/linux-pam/linux-pam/pull/408/files) they are quite independent now. But this makes the code quite complex because we have three scenarios: without_vendordir; with_vendordir_without_econf ; with_vendordir_with_econf. 
If econf can be used for every stuff which handles <module_name>.d directories I would suggest that we drop the mode with_vendordir_without_econf. So if VENDORDIR is set we are requiring and using econf only.

What do you think about it ?

schubi2:
> > * if `/etc/security/limits.conf` doesn't exist, then `%vendordir%/security/limits.conf` should not be used; I don't understand, if `/etc/security/limits.conf` does not exisst then `%vendordir%/security/limits.conf` is the fallback and should be used. Why shouldn't it be used?
> Sorry, it was a typo, of course I meant that if /etc/security/limits.conf exists, then %vendordir%/security/limits.conf should not be used.

This will be supported by the PR.

schubi2:
> Assuming that I understand the whole idea correctly, the following has to be implemented:
> 
> * if `conf=` option is specified, then no default `limits.conf` or `limits.d/*.conf` files are used;
> * if `/etc/security/limits.conf` doesn't exist, then `%vendordir%/security/limits.conf` should be used;
> * if `etc/security/limits.d/@filename@.conf` exists, then `%vendordir%/security/limits.d/@filename@.conf` should not be used;
> * all `limits.d/*.conf` files are sorted by their `@filename@.conf` in lexicographic order regardless of which of the directories they reside in.

I have adapted the code to these requirements. Currently without libeconfs until we have come to a decision how we are proceeding with the VENDORDIR and USE_ECONF definition.

thkukuk:

modules/pam_limits/pam_limits.c
```diff
@@ -1139,44 +1212,37 @@ pam_sm_open_session (pam_handle_t *pamh, int flags UNUSED,
...
+    for (i = 0; filename_list[i] != NULL; i++) {
+	pl->conf_file = filename_list[i];
+	retval = parse_config_file(pamh, pwd->pw_name, pwd->pw_uid, pwd->pw_gid, ctrl, pl);
+	if (retval == PAM_IGNORE) {
+	   D(("the configuration file ('%s') has an applicable '<domain> -' entry", pl->conf_file));
+	   for (int j = 0; filename_list[j] != NULL; j++)
+	       free(filename_list[j]);
+	   free(filename_list);
+	   return PAM_SUCCESS;
+	}
```
ldv-alt, line 1233: Please move the `PAM_IGNORE` check out of the loop, after `filename_list` cleanup.

ldv-alt (changes requested):

thkukuk:

ldv-alt:

modules/pam_limits/pam_limits.c
```diff
@@ -1095,33 +1087,128 @@ static int setup_limits(pam_handle_t *pamh,
...
+#ifdef VENDOR_LIMITS_FILE
+    const char *default_config = LIMITS_FILE;
+    if (pl->conf_file == NULL) {
+        /* Check whether LIMITS_FILE file is available.
+         * If it does not exist, fall back to VENDOR_LIMITS_FILE file. */
+        struct stat buffer;
+        pl->conf_file = LIMITS_FILE;
+        if (stat(pl->conf_file, &buffer) != 0 && errno == ENOENT) {
+            default_config = VENDOR_LIMITS_FILE;
+	    pl->conf_file = default_config;
```
ldv-alt, line 1205: Something went wrong with the indentation here.

modules/pam_limits/pam_limits.c
```diff
@@ -1139,44 +1226,35 @@ pam_sm_open_session (pam_handle_t *pamh, int flags UNUSED,
...
 	return PAM_SUCCESS;
     }
-    if (retval != PAM_SUCCESS || pl->conf_file != NULL)
+    if (retval != PAM_SUCCESS || conf_file_set_by_user)
 	/* skip reading limits.d if config file explicitly specified */
 	goto out;
 
     /* Read subsequent *.conf files, if they exist. */
+    char **filename_list = read_limits_dir(pamh);
+    for (i = 0; filename_list[i] != NULL; i++) {
```
ldv-alt, line 1238: Note that `read_limits_dir` can return NULL.

modules/pam_limits/pam_limits.c
```diff
@@ -816,30 +820,18 @@ parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
...
-         the user, try the vendor file as fallback. */
-      if (pl->conf_file == NULL && err == ENOENT)
-        fil = fopen(VENDOR_SCONFIGDIR "/limits.conf", "r");
-
-      if (fil == NULL)
-#endif
-        {
-          if (err == ENOENT)
+        if (errno == ENOENT)
             return PAM_SUCCESS;
```
ldv-alt, line 829: This would treat ENOENT as PAM_SUCCESS not just for default limits.conf, but also for config files specified using `conf=` options.  This is a change in behaviour.

ldv-alt (changes requested):

ldv-alt:

thkukuk:

ldv-alt:

modules/pam_limits/pam_limits.8.xml
```diff
@@ -58,9 +58,21 @@
...
     <para condition="with_vendordir">
-      If there is no explicitly specified configuration file and
-      <filename>/etc/security/limits.conf</filename> does not exist,
-      <filename>%vendordir%/security/limits.conf</filename> is used.
+      By default limits are taken from the <filename>/etc/security/limits.conf</filename>
+      config file or, if that one is not present, the file
+      <filename>%vendordir%/security/limits.conf</filename>.
+      Then individual <filename>*.conf</filename> files from the
+      <filename>/etc/security/limits.d/</filename> and
+      <filename>%vendordir%/security/limits.d</filename> directory are read.
```
ldv-alt, line 66: Since you are talking about two directories here, I suppose `directories are read` would be more correct.

modules/pam_limits/pam_limits.8.xml
```diff
@@ -58,9 +58,21 @@
...
+      If <filename>/etc/security/limits.d/@filename@.conf</filename> exists, then
+      <filename>%vendordir%/security/limits.d/@filename@.conf</filename> will not be used.
+      All <filename>limits.d/*.conf</filename> files are sorted by their
+      <filename>@filename@.conf</filename> in lexicographic order regardless of which
+      of the directories they reside in.
+      The effect of the individual files is the same as if all the files were
+      concatenated together in the order of parsing.
+      If a config file is explicitly specified with the <option>config</option>
+      option the files in the above directory are not parsed.
```
ldv-alt, line 75: Likewise, since you are talking about two directories here, I suppose `directories are not parsed` would be more correct.

modules/pam_limits/pam_limits.c
```diff
@@ -1095,33 +1087,134 @@ static int setup_limits(pam_handle_t *pamh,
...
+static const char *base_name(const char *path)
+{
+    const char *base = strrchr(path, '/');
+    return base ? base+1 : path;
+}
+
+static int
+compare_filename(const void *a, const void *b)
+{
+	return strcmp(base_name(* (const char * const *) a), base_name(* (const char * const *) b));
```
ldv-alt, line 1100: This line is too long, please wrap, e.g.
```
```

modules/pam_limits/pam_limits.c
```diff
@@ -1095,33 +1087,134 @@ static int setup_limits(pam_handle_t *pamh,
...
+ *   %vendordir%/security/limits.d/@filename@.conf should not be used.
+ * - All files in both limits.d directories are sorted by their @filename@.conf in
+ *   lexicographic order regardless of which of the directories they reside in. */
+static char **read_limits_dir(pam_handle_t *pamh)
+{
+	glob_t globbuf;
+	size_t i=0;
+	int glob_rv = glob(LIMITS_CONF_GLOB, GLOB_ERR | GLOB_NOSORT, NULL, &globbuf);
+	char **file_list;
+	int file_list_size = glob_rv == 0 ? globbuf.gl_pathc : 0;
```
ldv-alt, line 1115: `file_list_size` should be of the same type as `globbuf.gl_pathc`, which is `size_t`.

modules/pam_limits/pam_limits.c
```diff
@@ -1095,33 +1087,134 @@ static int setup_limits(pam_handle_t *pamh,
...
+	    }
+	}
+#ifdef VENDOR_LIMITS_CONF_GLOB
+	if (glob_rv_vendor == 0) {
+	    for (size_t j = 0; j < globbuf_vendor.gl_pathc; j++) {
+		if (glob_rv == 0 && globbuf.gl_pathc > 0) {
+		    int double_found = 0;
+		    for (size_t k = 0; k < globbuf.gl_pathc; k++) {
+			if (strcmp(basename(globbuf.gl_pathv[k]),
+				   basename(globbuf_vendor.gl_pathv[j])) == 0) {
```
ldv-alt, line 1151: Please consider using `base_name` instead of `basename` here as well.

modules/pam_limits/pam_limits.c
```diff
@@ -1095,33 +1087,134 @@ static int setup_limits(pam_handle_t *pamh,
...
+		    break;
+		}
+		i++;
+	    }
+	    globfree(&globbuf_vendor);
+	}
+#endif
+	file_list[i] = NULL;
+	if (i>0)
+	    qsort(file_list, i, sizeof(char *), compare_filename);
```
ldv-alt, line 1171: The `i>0` check is redundant.

modules/pam_limits/pam_limits.c
```diff
@@ -1095,33 +1087,134 @@ static int setup_limits(pam_handle_t *pamh,
...
+        if (glob_rv == 0)
+            globfree(&globbuf);
```
ldv-alt, line 1173: Please indent these lines the same way as the neighboring lines.

modules/pam_limits/pam_limits.c
```diff
@@ -1095,33 +1087,134 @@ static int setup_limits(pam_handle_t *pamh,
...
+        pl->conf_file = LIMITS_FILE;
+        if (stat(pl->conf_file, &buffer) != 0 && errno == ENOENT) {
+            default_config = VENDOR_LIMITS_FILE;
+            pl->conf_file = default_config;
+        }
+    }
+#else
+    if (pl->conf_file == NULL)
+       pl->conf_file = LIMITS_FILE;
+#endif
```
ldv-alt, line 1217:
```
    if (pl->conf_file == NULL) {
        pl->conf_file = LIMITS_FILE;
#ifdef VENDOR_LIMITS_FILE
        /*
         * Check whether LIMITS_FILE file is available.
         * If it does not exist, fall back to VENDOR_LIMITS_FILE file.
         */
        struct stat buffer;
        if (stat(pl->conf_file, &buffer) != 0 && errno == ENOENT)
            pl->conf_file = VENDOR_LIMITS_FILE;
#endif
    }
```

ldv-alt (changes requested):
Also, in the commit message, it should be just `pam_limits`, not `module/pam_limits`. By the way the directory where pam_limits is located is called `modules`.

In the changelog part of the commit message, there is no need to insert an empty line between `* pam_limits.c` and `*
`pam_limits.8.xml`.

modules/pam_limits/pam_limits.c
schubi2: thanks

modules/pam_limits/pam_limits.c
schubi2: look much more better. Thanks !

ldv-alt (approved):

PR #440, opened 2022-02-16
pam_namespace: use vendor specific namespace.conf and namespace.init as fallback

schubi2:
Use the vendor directory as fallback for a distribution provided default config and scripts if there is no configuration in /etc.

pam_namespace.c: Take care about the fallback configuration in vendor directory.
pam_namespace.h: Defining vendor specific files and directories.
pam_namespace.8.xml: Added description for vendor directories and files.
namespace.conf.5.xml: Added description for vendor directories and files.

### DIFF
--- a/modules/pam_namespace/namespace.conf.5.xml
+++ b/modules/pam_namespace/namespace.conf.5.xml
@@ -30,11 +30,27 @@
       directory path and the instance directory path as its arguments.
     </para>
 
-    <para>
+    <para condition="without_vendordir">
       The <filename>/etc/security/namespace.conf</filename> file specifies
       which directories are polyinstantiated, how they are polyinstantiated,
       how instance directories would be named, and any users for whom
       polyinstantiation would not be performed.
+    </para>
+
+    <para condition="with_vendordir">
+      The <filename>/etc/security/namespace.conf</filename> file
+      ( or <filename>%vendordir%/security/namespace.conf</filename> if it does
+      not exist) specifies which directories are polyinstantiated, how they are
+      polyinstantiated, how instance directories would be named, and any users
+      for whom polyinstantiation would not be performed.
+      Then individual <filename>*.conf</filename> files from the
+      <filename>/etc/security/namespace.d/</filename> and
+      <filename>%vendordir%/security/namespace.d</filename> directories are taken too.
+      If <filename>/etc/security/namespace.d/@filename@.conf</filename> exists, then
+      <filename>%vendordir%/security/namespace.d/@filename@.conf</filename> will not be used.
+      All <filename>namespace.d/*.conf</filename> files are sorted by their
+      <filename>@filename@.conf</filename> in lexicographic order regardless of which
+      of the directories they reside in.
     </para>
 
     <para>
--- a/modules/pam_namespace/pam_namespace.c
+++ b/modules/pam_namespace/pam_namespace.c
@@ -39,6 +39,94 @@
 #include "pam_namespace.h"
 #include "argv_parse.h"
 
+/* --- evaluting all files in VENDORDIR/security/namespace.d and /etc/security/namespace.d --- */
+static const char *base_name(const char *path)
+{
+    const char *base = strrchr(path, '/');
+    return base ? base+1 : path;
+}
+
+static int
+compare_filename(const void *a, const void *b)
+{
+	return strcmp(base_name(* (char * const *) a),
+		      base_name(* (char * const *) b));
+}
+
+/* Evaluating a list of files which have to be parsed in the right order:
+ *
+ * - If etc/security/namespace.d/@filename@.conf exists, then
+ *   %vendordir%/security/namespace.d/@filename@.conf should not be used.
+ * - All files in both namespace.d directories are sorted by their @filename@.conf in
+ *   lexicographic order regardless of which of the directories they reside in. */
+static char **read_namespace_dir(struct instance_data *idata)
+{
+	glob_t globbuf;
+	size_t i=0;
+	int glob_rv = glob(NAMESPACE_D_GLOB, GLOB_ERR | GLOB_NOSORT, NULL, &globbuf);
+	char **file_list;
+	size_t file_list_size = glob_rv == 0 ? globbuf.gl_pathc : 0;
+
+#ifdef VENDOR_NAMESPACE_D_GLOB
+	glob_t globbuf_vendor;
+	int glob_rv_vendor = glob(VENDOR_NAMESPACE_D_GLOB, GLOB_ERR | GLOB_NOSORT, NULL, &globbuf_vendor);
+	if (glob_rv_vendor == 0)
+	    file_list_size += globbuf_vendor.gl_pathc;
+#endif
+	file_list = malloc((file_list_size + 1) * sizeof(char*));
+	if (file_list == NULL) {
+	    pam_syslog(idata->pamh, LOG_ERR, "Cannot allocate memory for file list: %m");
+#ifdef VENDOR_NAMESPACE_D_GLOB
+	    if (glob_rv_vendor == 0)
+		globfree(&globbuf_vendor);
+#endif
+	    if (glob_rv == 0)
+		globfree(&globbuf);
+	    return NULL;
+	}
+
+	if (glob_rv == 0) {
+	    for (i = 0; i < globbuf.gl_pathc; i++) {
+		file_list[i] = strdup(globbuf.gl_pathv[i]);
+		if (file_list[i] == NULL) {
+		    pam_syslog(idata->pamh, LOG_ERR, "strdup failed: %m");
+		    break;
+		}
+	    }
+	}
+#ifdef VENDOR_NAMESPACE_D_GLOB
+	if (glob_rv_vendor == 0) {
+	    for (size_t j = 0; j < globbuf_vendor.gl_pathc; j++) {
+		if (glob_rv == 0 && globbuf.gl_pathc > 0) {
+		    int double_found = 0;
+		    for (size_t k = 0; k < globbuf.gl_pathc; k++) {
+			if (strcmp(base_name(globbuf.gl_pathv[k]),
+				   base_name(globbuf_vendor.gl_pathv[j])) == 0) {
+				double_found = 1;
+				break;
+			}
+		    }
+		    if (double_found)
+			continue;
+		}
+		file_list[i] = strdup(globbuf_vendor.gl_pathv[j]);
+		if (file_list[i] == NULL) {
+		    pam_syslog(idata->pamh, LOG_ERR, "strdup failed: %m");
+		    break;
+		}
+		i++;
+	    }
+	    globfree(&globbuf_vendor);
+	}
+#endif
+	file_list[i] = NULL;
+	qsort(file_list, i, sizeof(char *), compare_filename);
+	if (glob_rv == 0)
+	    globfree(&globbuf);
+
+	return file_list;
+}
+
 /*
  * Adds an entry for a polyinstantiated directory to the linked list of
  * polyinstantiated directories. It is called from process_line() while
@@ -624,8 +712,6 @@ static int parse_config_file(struct instance_data *idata)
     char *line;
     int retval;
     size_t len = 0;
-    glob_t globbuf;
-    const char *oldlocale;
     size_t n;
 
     /*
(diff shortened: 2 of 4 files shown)


schubi2:
Currently we do not use libeconf here until we have come to a decision how we are proceeding with the VENDORDIR and USE_ECONF definition. 
If we do not use libeconf here I would suggest to move the "read_access_dir" and "compare_filename" to an own library because they are already be used by different modules in he same way. This would reduce the code :-)

modules/pam_namespace/pam_namespace.c
```diff
@@ -664,13 +727,16 @@ static int parse_config_file(struct instance_data *idata)
      * process_line to process each line.
      */
 
-    memset(&globbuf, '\0', sizeof(globbuf));
-    oldlocale = setlocale(LC_COLLATE, "C");
-    glob(NAMESPACE_D_GLOB, 0, NULL, &globbuf);
-    if (oldlocale != NULL)
-	setlocale(LC_COLLATE, oldlocale);
-
```
thkukuk, line 672: The old code did change the locale to "C" to have a reproduceable output.
The new code does not have this. Since the glob() output depends on LC_COLLATE, we should set the locale again to make sure that no surprises happen.

thkukuk (changes requested):

modules/pam_namespace/pam_namespace.c
thkukuk: Sorry, I see that we have an additional "sort" with "strcmp".
So the result can be different than before, as glob() uses strcoll() or something similar internal.
If admins only use ASCII characters it could only become a problem with spaces or '-', not sure here. If admins uses UTF-8 characters, then the result may be surprising. But as we used "C" and not "C.UTF-8" already before, filenames with UTF-8 are not really well handled anyways.

@ldv-alt @t8m what do you think?

I see the following options:
- keep the current strcmp() implementation, which always returns something reproducible, but may be different to before. Advantage is, we don't need to mess around with setlocale() and locale handling
- use strcoll() and set locale to "C"
- use strcoll() and try to set locale to "C.UTF-8" and if that does not exist, fallback to "C".

ldv-alt: The manual page states that `In the POSIX or C locales strcoll(3) is equivalent to strcmp(3)`.
If that claim is correct, then we can safely use strcmp.

thkukuk: I cannot find that in my versions of strcoll manual pages nor in the POSIX version. But with the C locale we sort in byte order, which matches strcmp, so we seem to be fine here.

thkukuk (approved):

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,71 @@
...
+compare_filename(const void *a, const void *b)
+{
+	char *a_dup = strdup(*(const char * const *)a);
+	char *b_dup = strdup(*(const char * const *)b);
+	int ret = strcmp(basename(a_dup),
+			 basename(b_dup));
+	free(a_dup);
+	free(b_dup);
+	return ret;
+}
```
ldv-alt, line 53: As no memory allocation is needed here, please find a way to avoid it, consider using strrchr.

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,71 @@
...
+	glob_t globbuf_vendor;
+	int glob_rv_vendor = glob(VENDOR_NAMESPACE_D_GLOB, GLOB_ERR, NULL, &globbuf_vendor);
+	char **file_list = malloc((globbuf_vendor.gl_pathc + globbuf.gl_pathc + 1) * sizeof(char*));
+#else
+	char **file_list = malloc((globbuf.gl_pathc + 1) * sizeof(char*));
+#endif
+
+	if (!glob_rv) {
+	    for (i = 0; globbuf.gl_pathv[i] != NULL; i++) {
+	        file_list[i] = strdup(globbuf.gl_pathv[i]);
```
ldv-alt, line 76: `strdup` may return NULL anytime.

ldv-alt (changes requested):
This is not a complete review yet, please double check all error handling in the patch.

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,71 @@
...
+ *
+ * - If etc/security/namespace.d/@filename@.conf exists, then
+ *   %vendordir%/security/namespace.d/@filename@.conf should not be used.
+ * - All files in both namespace.d directories are sorted by their @filename@.conf in
+ *   lexicographic order regardless of which of the directories they reside in. */
+static char **read_access_dir(void)
+{
+	glob_t globbuf;
+	int i=0;
+	int glob_rv = glob(NAMESPACE_D_GLOB, GLOB_ERR, NULL, &globbuf);
```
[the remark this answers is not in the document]
schubi2: good point !

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,71 @@
...
+#ifdef VENDOR_NAMESPACE_D_GLOB
+	glob_t globbuf_vendor;
+	int glob_rv_vendor = glob(VENDOR_NAMESPACE_D_GLOB, GLOB_ERR, NULL, &globbuf_vendor);
+	char **file_list = malloc((globbuf_vendor.gl_pathc + globbuf.gl_pathc + 1) * sizeof(char*));
+#else
+	char **file_list = malloc((globbuf.gl_pathc + 1) * sizeof(char*));
```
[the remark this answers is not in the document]
schubi2: I have changed it. Thanks !

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,71 @@
...
+#endif
+
+	if (!glob_rv) {
+	    for (i = 0; globbuf.gl_pathv[i] != NULL; i++) {
```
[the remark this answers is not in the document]
schubi2: I think it is NULL terminated. Otherwise all my tests would have been failed. But the docu says nothing about it. So, you are right to use gl_pathc instead.

modules/pam_namespace/pam_namespace.c
schubi2: File_list will be terminated with NULL. I have added an error message if this happens

ldv-alt (dismissed):
This PR has issues similar to those found in #434 and #435, please address them.

schubi2:
> This PR has issues similar to those found in #434 and #435, please address them.

I hope, I have fixed all of them.

thkukuk:

ldv-alt:

ldv-alt (dismissed):
This patch has the same issues as #434 and #435, please address them.

schubi2:
> This patch has the same issues as #434 and #435, please address them.

I hope I have matched all of them.

ldv-alt:

thkukuk:
If nobody objects I will merge the next days.

ldv-alt:

modules/pam_namespace/namespace.conf.5.xml
```diff
@@ -30,13 +30,29 @@
...
 
+    <para condition="with_vendordir">
+      The <filename>/etc/security/namespace.conf</filename> file
+      ( or <filename>%vendordir%/security/namespace.conf</filename> if it does
+      not exist) specifies which directories are polyinstantiated, how they are
+      polyinstantiated, how instance directories would be named, and any users
+      for whom polyinstantiation would not be performed.
+      Then individual <filename>*.conf</filename> files from the
+      <filename>/etc/security/namespace.d/</filename> and
+      <filename>%vendordir%/security/namespace.d</filename> directory are taken too.
```
ldv-alt, line 48: Since you are talking about two directories here, I suppose `directories are read` would be more correct.

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,95 @@
...
+static const char *base_name(const char *path)
+{
+    const char *base = strrchr(path, '/');
+    return base ? base+1 : path;
+}
+
+static int
+compare_filename(const void *a, const void *b)
+{
+	return strcmp(base_name(* (const char * const *) a), base_name(* (const char * const *) b));
```
ldv-alt, line 52: This line is too long, please wrap, e.g.
```
```

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,95 @@
...
+ *   %vendordir%/security/namespace.d/@filename@.conf should not be used.
+ * - All files in both namespace.d directories are sorted by their @filename@.conf in
+ *   lexicographic order regardless of which of the directories they reside in. */
+static char **read_namespace_dir(struct instance_data *idata)
+{
+	glob_t globbuf;
+	size_t i=0;
+	int glob_rv = glob(NAMESPACE_D_GLOB, GLOB_ERR | GLOB_NOSORT, NULL, &globbuf);
+	char **file_list;
+	int file_list_size = glob_rv == 0 ? globbuf.gl_pathc : 0;
```
ldv-alt, line 67: `file_list_size` should be of the same type as `globbuf.gl_pathc`, which is `size_t`.

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,95 @@
...
+	    }
+	}
+#ifdef VENDOR_NAMESPACE_D_GLOB
+	if (glob_rv_vendor == 0) {
+	    for (size_t j = 0; j < globbuf_vendor.gl_pathc; j++) {
+		if (glob_rv == 0 && globbuf.gl_pathc > 0) {
+		    int double_found = 0;
+		    for (size_t k = 0; k < globbuf.gl_pathc; k++) {
+		        if (strcmp(basename(globbuf.gl_pathv[k]),
+				   basename(globbuf_vendor.gl_pathv[j])) == 0) {
```
ldv-alt, line 103: Please consider using `base_name` instead of `basename` here as well.

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,95 @@
...
+		    break;
+		}
+		i++;
+	    }
+	    globfree(&globbuf_vendor);
+	}
+#endif
+	file_list[i] = NULL;
+	if (i>0)
+	    qsort(file_list, i, sizeof(char *), compare_filename);
```
ldv-alt, line 123: The `i>0` check is redundant here.

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,95 @@
...
+
+        if (glob_rv == 0)
+            globfree(&globbuf);
```
ldv-alt, line 126: Please indent these lines the same way as the neighboring lines.

modules/pam_namespace/pam_namespace.c
```diff
@@ -707,14 +795,17 @@ static int parse_config_file(struct instance_data *idata)
...
-	if (n >= globbuf.gl_pathc)
+	if (filename_list == NULL || filename_list[n] == NULL)
 	    break;
 
-	confname = globbuf.gl_pathv[n];
-	n++;
+	confname = filename_list[n++];
+    }
+    if (filename_list != NULL) {
+        for (int i = 0; filename_list[i] != NULL; i++)
```
ldv-alt, line 804: `i` should be of the same type here as `i` in the function where the array is initialized.

ldv-alt (changes requested):
Also, there is no need to insert empty lines in the changelog part of the commit message.

modules/pam_namespace/namespace.conf.5.xml
schubi2: fixed

modules/pam_namespace/pam_namespace.c
schubi2: fixed

modules/pam_namespace/pam_namespace.c
schubi2: fixed

modules/pam_namespace/pam_namespace.c
schubi2: fixed

modules/pam_namespace/pam_namespace.c
schubi2: removed

modules/pam_namespace/pam_namespace.c
schubi2: Hm, somehow the "if (glob_rv == 0)" is outdated and I have not found it in previous code too. At least I have fixed the indent.

modules/pam_namespace/pam_namespace.c
schubi2: fixed

ldv-alt (changes requested):
Looks like you've reverted most of the changes made in the previous round, please fix that.

schubi2:
> > 
> 
> > > Looks like you've reverted most of the changes made in the previous round, please fix that.
> > 
> > 
> > Yes, you are right. I have no idea why it has gone wrong. I hope I have fixed it.......
> > Just a moment please, I have seen something around setlocale which I do not understand and which I have to check it again.....

It is OK. It is regarding the sort algorithm and we have had already discussed it.

modules/pam_namespace/pam_namespace.c
```diff
@@ -39,6 +39,94 @@
...
+	if (file_list == NULL) {
+#ifdef VENDOR_NAMESPACE_D_GLOB
+            if (glob_rv_vendor == 0)
+                globfree(&globbuf_vendor);
+#endif
+            if (glob_rv == 0)
+                globfree(&globbuf);
+            pam_syslog(idata->pamh, LOG_ERR, "Cannot allocate memory for file list: %m");
+	    return NULL;
+	}
```
ldv-alt, line 86: `pam_syslog` should be done before `globfree` because the former uses `%m` and the latter can clobber `errno`.
It is a regression since commit 9bbecd3f.

ldv-alt (changes requested):
Also, if your commit messages contain long lines, please wrap them, too.

ldv-alt (approved):

modules/pam_namespace/pam_namespace.c
schubi2: Ah, I see you have already fixed it. Thank you !

PR #441, opened 2022-02-16
Default enablement  of vendordir

schubi2:
Vendordir is enabled via default and can be disabled by the option --disable-vendor again.

configure.ac: new option --disable-vendor.


thkukuk:
Discussed with Stefan, as long as there is no general agreement where the vendordir will be we withdraw this, doesn't make much sense. Waiting for https://uapi-group.org to define that.

PR #442, opened 2022-02-17
module/pam_time: Fix clang compilation warning

schubi2:
Fixed warning: "Result of comparison against a string literal is unspecified
(use an explicit string comparison function instead)"

pam_time.c: Do not compare char * string with a constant.

### DIFF
--- a/modules/pam_time/pam_time.c
+++ b/modules/pam_time/pam_time.c
@@ -58,7 +58,7 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv, const char **
 {
     int ctrl = 0;
 
-    *conffile = PAM_TIME_CONF;
+    *conffile = NULL;
     /* step through arguments */
     for (; argc-- > 0; ++argv) {
 	const char *str;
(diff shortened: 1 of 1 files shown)


t8m (approved):

modules/pam_time/pam_time.c
```diff
@@ -83,7 +83,7 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv, const char **
     }
 
 #ifdef VENDOR_PAM_TIME_CONF
-    if (*conffile == PAM_TIME_CONF) {
+    if (strcmp(*conffile, PAM_TIME_CONF) == 0) {
```
t8m, line 86: Another option would be to make the default value to be an static const char default_conf[] = PAM_TIME_CONF; And then you can keep using the pointer comparison.

ldv-alt (changes requested):
Comparing pointers is a perfectly valid operation.
Comparing strings, on the other hand, would change `conffile=` semantics.
I'd suggest to initialize `*conffile` with `NULL` instead of `PAM_TIME_CONF`.

schubi2:
> Comparing pointers is a perfectly valid operation. Comparing strings, on the other hand, would change `conffile=` semantics. I'd suggest to initialize `*conffile` with `NULL` instead of `PAM_TIME_CONF`.

Yes, good idea !

thkukuk (approved):

ldv-alt:
The patch itself is fine.
Could you cite the text printed by the compiler in the commit message, please?
Also please wrap long lines in commit messages, too.
I suppose it's a warning upgraded to error by -Werror.
If that's the case, then the commit message should say `fix clang compilation warning`
rather than the current `Fixed compile error of clang.`

schubi2:
> The patch itself is fine. Could you cite the text printed by the compiler in the commit message, please? Also please wrap long lines in commit messages, too. I suppose it's a warning upgraded to error by -Werror. If that's the case, then the commit message should say `fix clang compilation warning` rather than the current `Fixed compile error of clang.`

Thanks. I have changed it.

ldv-alt (approved):

PR #443, opened 2022-02-17
pam_usertype: only use SYS_UID_MAX for system users

ikerexxe:
modules/pam_usertype/pam_usertype.c: stop using SYS_UID_MIN to check if
it is a system account, because all accounts below the SYS_UID_MAX are
system users.
modules/pam_usertype/pam_usertype.8.xml: remove reference to SYS_UID_MIN
as it is no longer user to calculate the system accounts.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1949137

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -632,11 +632,6 @@ test -n "$opt_uidmin" ||
           opt_uidmin=1000
 AC_DEFINE_UNQUOTED(PAM_USERTYPE_UIDMIN, $opt_uidmin, [Minimum regular user uid.])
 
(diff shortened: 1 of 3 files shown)


ldv-alt:
Is PAM_USERTYPE_SYSUIDMIN needed after this change,
or should it rather be removed, too?

ikerexxe:
I didn't notice that `PAM_USERTYPE_SYSUIDMIN` was defined in configure.ac. As we aren't using it anymore I've removed it. Thanks for catching that.

ldv-alt (approved):

thkukuk (approved):

PR #445, opened 2022-02-23
pam_nologin: don't print empty message

thkukuk:
Conversation functions most of the time add an extra "\n" to messages from PAM modules. This leads to disturbing line breaks and empty lines in the output.
Better to do not print an empty message at all.

### DIFF
--- a/modules/pam_nologin/pam_nologin.c
+++ b/modules/pam_nologin/pam_nologin.c
@@ -79,7 +79,6 @@ static int perform_check(pam_handle_t *pamh, struct opt_s *opts)
 
     if (fd >= 0) {
 
-	char *mtmp=NULL;
 	int msg_style = PAM_TEXT_INFO;
 	struct passwd *user_pwd;
 	struct stat st;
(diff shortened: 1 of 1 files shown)


modules/pam_nologin/pam_nologin.c
```diff
@@ -99,21 +99,25 @@ static int perform_check(pam_handle_t *pamh, struct opt_s *opts)
...
-
-	if (pam_modutil_read(fd, mtmp, st.st_size) == st.st_size) {
+	/* Don't print anything if the message is empty, will only
+	   disturb the output with empty lines */
+	if (st.st_size > 0) {
+	    mtmp = malloc(st.st_size+1);
+	    if (!mtmp) {
+	        pam_syslog(pamh, LOG_CRIT, "out of memory");
+		retval = PAM_BUF_ERR;
+		goto clean_up_fd;
```
ikerexxe, line 109: Different indentation level for content inside a block.

modules/pam_nologin/pam_nologin.c
```diff
@@ -99,21 +99,25 @@ static int perform_check(pam_handle_t *pamh, struct opt_s *opts)
...
+	    }
+
+	    if (pam_modutil_read(fd, mtmp, st.st_size) == st.st_size) {
 		mtmp[st.st_size] = '\0';
 		(void) pam_prompt (pamh, msg_style, NULL, "%s", mtmp);
```
ikerexxe, line 114: The content inside the block appears to have the same level of indentation as that outside the block.

modules/pam_nologin/pam_nologin.c
thkukuk: Github does not seem to show TABs and whitespaces correct...

modules/pam_nologin/pam_nologin.c
thkukuk: Don't trust your webbrowser...

modules/pam_nologin/pam_nologin.c
```diff
@@ -99,21 +99,25 @@ static int perform_check(pam_handle_t *pamh, struct opt_s *opts)
...
-	    goto clean_up_fd;
-	}
-
-	if (pam_modutil_read(fd, mtmp, st.st_size) == st.st_size) {
-		mtmp[st.st_size] = '\0';
-		(void) pam_prompt (pamh, msg_style, NULL, "%s", mtmp);
+	/* Don't print anything if the message is empty, will only
+	   disturb the output with empty lines */
+	if (st.st_size > 0) {
+	    mtmp = malloc(st.st_size+1);
```
ldv-alt, line 105: Could you also move the declaration of `mtmp` from
https://github.com/linux-pam/linux-pam/blob/016d5e2bc0423f5226191143329ad7e50fde82fd/modules/pam_nologin/pam_nologin.c#L82
inside this new block, e.g.
```
if (st.st_size > 0) {
    char *mtmp = malloc(st.st_size+1);
```
?

ldv-alt (approved):

PR #447, opened 2022-02-24
pam_access: handle hostnames in access.conf

thkukuk:
According to the manual page, the following entry is valid but does not work:
-:root:ALL EXCEPT localhost

See https://bugzilla.suse.com/show_bug.cgi?id=1019866

Patched is based on PR#226 from Josef Moellers

### DIFF
--- a/modules/pam_access/pam_access.c
+++ b/modules/pam_access/pam_access.c
@@ -640,7 +640,7 @@ remote_match (pam_handle_t *pamh, char *tok, struct login_info *item)
       if ((str_len = strlen(string)) > tok_len
 	  && strcasecmp(tok, string + str_len - tok_len) == 0)
 	return YES;
-    } else if (tok[tok_len - 1] == '.') {
+    } else if (tok[tok_len - 1] == '.') {       /* internet network numbers (end with ".") */
       struct addrinfo hint;
 
       memset (&hint, '\0', sizeof (hint));
@@ -681,7 +681,7 @@ remote_match (pam_handle_t *pamh, char *tok, struct login_info *item)
       return NO;
     }
 
-    /* Assume network/netmask with an IP of a host.  */
+    /* Assume network/netmask, IP address or hostname.  */
     return network_netmask_match(pamh, tok, string, item);
 }
 
@@ -699,7 +699,7 @@ string_match (pam_handle_t *pamh, const char *tok, const char *string,
     /*
      * If the token has the magic value "ALL" the match always succeeds.
      * Otherwise, return YES if the token fully matches the string.
-	 * "NONE" token matches NULL string.
+     * "NONE" token matches NULL string.
      */
 
     if (strcasecmp(tok, "ALL") == 0) {		/* all: always matches */
@@ -717,7 +717,8 @@ string_match (pam_handle_t *pamh, const char *tok, const char *string,
 
 /* network_netmask_match - match a string against one token
  * where string is a hostname or ip (v4,v6) address and tok
- * represents either a single ip (v4,v6) address or a network/netmask
+ * represents either a hostname, a single ip (v4,v6) address
+ * or a network/netmask
  */
 static int
 network_netmask_match (pam_handle_t *pamh,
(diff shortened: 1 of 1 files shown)


modules/pam_access/pam_access.c
```diff
@@ -739,7 +744,7 @@ network_netmask_match (pam_handle_t *pamh,
 	*netmask_ptr = 0;
 	netmask_ptr++;
 
-	if (isipaddr(tok, &addr_type, NULL) == NO)
+	if (isipaddr(tok, &addr_type, &tok_addr) == NO)
```
ldv-alt, line 747: `tok_addr` seems to be unused after this initialization, I wonder what was the intent of this change.

modules/pam_access/pam_access.c
```diff
@@ -763,19 +768,47 @@ network_netmask_match (pam_handle_t *pamh,
...
 	  }
-	}
+
+	/*
+	 * Although isipaddr() has already converted the IP address,
+	 * we call getaddrinfo here to properly construct an addrinfo list
+	 */
+	memset (&hint, '\0', sizeof (hint));
+	hint.ai_flags = 0;
+	hint.ai_family = AF_UNSPEC;
```
ldv-alt, line 778: `ai_family` should probably be assigned to `addr_type` instead of `AF_UNSPEC`.

modules/pam_access/pam_access.c
```diff
@@ -763,19 +768,47 @@ network_netmask_match (pam_handle_t *pamh,
...
+
+	ai = NULL;	/* just to be on the safe side */
+
+	if (getaddrinfo (string, NULL, &hint, &ai) != 0)
 	  {
+	    pam_syslog(pamh, LOG_ERR, "cannot resolve hostname \"%s\"", string);
```
ldv-alt, line 802: Is it really an error?

modules/pam_access/pam_access.c
```diff
@@ -763,19 +768,47 @@ network_netmask_match (pam_handle_t *pamh,
...
+	 * It is either an IP address or a hostname.
+	 * Let getaddrinfo sort everything out
+	 */
+	memset (&hint, '\0', sizeof (hint));
+	hint.ai_flags = 0;
+	hint.ai_family = AF_UNSPEC;
+
+	ai = NULL;	/* just to be on the safe side */
+
+	if (getaddrinfo (string, NULL, &hint, &ai) != 0)
```
ldv-alt, line 800: Should it be `tok` instead of of `string` here?

modules/pam_access/pam_access.c
```diff
@@ -788,29 +821,52 @@ network_netmask_match (pam_handle_t *pamh,
...
-		if (are_addresses_equal(buf, tok, netmask_ptr))
+		for (runp1 = ai; runp1 != NULL; runp1 = runp1->ai_next)
 		  {
-		    return YES;
+                    char buf1[INET6_ADDRSTRLEN];
+
+                    if (runp->ai_family != runp1->ai_family)
+                      continue;
+
+                    (void) getnameinfo (runp1->ai_addr, runp1->ai_addrlen, buf1, sizeof (buf1), NULL, 0, NI_NUMERICHOST);
```
ldv-alt, line 839: Is it OK to ignore the return value of `getnameinfo`?

modules/pam_access/pam_access.c
```diff
@@ -788,29 +821,52 @@ network_netmask_match (pam_handle_t *pamh,
         else
 	  {
 	    struct addrinfo *runp = item->res;
+	    struct addrinfo *runp1;
```
ldv-alt, line 824: There are two `return NO` statements without `freeaddrinfo(ai)`.

ldv-alt (dismissed):

modules/pam_access/pam_access.c
thkukuk: Correct, seems to come from an old version of the patch, reverted.

modules/pam_access/pam_access.c
thkukuk: The whole "hint" isn't necessary at all here, removed.

modules/pam_access/pam_access.c
thkukuk: Yes, this is really an error here.

modules/pam_access/pam_access.c
thkukuk: Thanks for spotting that!

modules/pam_access/pam_access.c
thkukuk: Since we only convert a valid IP address into a string using a provided buffer, this can theoretically not fail.
But should something change in the implementation and it fails, we compare random data.
To be on the really secure side, I added checks.

ldv-alt (approved):
All issues I found in the previous review seem to be addressed.

ldv-alt:
Ideally I'd like to have a test for such changes, but that's out of scope of this PR.

t8m (approved):

PR #449, opened 2022-03-10
Update Korean translation

darkcircle:
The original translation came from fedoraproject Weblate has totally broken, and it has no formal expression.

It would be helpful for keeping the consistency of the Korean translation in the Linux CLI environment.


ldv-alt:
Since the project uses fedora weblate to maintain translations, please consider using 
https://translate.fedoraproject.org/projects/linux-pam/master/ko/ instead of this PR.

darkcircle:
@ldv-alt  but I cannot access there due to unknown reason, I sent a message to ask let me know whether there is something problem to access weblate or not, but I did not received any comment yet.

darkcircle:
please accept this PR, I still cannot access weblate. There is no response from server operator.

PR #456, opened 2022-04-15
Fixing issue#455

pjessen:
in module pam_motd, filter_dirents() relied on the filetype being available.

### DIFF
--- a/modules/pam_motd/pam_motd.c
+++ b/modules/pam_motd/pam_motd.c
@@ -166,9 +166,24 @@ static int compare_strings(const void *a, const void *b)
     }
 }
 
+static const char* motddir;
+
 static int filter_dirents(const struct dirent *d)
 {
-    return (d->d_type == DT_REG || d->d_type == DT_LNK);
+    struct stat s;
+    char *fullpath;
+    int rc;
+
+    while(1){
+        if ( d->d_type==DT_DIR ) break;
+        if ( join_dir_strings( &fullpath, motddir, d->d_name )<=0) break;
+        rc=stat( fullpath, &s );
+        free(fullpath);     /* free the memory alloc'ed by join_dir_strings */
+        if ( rc!=0 ) break; /* if the stat() somehow failed */
+
+        return S_IFREG==(s.st_mode&S_IFMT) || S_IFLNK==(s.st_mode&S_IFMT);
+    }
+    return 0;
 }
 
 static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
@@ -199,8 +214,8 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
 
     for (i = 0; i < num_motd_dirs; i++) {
 	int rv;
-	rv = scandir(motd_dir_path_split[i], &(dirscans[i]),
-		filter_dirents, alphasort);
+        motddir=motd_dir_path_split[i];
+        rv = scandir(motddir, &(dirscans[i]), filter_dirents, alphasort);
 	if (rv < 0) {
 	    if (errno != ENOENT || report_missing) {
 		pam_syslog(pamh, LOG_ERR, "error scanning directory %s: %m",
--- a/modules/pam_motd/pam_motd.c
+++ b/modules/pam_motd/pam_motd.c
@@ -174,16 +174,26 @@ static int filter_dirents(const struct dirent *d)
     char *fullpath;
     int rc;
 
-    while(1){
-        if ( d->d_type==DT_DIR ) break;
-        if ( join_dir_strings( &fullpath, motddir, d->d_name )<=0) break;
-        rc=stat( fullpath, &s );
-        free(fullpath);     /* free the memory alloc'ed by join_dir_strings */
-        if ( rc!=0 ) break; /* if the stat() somehow failed */
-
-        return S_IFREG==(s.st_mode&S_IFMT) || S_IFLNK==(s.st_mode&S_IFMT);
+    switch( d->d_type )
+    {
+	case DT_REG:
+	case DT_LNK:
+            return 1;      /* regular files and symlinks are good         */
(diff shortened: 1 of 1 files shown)


modules/pam_motd/pam_motd.c
```diff
@@ -166,9 +166,24 @@ static int compare_strings(const void *a, const void *b)
     }
 }
 
+static const char* motddir;
+
 static int filter_dirents(const struct dirent *d)
 {
-    return (d->d_type == DT_REG || d->d_type == DT_LNK);
```
arvidjaar, line 171: This will punish every filesystem that already returns filetype here. Fallback is needed only if `d_type == DT_UNKNOWN`.

modules/pam_motd/pam_motd.c
```diff
@@ -166,9 +166,24 @@ static int compare_strings(const void *a, const void *b)
...
+    struct stat s;
+    char *fullpath;
+    int rc;
+
+    while(1){
```
arvidjaar, line 177: There is no loop here, why `while`?

modules/pam_motd/pam_motd.c
```diff
@@ -166,9 +166,24 @@ static int compare_strings(const void *a, const void *b)
...
+        if ( d->d_type==DT_DIR ) break;
+        if ( join_dir_strings( &fullpath, motddir, d->d_name )<=0) break;
+        rc=stat( fullpath, &s );
+        free(fullpath);     /* free the memory alloc'ed by join_dir_strings */
```
arvidjaar, line 181: Other code is using `_pam_drop` here.

modules/pam_motd/pam_motd.c
```diff
@@ -166,9 +166,24 @@ static int compare_strings(const void *a, const void *b)
...
+        if ( rc!=0 ) break; /* if the stat() somehow failed */
+
+        return S_IFREG==(s.st_mode&S_IFMT) || S_IFLNK==(s.st_mode&S_IFMT);
```
arvidjaar, line 184: `stat()` will never resturn `S_IFLINK`. To match original condition, it should be `lstat()`.

modules/pam_motd/pam_motd.c
pjessen: My thinking was that it will be at most 2-3 stat() calls, in a process that isn't performance critical. Keep the code simple and small.  I'll fix it.

modules/pam_motd/pam_motd.c
pjessen: I'm "grouping" the conditions inside, enables me to use break.

modules/pam_motd/pam_motd.c
pjessen: Makes sense, I had not noticed.

modules/pam_motd/pam_motd.c
ldv-alt: Please use
```
if (d->d_type != DT_UNKNOWN)
    return d->d_type == DT_REG || d->d_type == DT_LNK;
```

modules/pam_motd/pam_motd.c
ldv-alt: The loop is not needed, you can just do `return 0` instead;

modules/pam_motd/pam_motd.c
pjessen: Thanks for the feedback. Any comments on using this instead:
```
static int filter_dirents(const struct dirent *d)
{ 
    struct stat s;
    char *fullpath;
    int rc;

    switch( d->d_type )
    {
        case DT_REG:
        case DT_LNK:
            return 1;      /* regular files and symlinks are good         */
        case DT_UNKNOWN:   /* with unknown, we have a bit more work to do */
            break;
        case DT_DIR:       /* We don't want directories   */
        default:           /* nor anything else           */
            return 0;
    }

```
followed by the code that does lstat() etc.

modules/pam_motd/pam_motd.c
pjessen: Yup agree, just a personal "preference".

modules/pam_motd/pam_motd.c
pjessen: right, I missed that.

modules/pam_motd/pam_motd.c
ldv-alt: Yes, this looks better.
btw, please follow the style of this module, e.g.
```
switch (d->d_type) {
```

modules/pam_motd/pam_motd.c
```diff
@@ -166,9 +166,24 @@ static int compare_strings(const void *a, const void *b)
     }
 }
 
+static const char* motddir;
```
ldv-alt, line 169: Static non-const variable is problematic for those of pam modules that are intended to be MT-safe.

modules/pam_motd/pam_motd.c
pjessen: yup. muscle memory.

modules/pam_motd/pam_motd.c
pjessen: yes - is motd supposed to be mt-safe?  it's using scandir(), which in turn uses readdir().  
Do you have a suggestion of an alternative solution?

pjessen: I think this might solve it - comments?
```
static const char* motddir( char *d )
{
    static const char* dir;

    if ( d ) dir=d;

    return dir;
}
```

ldv-alt: > yes - is motd supposed to be mt-safe? it's using scandir(), which in turn uses readdir(). Do you have a suggestion of an alternative solution?

scandir(3) manual page claims that scandir() is MT-safe.
I suppose `__thread` qualifier can be used if available.

PR #457, opened 2022-04-20
faillock: load configuration from file

ikerexxe:
* modules/pam_faillock/faillock_config.c: Move configuration loading
functions (read_config_file and set_conf_opt) to this file.
* modules/pam_faillock/faillock_config.h: Move configuration loading
macros and structures.
* modules/pam_faillock/Makefile.am: Add faillock_config.
* modules/pam_faillock/faillock.h: Remove configuration loading macros.
* modules/pam_faillock/pam_faillock.c: Remove configuration loading
functions, macros and structures.
* modules/pam_faillock/main.c: Load configuration from file.
* modules/pam_faillock/faillock.8.xml: Update with the new configuration
option.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1978029


ikerexxe:
I need some time to review those failures as they aren't happening on my system (I don't use clang compiler).

ikerexxe:
Now that the compilation problems are fixed the PR is ready for review.

modules/pam_faillock/main.c
```diff
@@ -51,29 +51,37 @@
...
-		if (strcmp(argv[i], "--dir") == 0) {
+		if (strcmp(argv[i], "--conf") == 0) {
 			++i;
 			if (i >= argc || strlen(argv[i]) == 0) {
-				fprintf(stderr, "%s: No directory supplied.\n", argv[0]);
+				fprintf(stderr, "%s: No configuration directory supplied.\n",
+						argv[0]);
+				return -1;
+			}
+			conf = pam_str_skip_prefix(argv[i], "--conf=");
```
t8m, line 78: This does not look correct. Why would you skip the prefix? There should be none.

modules/pam_faillock/faillock.8.xml
```diff
@@ -55,14 +55,31 @@
 
     <title>OPTIONS</title>
          <variablelist>
+            <varlistentry>
+              <term>
+                <option>--conf <replaceable>/path/to/config-file</replaceable></option>
+              </term>
+              <listitem>
+                <para>
+                  The directory where the configuration file is located. The
```
t8m, line 64: This is not a directory. It's just the file. `The path to the configuration file. The...`

modules/pam_faillock/main.c
```diff
@@ -51,29 +51,37 @@
...
+		}
+		else if (strcmp(argv[i], "--dir") == 0) {
+			++i;
+			if (i >= argc || strlen(argv[i]) == 0) {
+				fprintf(stderr, "%s: No records directory supplied.\n",
+						argv[0]);
 				return -1;
 			}
 			opts->dir = argv[i];
```
t8m, line 87: The way it is implemented the value from configuration file if it is set there always overrides the --dir option.

t8m (changes requested):

modules/pam_faillock/main.c
ikerexxe: This part of the code loads all the command line arguments in variables, but it doesn't set any priority for them. The coding lines that sets the priority for the tally directory are located in [faillock_config.c 169-178](https://github.com/linux-pam/linux-pam/pull/457/files#diff-e188da21b29b52dc3cac68918c0cc91d437eb192ccf9fba04cf19c7dbf9c72a0R169-R178).

modules/pam_faillock/faillock.8.xml
```diff
@@ -55,14 +55,31 @@
...
               </term>
               <listitem>
                 <para>
-                  The directory where the user files with the failure records are kept. The
-                  default is <filename>/var/run/faillock</filename>.
+                  The directory where the user files with the failure records are kept.
+                </para>
+                <para>
+                  The priority to set this option is to use the value provided
+                  from the terminal. If this isn't provided, then the value
```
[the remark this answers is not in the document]
ikerexxe: Fixed

modules/pam_faillock/faillock_config.h
```diff
@@ -0,0 +1,96 @@
...
+};
+
+extern const char default_faillock_conf[PATH_MAX];
+
+int read_config_file(pam_handle_t *pamh, struct options *opts,
+					 const char *cfgfile);
+void set_conf_opt(pam_handle_t *pamh, struct options *opts, const char *name,
+		  const char *value);
+
+#endif /* _FAILLOCK_CONFIG_H */
```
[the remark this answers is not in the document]
ikerexxe: Fixed

modules/pam_faillock/main.c
ikerexxe: You are right. I fixed it.

modules/pam_faillock/faillock.8.xml
ikerexxe: Fixed

modules/pam_faillock/main.c
t8m: Hmm, it still means that `--dir /var/run/faillock` will be overridden.

t8m: And also you need to use `opts->dir = strdup(argv[i]);` here.

t8m: I recommend instead of doing the strcmp in faillock_config, just save the dir option value here and override it in the opts after read_config_file() is called if it is non-NULL.

ikerexxe: Based on your comment I have changed this part of the logic.

I also had to change other things to accommodate everything so I added a new commit. I did this to get a clear view of what changed. If you agree with those changes I'll squash the commit.

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,257 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <syslog.h>
+
+#include <security/pam_modules.h>
+
+#include "faillock_config.h"
+#include "faillock.h"
+
+const char default_faillock_conf[] = FAILLOCK_DEFAULT_CONF;
```
ldv-alt, line 51: I'm sorry, but I don't see why do you export `default_faillock_conf` in the first place.
I suggest making this variable internal to `read_config_file`.
Maybe you won't need `default_faillock_conf` at all.

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,257 @@
...
+	   */
+	  f = fopen(VENDOR_FAILLOCK_DEFAULT_CONF, "r");
+	}
+#endif
+	if (f == NULL) {
+		/* ignore non-existent default config file */
+		if (errno == ENOENT && cfgfile == default_faillock_conf)
+			return PAM_SUCCESS;
+		return PAM_SERVICE_ERR;
+	}
```
ldv-alt, line 96: I suggest to change semantics a bit and use `cfgfile == NULL` as the way to specify the default, e.g.

```
       const char *fname = (cfgfile != NULL) ? cfgfile : FAILLOCK_DEFAULT_CONF;
       FILE *f = fopen(fname, "r");
#ifdef VENDOR_FAILLOCK_DEFAULT_CONF
       if (f == NULL && errno == ENOENT && cfgfile == NULL) {
         /*
          * If the default configuration file in /etc does not exist,
          * try the vendor configuration file as fallback.
          */
         f = fopen(VENDOR_FAILLOCK_DEFAULT_CONF, "r");
       }
#endif
       if (f == NULL) {
               /* ignore non-existent default config file */
               if (errno == ENOENT && cfgfile == NULL)
                       return PAM_SUCCESS;
               return PAM_SERVICE_ERR;
       }
```

modules/pam_faillock/pam_faillock.c
```diff
@@ -175,193 +130,18 @@ args_parse(pam_handle_t *pamh, int argc, const char **argv,
...
 
 	if (opts->dir == NULL) {
-		pam_syslog(pamh, LOG_CRIT, "Error allocating memory: %m");
-		opts->fatal_error = 1;
+		opts->dir = strdup(FAILLOCK_DEFAULT_TALLYDIR);
+		if (opts->dir == NULL) {
+			pam_syslog(pamh, LOG_CRIT, "Error allocating memory: %m");
+			opts->fatal_error = 1;
+		}
 	}
```
ldv-alt, line 138: Could we avoid unnecessary memory allocations, please?
For example, could we leave `opts->dir == NULL` unchanged and treat it later as `FAILLOCK_DEFAULT_TALLYDIR`?

modules/pam_faillock/faillock_config.c
ikerexxe: You are right. I got rid of that variable.

modules/pam_faillock/faillock_config.c
ikerexxe: Thanks for the suggestion, it helped me understand your intentions.

modules/pam_faillock/pam_faillock.c
ikerexxe: Done

ikerexxe:
Could I get another round of reviews?

ikerexxe:
@t8m any feedback?

modules/pam_faillock/main.c
```diff
@@ -51,32 +51,44 @@
...
+	opts->dir = NULL;
 	opts->progname = argv[0];
 
 	for (i = 1; i < argc; ++i) {
-		if (strcmp(argv[i], "--dir") == 0) {
+		if (strcmp(argv[i], "--conf") == 0) {
 			++i;
 			if (i >= argc || strlen(argv[i]) == 0) {
-				fprintf(stderr, "%s: No directory supplied.\n", argv[0]);
+				fprintf(stderr, "%s: No configuration directory supplied.\n",
```
t8m, line 74: `No configuration file supplied.`

modules/pam_faillock/main.c
```diff
@@ -111,10 +129,11 @@ do_user(struct options *opts, const char *user)
 	int rv;
 	struct tally_data tallies;
 	struct passwd *pwd;
+	const char *dir = (opts->dir != NULL) ? opts->dir : FAILLOCK_DEFAULT_TALLYDIR;
```
t8m, line 132: Please replace this repeated expression with a function `const char *get_tally_dir(struct option *opts)` implemented in faillock_conf.c.

t8m (changes requested):
I am sorry for neglecting this for so long.

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,262 @@
...
+	(void)fclose(f);
+	return PAM_SUCCESS;
+}
+
+void
+set_conf_opt(pam_handle_t *pamh, struct options *opts, const char *name,
+			 const char *value)
+{
+	if (strcmp(name, "dir") == 0) {
+		if (opts->dir == NULL) {
```
t8m, line 174: There is a small behavioral change introduced here. If there were multiple dir options before, the last one was used. Now the first one is used which is slightly counterintuitive, so I am not sure if this change is a good idea.

ikerexxe: I see your point, but why would anybody want to define several directories when only one of them would be used?

t8m: By a mistake, or for example due to some scripting.

However that is not important, the important thing is that currently we allow such configuration and by changing the semantics we could trigger a serious configuration issue.

modules/pam_faillock/main.c
```diff
@@ -51,32 +51,44 @@
...
-};
+#include "faillock_config.h"
 
 static int
 args_parse(int argc, char **argv, struct options *opts)
 {
 	int i;
+	int rv;
+	const char *conf = NULL;
+
```
t8m, line 64:
```suggestion
    const char *dir = NULL;

```

modules/pam_faillock/faillock_config.c
ldv-alt: There is no need to conditionalize on `opts->dir != NULL` to do `free(opts->dir)`.

modules/pam_faillock/main.c
```diff
@@ -94,6 +106,12 @@ args_parse(int argc, char **argv, struct options *opts)
 			return -1;
 		}
 	}
+
+	if ((rv = read_config_file(NULL, opts, conf)) != PAM_SUCCESS) {
+		fprintf(stderr, "Configuration file missing or broken");
+		return rv;
+	}
+
```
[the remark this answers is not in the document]
ldv-alt: There is no need to conditionalize on `opts->dir != NULL` to do `free(opts->dir)`.

modules/pam_faillock/main.c
ikerexxe: Fixed

modules/pam_faillock/main.c
ikerexxe: Done

modules/pam_faillock/faillock_config.c
ikerexxe: I changed this part of the code but not exactly as you suggested. Can you take a look at them?

modules/pam_faillock/main.c
```diff
@@ -51,32 +51,44 @@
...
+			if (i >= argc || strlen(argv[i]) == 0) {
+				fprintf(stderr, "%s: No records directory supplied.\n",
+						argv[0]);
+				return -1;
+			}
+			opts->dir = strdup(argv[i]);
+			if (opts->dir == NULL) {
+				fprintf(stderr, "Error allocating memory: %m");
 				return -1;
 			}
```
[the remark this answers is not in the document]
ikerexxe: Done

modules/pam_faillock/main.c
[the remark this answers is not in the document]
ikerexxe: Done

modules/pam_faillock/main.c
ikerexxe: Done

ikerexxe:
Thank you for the review. I have taken your comments into account to provide a revised version of the patches.

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,266 @@
...
+	}
+	else {
+		config_log(pamh, LOG_ERR, "Unknown option: %s", name);
+	}
+}
+
+const char *get_tally_dir(struct options *opts)
+{
+	return (opts->dir != NULL) ? opts->dir : FAILLOCK_DEFAULT_TALLYDIR;
+}
```
t8m, line 266: Missing EOL at EOF. Please add it.

ikerexxe: Fixed. Thanks for catching that. I don't know why my IDE has decided to stop doing that work automatically.

t8m (approved):

modules/pam_faillock/main.c
ikerexxe: I just squashed the commits and improved the commit message wording.

t8m (approved):

ldv-alt:
There are two Signed-off-by lines in the last commit, probably due to squashing.

ikerexxe:
> There are two Signed-off-by lines in the last commit, probably due to squashing.

Fixed.

> I'm not ready to review anything due to relocation, please wait a day or two if you want my review.

That's fine. May your move be smooth.

ikerexxe:
@ldv-alt I hope the relocation went smoothly.

Can you take a look at it?

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,266 @@
...
+#include <ctype.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <syslog.h>
+
+#include <security/pam_modules.h>
+
+#include "config.h"
```
ldv-alt, line 48:
```
#include "config.h"
```
should be the first `#include` because it affects processing of other header files.

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,266 @@
...
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#define _GNU_SOURCE
```
ldv-alt, line 37: Once `config.h` is included, here is no need for this because the project uses `AC_USE_SYSTEM_EXTENSIONS`.

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,266 @@
...
+#include "config.h"
+#include "faillock_config.h"
+#include "faillock.h"
+
+#define FAILLOCK_DEFAULT_CONF SCONFIGDIR "/faillock.conf"
+#ifdef VENDOR_SCONFIGDIR
+#define VENDOR_FAILLOCK_DEFAULT_CONF VENDOR_SCONFIGDIR "/faillock.conf"
+#endif
+
+static void PAM_FORMAT((printf, 3, 0)) PAM_NONNULL((3))
```
ldv-alt, line 57: The prototype of `pam_syslog` has `PAM_FORMAT((printf, 3, 4))` instead of this `PAM_FORMAT((printf, 3, 0))`.

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,266 @@
...
+config_log(const pam_handle_t *pamh, int priority, const char *fmt, ...)
+{
+	va_list args;
+	char *buf = NULL;
```
ldv-alt, line 61: Let's move the definition of `buf` inside the block where it's used.

modules/pam_faillock/faillock_config.c
```diff
@@ -0,0 +1,266 @@
...
+	}
+	else if (strcmp(name, "nodelay") == 0) {
+		opts->flags |= FAILLOCK_FLAG_NO_DELAY;
+	}
+	else {
+		config_log(pamh, LOG_ERR, "Unknown option: %s", name);
+	}
+}
+
+const char *get_tally_dir(struct options *opts)
```
ldv-alt, line 263: I suggest adding `const` qualifier to `struct options *` argument.

modules/pam_faillock/main.c
```diff
@@ -51,32 +51,41 @@
...
 {
 	int i;
+	int rv;
+	const char *dir = NULL;
+	const char *conf = NULL;
+
 	memset(opts, 0, sizeof(*opts));
 
-	opts->dir = FAILLOCK_DEFAULT_TALLYDIR;
+	opts->dir = NULL;
```
ldv-alt, line 68: `opts->dir = NULL` right after `memset(opts, 0, sizeof(*opts))` looks redundant.

modules/pam_faillock/pam_faillock.c
```diff
@@ -56,67 +55,24 @@
...
 	int i;
 	int config_arg_index = -1;
 	int rv;
-	const char *conf = default_faillock_conf;
+	const char *conf = NULL;
 
 	memset(opts, 0, sizeof(*opts));
 
-	opts->dir = strdup(FAILLOCK_DEFAULT_TALLYDIR);
+	opts->dir = NULL;
```
ldv-alt, line 75: `opts->dir = NULL` right after `memset(opts, 0, sizeof(*opts))` looks redundant.

ldv-alt (changes requested):
I couldn't find anything wrong, just a few simple technical issues.

t8m (approved):

ldv-alt (approved):

PR #458, opened 2022-04-22
pam_motd: do not rely on all filesystems providing a filetype

pjessen:
…y, we wrongly relied on all filesystems providing a filetype. This fixes that by diverting to lstat() when we have no filetype.  To maintain MT safety, it wasn't possible to use lstat() in the scandir() filter function, so I have moved all of the filtering to an additional loop after scanning all the motd dirs.

I know I could do some initial filtering in the scandir(), but I find it better to keep the filtering in one place.

### DIFF
--- a/modules/pam_motd/pam_motd.c
+++ b/modules/pam_motd/pam_motd.c
@@ -166,11 +166,6 @@ static int compare_strings(const void *a, const void *b)
     }
 }
 
-static int filter_dirents(const struct dirent *d)
-{
-    return (d->d_type == DT_REG || d->d_type == DT_LNK);
-}
-
 static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
 	char **motd_dir_path_split, unsigned int num_motd_dirs, int report_missing)
 {
@@ -199,8 +194,7 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
 
     for (i = 0; i < num_motd_dirs; i++) {
 	int rv;
-	rv = scandir(motd_dir_path_split[i], &(dirscans[i]),
-		filter_dirents, alphasort);
+	rv = scandir(motd_dir_path_split[i], &(dirscans[i]), NULL, NULL);
 	if (rv < 0) {
 	    if (errno != ENOENT || report_missing) {
 		pam_syslog(pamh, LOG_ERR, "error scanning directory %s: %m",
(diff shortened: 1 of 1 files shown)


modules/pam_motd/pam_motd.c
```diff
@@ -199,8 +194,7 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
 
     for (i = 0; i < num_motd_dirs; i++) {
 	int rv;
-	rv = scandir(motd_dir_path_split[i], &(dirscans[i]),
-		filter_dirents, alphasort);
+        rv = scandir(motd_dir_path_split[i], &(dirscans[i]), NULL, NULL);
```
ldv-alt, line 197: What was the rationale behind this `alphasort` removal?

modules/pam_motd/pam_motd.c
```diff
@@ -215,6 +209,45 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
...
+            case DT_LNK:              /* symlinks              */
+		drop=0;               /* are good.             */
+		break;
+            case DT_UNKNOWN:   /* for file systems that do not provide */
+			       /* a filetype, we use lstat()           */
+	        if ( join_dir_strings( &fullpath, motd_dir_path_split[i],
+				       d[j]->d_name )<=0 ) break;
+                rc=lstat( fullpath, &s );
+                _pam_drop(fullpath);  /* free the memory alloc'ed by join_dir_strings */
+                if ( rc!=0 ) break;   /* if the lstat() somehow failed */
```
ldv-alt, line 234: Please use consistent indentation.

Please follow the style used in the file being patched, e.g.
`=`, `!=`, and `<=` should be surrounded by spaces,
the should be no space after `(` and before `)`,
there should be a space after `switch` keyword,
`break;` statement should be placed on a separate line.

modules/pam_motd/pam_motd.c
```diff
@@ -215,6 +209,45 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
...
+
+		if ( S_IFREG==(s.st_mode&S_IFMT) ||         /* regular files and  */
+                     S_IFLNK==(s.st_mode&S_IFMT) ) drop=0;  /* symlinks are good  */
```
ldv-alt, line 237: Please use `S_ISREG` and `S_ISLNK` instead.

ldv-alt (changes requested):
Also, could you write a proper commit message, please?

modules/pam_motd/pam_motd.c
pjessen: That `alphasort` was always superfluous, the final sorting happens later anyway, by `qsort`.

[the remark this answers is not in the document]
pjessen: oh, I'm sorry, I thought spaces were preferred over tabs. Dunno why.

modules/pam_motd/pam_motd.c
pjessen: Good point, tnx.

modules/pam_motd/pam_motd.c
pjessen: Understood.

modules/pam_motd/pam_motd.c
```diff
@@ -215,6 +209,45 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
     if (dirscans_size_total == 0)
         goto out;
 
+    /* filter out unwanted names, directories, and complement data with lstat() */
+    for (i = 0; i < num_motd_dirs; i++) {
+        unsigned int j;
+        int rc, drop;
+        struct dirent **d;
+	char *fullpath;
+	struct stat s;
```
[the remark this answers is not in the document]
pjessen: right, understood.

modules/pam_motd/pam_motd.c
```diff
@@ -215,6 +209,47 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
     if (dirscans_size_total == 0)
         goto out;
 
+    /* filter out unwanted names, directories, and complement data with lstat() */
+    for (i = 0; i < num_motd_dirs; i++) {
+	unsigned int j;
+	int rc, drop;
+	struct dirent **d;
+	char *fullpath;
+	struct stat s;
```
ldv-alt, line 218: Since most of these variables are used inside the inner loop over `j` variable,
it would be better to move them inside that loop.

modules/pam_motd/pam_motd.c
```diff
@@ -215,6 +209,47 @@ static void try_to_display_directories_with_overrides(pam_handle_t *pamh,
...
+
+	d = dirscans[i];
+	for (j = 0; j < dirscans_sizes[i]; j++) {
+	    drop = 1;
+	    switch(d[j]->d_type) {    /* the filetype determines how to proceed */
+	    case DT_REG:              /* regular files and     */
+	    case DT_LNK:              /* symlinks              */
+		drop = 0;             /* are good.             */
```
ldv-alt, line 226: I think we would better off without `drop` variable at all.
For example, if `drop = 0` is replaced with `continue`, then
there would be no need to assign anything to `drop` and check it afterwards.

ldv-alt (changes requested):
Could you follow `Commit log requirements` listed in README-hacking file, please?

ldv-alt:
Also please rebase on top of `master` branch.

modules/pam_motd/pam_motd.c
pjessen: thanks for that, that is a good idea.

modules/pam_motd/pam_motd.c
pjessen: Okay, agree.

ldv-alt (approved):

t8m (approved):

PR #470, opened 2022-06-29
faillock: add support to print login failure info in legacy format

sshedi:
pam_tally2 had a simple and minimalstic output to show login failure
info, new output of faillock makes the output look a bit complex and
doesn't show failure counts in a straight manner.

Also if user account doesn't exist faillock doesn't error out.
For first time successful login, failure field should be 0 and faillock
should show something like pam_tally2 did.

This patch fixes the above issue by adding "--sfos" flag to faillock
which makes it possible to get output in pam_tally2 style.

Signed-off-by: Shreenidhi Shedi <sshedi@vmware.com>


sshedi:
> This is not a proper review, just a remark: a new option should be either documented or self-explanatory, but `--sfos` is neither.

Thanks for the input. I have changed it to `--legacy-output`
I believe this change will help many linux-pam users who are moving to newer versions.

modules/pam_faillock/main.c
```diff
@@ -123,8 +126,73 @@ args_parse(int argc, char **argv, struct options *opts)
...
 {
-	fprintf(stderr, _("Usage: %s [--dir /path/to/tally-directory] [--user username] [--reset]\n"),
-		progname);
+	fprintf(stderr,
+		_("Usage: %s [--dir /path/to/tally-directory] "
+		  " [--user username] [--reset] [--legacy-output]\n"), progname);
+}
+
+static void
+print_in_new_format(struct options *opts, struct tally_data tallies, const char *user)
```
ldv-alt, line 135: Why the structure is passed to this function not by reference but by value?

modules/pam_faillock/main.c
```diff
@@ -123,8 +126,73 @@ args_parse(int argc, char **argv, struct options *opts)
...
+{
+	uint32_t i;
+	time_t when;
+	struct tm *tm;
+	uint16_t status;
+	char timebuf[80] = {};
```
ldv-alt, line 141: Why the buffer is initialized here?  It was not the case in the old code.

If you are refactoring this, consider moving into the loop all variables that are initialized and used only inside that loop.

modules/pam_faillock/main.c
```diff
@@ -137,20 +205,23 @@ do_user(struct options *opts, const char *user)
 	const char *dir = get_tally_dir(opts);
 
 	pwd = getpwnam(user);
+	if (!pwd) {
+	    fprintf(stderr, "%s: getpwnam; no such user: %s\n", opts->progname, user);
+	    return 1;
+	}
```
ldv-alt, line 211: If this is a fix, it should rather go to a separate commit.

Nitpicking: why getpwnam in the diagnostics is followed by a semicolon?

ldv-alt (changes requested):
I suggest splitting this into three commits: a fix, a refactoring that introduces print_in_new_format, and the new feature.

sshedi:
> I suggest splitting this into three commits: a fix, a refactoring that introduces print_in_new_format, and the new feature.

Understood, thanks for the review.

modules/pam_faillock/main.c
sshedi: We don't modify anything in that structure. We just need the data it has. So it doesn't matter if we call it by value or reference. It's like a read only member.

modules/pam_faillock/main.c
sshedi: Will do, thanks. better to initialize variables before passing them to other functions.

modules/pam_faillock/main.c
sshedi: Done.

modules/pam_faillock/main.c
ldv-alt: While passing structures by value is technically possible, it is not something people do when they need to pass a structure for reading.  A structure passed by value raises the question: what is so specific about the case that requires a structure to be passed by value.

modules/pam_faillock/main.c
ldv-alt: Buffers that are not expected to be accessed before being written are not required to be initialized.
When I see a buffer being initialized this way, my first thought is where the buffer is accessed before being written.

modules/pam_faillock/main.c
```diff
@@ -123,8 +126,80 @@ args_parse(int argc, char **argv, struct options *opts)
...
 {
-	fprintf(stderr, _("Usage: %s [--dir /path/to/tally-directory] [--user username] [--reset]\n"),
-		progname);
+	fprintf(stderr,
+		_("Usage: %s [--dir /path/to/tally-directory] "
+		  " [--user username] [--reset] [--legacy-output]\n"), progname);
+
+}
+
+static int get_local_time(time_t when, char *timebuf, unsigned long timebuf_size)
```
ldv-alt, line 135: Please consider using `size_t` instead.

Also note that at least in this file the type of function is placed on a separate line.

modules/pam_faillock/main.c
```diff
@@ -123,8 +126,80 @@ args_parse(int argc, char **argv, struct options *opts)
...
+{
+	struct tm *tm;
+
+	tm = localtime(&when);
+	if(tm == NULL) {
```
ldv-alt, line 140: Even if `if(tm == NULL) ` comes from an old code, the absence of space after `if` looks odd here.

ldv-alt (changes requested):

modules/pam_faillock/main.c
sshedi: Okay, I will make it pass by reference. Thanks.

modules/pam_faillock/main.c
sshedi: Yeah, missed it. Will do.

modules/pam_faillock/main.c
```diff
@@ -123,8 +126,84 @@ args_parse(int argc, char **argv, struct options *opts)
...
+		  " [--user username] [--reset] [--legacy-output]\n"), progname);
+
+}
+
+static int
+get_local_time(time_t when, char *timebuf, size_t timebuf_size)
+{
+	struct tm *tm;
+
+	tm = localtime(&when);
```
ldv-alt, line 140: You could have merged these two lines, but that's not necessary.

ldv-alt (approved):

ldv-alt:
@t8m Do you have any comments about this PR?

t8m (approved):
Maybe it could be called a `--terse-output` instead of  legacy but I am OK with it as is too.

modules/pam_faillock/main.c
```diff
@@ -123,8 +126,84 @@ args_parse(int argc, char **argv, struct options *opts)
...
+	if (tm == NULL) {
+		return -1;
+	}
+	strftime(timebuf, timebuf_size, "%Y-%m-%d %H:%M:%S", tm);
+	return 0;
+}
+
+static void
+print_in_new_format(struct options *opts, struct tally_data *tallies, const char *user)
```
ldv-alt, line 149: Sorry, I've overlooked this, it should be `const struct tally_data *tallies`.

modules/pam_faillock/main.c
```diff
@@ -123,8 +126,84 @@ args_parse(int argc, char **argv, struct options *opts)
...
+		status = tallies->records[i].status;
+
+		printf("%-19s %-5s %-52.52s %s\n", timebuf,
+			status & TALLY_STATUS_RHOST ? "RHOST" : (status & TALLY_STATUS_TTY ? "TTY" : "SVC"),
+			tallies->records[i].source, status & TALLY_STATUS_VALID ? "V":"I");
+	}
+}
+
+static void
+print_in_legacy_format(struct options *opts, struct tally_data *tallies, const char *user)
```
ldv-alt, line 175: Likewise, it should be `const struct tally_data *tallies`.

ldv-alt (changes requested):

sshedi:
> Maybe it could be called a `--terse-output` instead of legacy but I am OK with it as is too.

I think is `--legacy-output` easy to remember. So, will keep it. Thanks for the review.

sshedi:
Thank you @ldv-alt and @t8m for your time and patience.

PR #473, opened 2022-07-08
pam_pwhistory: load config from file

ikerexxe:
I have added a new `conf` option in pam_pwhistory to select the configuration file to read. The default one being `/etc/security/pwhistory.conf`. On top of that I have used `pam_modutil_search_key()` function to read the pwhistory configuration file. This configuration is then loaded in the corresponding `options` structure.

Besides, I have documented the new `conf` option and the configuration file itself.

Finally, I have improved the `pam_modutil_search_key()` documentation.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2068461

### DIFF
--- a/libpam/include/security/pam_modutil.h
+++ b/libpam/include/security/pam_modutil.h
@@ -147,7 +147,16 @@ pam_modutil_sanitize_helper_fds(pam_handle_t *pamh,
 				enum pam_modutil_redirect_fd redirect_stdout,
 				enum pam_modutil_redirect_fd redirect_stderr);
 
-/* lookup a value for key in login.defs file or similar key value format */
+/**************************************************
+ * @brief Lookup a value for the key in the file (i.e. login.defs or a similar
+ * key-value format file).
+ *
+ * @param[in] pamh The pam handle structure
+ * @param[in] file_name Configuration file name
+ * @param[in] key Lookup key
+ *
+ * @return value, or NULL if key was not found.
+ **************************************************/
 extern char * PAM_NONNULL((1,2,3))
 pam_modutil_search_key(pam_handle_t *pamh,
 		       const char *file_name,
--- a/modules/pam_pwhistory/pwhistory_config.h
+++ b/modules/pam_pwhistory/pwhistory_config.h
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2022 Iker Pedrosa <ipedrosa@redhat.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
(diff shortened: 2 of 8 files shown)


thkukuk:
With pam_modutil_search_key we have already something similar...
Instead of re-inventing the wheel again and again, why not using one library doing the job? Especially as the new code is really unflexible. More and more projects allow to split config files in smaller parts or allow to override single values, so that you can separate between the vendor provided configuration file and the single changes the admin made. And since we have that for PAM already in many places, why should we introduce something again, which does not support this?
I'm strongly against adding something which limits us again.

We have code to use libeconf for reading login.defs, we have code using libeconf to read other configuration files, maybe it's time to consolidate on one best practices and one tool?

thkukuk:
Or in other words: are we 100% sure that we can use this generic parser for all current modules without the need to introduce another interface or another variant?
From my experience with this topic from libeconf, I heavily doubt that.

ikerexxe:
> With pam_modutil_search_key we have already something similar... Instead of re-inventing the wheel again and again, why not using one library doing the job?

You are absolutely right. I should have checked the existing APIs with more care before implementing a new one. Thanks for pointing me in the right direction @thkukuk.

I have updated the PR to use `pam_modutil_search_key()` and it's ready for review.

thkukuk:
Thanks. Looks good to me.

libpam/include/security/pam_modutil.h
```diff
@@ -147,7 +147,16 @@ pam_modutil_sanitize_helper_fds(pam_handle_t *pamh,
...
-/* lookup a value for key in login.defs file or similar key value format */
+/**************************************************
+ * @brief Lookup a value for key in a file (i.e. login.defs or similar
+ * key-value format file).
+ *
+ * @param[in] pamh The pam handle structure
+ * @param[in] file_name Configuration file name
+ * @param[in] key Lookup key
+ *
+ * @return value, or NULL if key was not found.
```
ldv-alt, line 158: I suggest adding some English grammar to this description, e.g.
`for the key in the file` and `a similar`.

modules/pam_pwhistory/pwhistory_config.h
```diff
@@ -0,0 +1,54 @@
...
+    int tries;
+    const char *filename;
+};
+typedef struct options_t options_t;
+
+void
+parse_config_file(pam_handle_t *pamh, int argc, const char **argv,
+                  struct options_t *options);
+
+#endif /* _PWHISTORY_CONFIG_H */
```
ldv-alt, line 54: No newline at end of file.

ldv-alt (changes requested):
From a cursory look, there are just a few very minor issues.

ikerexxe:
I fixed those issues and the PR is ready for review again.

Issue #475, opened 2022-07-12
pam_handlers: possible SEGV caused by not initialized next member of the list


LexPanov:
While calling `pam_start` and using some invalid paths in the configuration file included using include/substack, a pointer of a high value address is dereferenced.

### /etc/pam.d/crash_conf

One of the following lines:

```
0 include crash_include
0 substack crash_include
```

### /etc/pam.d/crash_include

One of the following lines:

```
0 0 /
0 0 .
0 0 ?
0 0 [
```

### main.c

```c
#include <security/pam_appl.h>
#include <security/pam_misc.h>
#include <stdio.h>

static struct pam_conv conv = {misc_conv, NULL};

int main() {
  pam_handle_t *pamh = NULL;
  int retval;
  const char *user = "user";

  retval = pam_start("crash_conf", user, &conv, &pamh);
  printf("pam_start: %d\n", retval);
  pam_end(pamh, retval);

  return 0;
}
```

### ASAN output

```
$ gcc main.c -o main -lpam -lpam_misc
$ ./main
AddressSanitizer:DEADLYSIGNAL
=================================================================
==161965==ERROR: AddressSanitizer: SEGV on unknown address (pc 0x00000031627f bp 0x7ffe219961f0 sp 0x7ffe21996100 T0)
==161965==The signal is caused by a READ memory access.
==161965==Hint: this fault was caused by a dereference of a high value address (see register values below).  Disassemble the provided pc to learn which register was used.
    #0 0x31627f in _pam_add_handler /tmp/Linux-PAM/libpam/pam_handlers.c:888:12
    #1 0x313dbc in _pam_parse_conf_file /tmp/Linux-PAM/libpam/pam_handlers.c:264:12
    #2 0x30e712 in _pam_init_handlers /tmp/Linux-PAM/libpam/pam_handlers.c:462:12
    #3 0x3075dd in pam_start /tmp/Linux-PAM/libpam/pam_start.c:124:10
    #4 0x3075dd in main /tmp/Linux-PAM/main.c:12:12
    #5 0x7f212c7b90b2 in __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:308:16
    #6 0x254bed in _start (/tmp/Linux-PAM/main+0x25864d)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /tmp/Linux-PAM/libpam/pam_handlers.c:888:12 in _pam_add_handler
[...]

PR #476, opened 2022-07-12
pam_handlers: fix next member of the list init

LexPanov:
Re-entering _pam_add_handler without initialization of the next member of
the list leads to SEGV caused by a dereference of a high value address.

Resolves: linux-pam/linux-pam#475

### DIFF
--- a/libpam/pam_handlers.c
+++ b/libpam/pam_handlers.c
@@ -902,10 +902,10 @@ int _pam_add_handler(pam_handle_t *pamh
     (*handler_p)->cached_retval_p = &((*handler_p)->cached_retval);
     (*handler_p)->argc = argc;
     (*handler_p)->argv = argv;                       /* not a copy */
-    if (((*handler_p)->mod_name = extract_modulename(mod_path)) == NULL)
-	return PAM_ABORT;
     (*handler_p)->grantor = 0;
     (*handler_p)->next = NULL;
+    if (((*handler_p)->mod_name = extract_modulename(mod_path)) == NULL)
+	return PAM_ABORT;
 
     /* some of the modules have a second calling function */
     if (handler_p2) {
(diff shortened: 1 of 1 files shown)


libpam/pam_handlers.c
```diff
@@ -936,10 +936,10 @@ int _pam_add_handler(pam_handle_t *pamh
 	} else {
 	    (*handler_p2)->argv = NULL;              /* no arguments */
 	}
-	if (((*handler_p2)->mod_name = extract_modulename(mod_path)) == NULL)
-	    return PAM_ABORT;
 	(*handler_p2)->grantor = 0;
 	(*handler_p2)->next = NULL;
+	if (((*handler_p2)->mod_name = extract_modulename(mod_path)) == NULL)
+	    return PAM_ABORT;
```
ldv-alt, line 942: This part of the change seems to be incomplete: there is another error path where these fields remain uninitialized.

ldv-alt (changes requested):
Good catch.
Please update the patch to address the issue with its second hunk.

libpam/pam_handlers.c
LexPanov: What do you think needs to be done with this part of the code?  I found it necessary to patch this too because this code is very similar to the one above.  (Although I haven't explored how to get to this part of the code)

t8m: I'd just use calloc() for the *handler_p2 allocation so it is initialized with zeros.

libpam/pam_handlers.c
```diff
@@ -902,10 +902,10 @@ int _pam_add_handler(pam_handle_t *pamh
     (*handler_p)->cached_retval_p = &((*handler_p)->cached_retval);
     (*handler_p)->argc = argc;
     (*handler_p)->argv = argv;                       /* not a copy */
-    if (((*handler_p)->mod_name = extract_modulename(mod_path)) == NULL)
-	return PAM_ABORT;
     (*handler_p)->grantor = 0;
     (*handler_p)->next = NULL;
+    if (((*handler_p)->mod_name = extract_modulename(mod_path)) == NULL)
+	return PAM_ABORT;
```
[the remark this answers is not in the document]
t8m: Or maybe better just use calloc to initialize the struct when allocating it.

ldv-alt:
On Wed, Jul 13, 2022 at 12:36:45AM -0700, Tomáš Mráz wrote:
> @t8m commented on this pull request.
> 
> > -	if (((*handler_p2)->mod_name = extract_modulename(mod_path)) == NULL)
> -	    return PAM_ABORT;
>  	(*handler_p2)->grantor = 0;
>  	(*handler_p2)->next = NULL;
> +	if (((*handler_p2)->mod_name = extract_modulename(mod_path)) == NULL)
> +	    return PAM_ABORT;
> 
> I'd just use calloc() for the *handler_p2 allocation so it is initialized with zeros.

Yes, calloc() is another viable option.

PR #484, opened 2022-08-19
[RFC] pam_unix: use helper on skeleton root entry from libnss-systemd

cgzones:
libnss-systemd started[1] to synthesize a root entry with an invalid
password.  On Debian since version 251.3-2[2] systemd enables the
systemd module for shadow and gsshadow in nsswitch.conf.  This leads on
SELinux enabled systems, where the files module can not access the
shadow file, to the root account being locked.

Fall back to the unix_chkpwd helper in case the root user was queried
and the response was a locked password ("!*"[3]).

See also https://github.com/systemd/systemd/issues/20299 for previous issue leading to https://github.com/linux-pam/linux-pam/commit/470823c4aacef5cb3b1180be6ed70846b61a3752.

[1]: https://github.com/systemd/systemd/commit/f43a19ecd6e3415e
[2]: https://salsa.debian.org/systemd-team/systemd/-/commit/bf9a307cfcbe62ab4fbcf2198e6e628a1bca211b
[3]: https://github.com/systemd/systemd/blob/b622e95f2f59fcb58e23ddafed745eee26a0f52f/src/nss-systemd/nss-systemd.c#L34

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -242,7 +242,9 @@ PAMH_ARG_DECL(int get_account_info,
 			 * if shadowing is enabled
 			 */
 			*spwdent = pam_modutil_getspnam(pamh, name);
-			if (*spwdent == NULL) {
+			if (*spwdent == NULL ||
+			    /* synthesized entry from libnss-systemd */
+			    (strcmp(name, "root") == 0 && (*spwdent)->sp_pwdp && strcmp((*spwdent)->sp_pwdp, "!*") == 0)) {
 #ifndef HELPER_COMPILE
 				/* still a chance the user can authenticate */
 				return PAM_UNIX_RUN_HELPER;


ldv-alt:
Since both the module and its helper use the same `pam_modutil_getspnam` interface, I wonder why such a change could have a desired effect.

ldv-alt:
And both are essentially the same thing, so is there any observable difference?

cgzones:
> Since both the module and its helper use the same pam_modutil_getspnam interface, I wonder why such a change could have a desired effect.

The helper is allowed to read */etc/shadow* (by SELinux), while the pam module caller is probably not, thus the first module `files` returns an entry without the latter `systemd` one being queried.

ldv-alt:
If that's the case, I'd suggest to change the module to use the helper unconditionally without even trying to use `pam_modutil_getspnam` when SELinux is enabled.

cgzones:
Wouldn't that defeat the whole purpose of nss?

An alternative would be to change `getenv_bool_secure()` to `getenv_bool()` at https://github.com/systemd/systemd/blob/a13231a904eb192db34cd58774ee8dd9e14d2c9e/src/nss-systemd/nss-systemd.c#L405, so then pam can export `SYSTEMD_NSS_BYPASS_SYNTHETIC=1` to avoid the synthesized entry (since currently the underlying `secure_getenv(3)` does not work with suid binaries and also with SELinux).

ldv-alt:
> Wouldn't that defeat the whole purpose of nss?

What would defeat the whole purpose of nss is a special handling of things like libnss-systemd in a generic code.
Just imagine how many different `getspnam` users exist in the world.

cgzones:
It's true, the added workaround specific for libnss-systemd is a bit messy, but circumventing nss entirely (e.g. ldap module) might break setups.

thkukuk:
| but circumventing nss entirely (e.g. ldap module) might break setups.
If SELinux is enabled and pam_unix calls directly the helper application, why should that circumvent nss entirely and might break setups?
Currently you have two different behaviors, depending on if your application is allowed to read the file or not. Everybody who has a setup depending on this should fix his setup anyways, because every SELinux policy update can also break his setup. But I doubt that this will break any setup, else this setup would be broken anyways for some applications who are allowed to read the shadow file.
PAM should not have workarounds or special hacks for random NSS modules or applications.

PR #485, opened 2022-08-21
po: Georgian translation

NorwayFun:
po: Georgian translation

In case it's needed now.


ldv-alt:
We normally take translations from https://translate.fedoraproject.org/projects/linux-pam/master/,
could you try submitting your translation there, please?

PR #486, opened 2022-08-23
security: Add landlock file restriction

nbouchinet-anssi:
This patch adds a minimalist Landlock support to PAM which restricts the
SUID unix_chkpwd binary to a strict readonly access to /etc/shadow. The
goal here is to provide security in depth limiting access to the few
resources this program needs limiting uncontrolled privileged access to
the system.

The patch should evolve to support finer restrictions on other PAM
modules/binaries.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -528,6 +528,26 @@ fi
 
 ECONF_CFLAGS=
 ECONF_LIBS=
+AC_ARG_ENABLE([landlock],
+    AS_HELP_STRING([--enable-landlock],[Enable Landlock sandbox]),
+    [WITH_LANDLOCK=$enableval], WITH_LANDLOCK=no)
+if test "$WITH_LANDLOCK" = "yes" ; then
+  AC_CHECK_DECL([__NR_landlock_create_ruleset], [], [], [#include <sys/syscall.h>])
+  AC_CHECK_DECL([__NR_landlock_add_rule], [], [], [#include <sys/syscall.h>])
+  AC_CHECK_DECL([__NR_landlock_restrict_self], [], [], [#include <sys/syscall.h>])
+  AC_CHECK_HEADER(linux/landlock.h, [have_landlock_header=1], [have_landlock_header=0])
+  if test "x$ac_cv_have_decl___NR_landlock_create_ruleset" = xyes &&
+     test "x$ac_cv_have_decl___NR_landlock_add_rule" = xyes &&
+     test "x$ac_cv_have_decl___NR_landlock_restrict_self" = xyes &&
+     test "$have_landlock_header" = 1; then
+    AC_DEFINE([WITH_LANDLOCK], 1, [Defined if Landlock sandbox support is compiled in])
+  else
+    WITH_LANDLOCK=no
+  fi
+else
+  WITH_LANDLOCK=no
+fi
+
 AC_ARG_ENABLE([econf],
   AS_HELP_STRING([--disable-econf], [do not use libeconf]),
   [WITH_ECONF=$enableval], [WITH_ECONF=yes])
@@ -710,6 +730,7 @@ AM_CONDITIONAL([COND_BUILD_PAM_SETQUOTA], [test "$ac_cv_func_quotactl" = yes])
 AM_CONDITIONAL([COND_BUILD_PAM_TTY_AUDIT], [test "$HAVE_AUDIT_TTY_STATUS" = yes])
 AM_CONDITIONAL([COND_BUILD_PAM_UNIX], [test "$enable_unix" = yes])
 AM_CONDITIONAL([COND_BUILD_PAM_USERDB], [test -n "$LIBDB"])
+AM_CONDITIONAL([COND_BUILD_LANDLOCK], [test "$WITH_LANDLOCK" = yes])
 
 dnl Files to be created from when we run configure
 AC_CONFIG_FILES([Makefile libpam/Makefile libpamc/Makefile libpamc/test/Makefile \
--- a/libpam/include/security/pam_landlock.h
+++ b/libpam/include/security/pam_landlock.h
@@ -1,0 +1,104 @@
+#include "_pam_macros.h"
+
+#include <errno.h>
+#include <fcntl.h>
+#include <linux/landlock.h>
+#include <linux/prctl.h>
+#include <stddef.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/prctl.h>
+#include <sys/stat.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#ifndef _SECURITY_PAM_LANDLOCK_H
+#define _SECURITY_PAM_LANDLOCK_H
+
+#ifndef landlock_create_ruleset
+static inline int
+landlock_create_ruleset(const struct landlock_ruleset_attr *const attr,
+		const size_t size, const __u32 flags)
+{
+	return syscall(__NR_landlock_create_ruleset, attr, size, flags);
+}
+#endif
+
+#ifndef landlock_add_rule
+static inline int landlock_add_rule(const int ruleset_fd,
+		const enum landlock_rule_type rule_type,
+		const void *const rule_attr,
+		const __u32 flags)
+{
+	return syscall(__NR_landlock_add_rule, ruleset_fd, rule_type, rule_attr,
+			flags);
+}
+#endif
+
+#ifndef landlock_restrict_self
+static inline int landlock_restrict_self(const int ruleset_fd,
+		const __u32 flags)
+{
+	return syscall(__NR_landlock_restrict_self, ruleset_fd, flags);
+}
+#endif
+
+#ifndef LANDLOCK_ACCESS_FS_REFER
+#define LANDLOCK_ACCESS_FS_REFER (1ULL << 13)
+#endif
+
+#define ACCESS_FS_ROUGHLY_WRITE ( \
+		LANDLOCK_ACCESS_FS_WRITE_FILE | \
+		LANDLOCK_ACCESS_FS_REMOVE_DIR | \
+		LANDLOCK_ACCESS_FS_REMOVE_FILE | \
+		LANDLOCK_ACCESS_FS_MAKE_CHAR | \
+		LANDLOCK_ACCESS_FS_MAKE_DIR | \
+		LANDLOCK_ACCESS_FS_MAKE_REG | \
+		LANDLOCK_ACCESS_FS_MAKE_SOCK | \
+		LANDLOCK_ACCESS_FS_MAKE_FIFO | \
+		LANDLOCK_ACCESS_FS_MAKE_BLOCK | \
+		LANDLOCK_ACCESS_FS_MAKE_SYM | \
+		LANDLOCK_ACCESS_FS_REFER)
+
+#define LANDLOCK_COMPAT_ERR -1
+#define LANDLOCK_FATAL_ERR 1
+
+static int _pam_landlock_create_ruleset(int * const ruleset_fd, const __u64 ruleset) {
+	int abi;
+
+	struct landlock_ruleset_attr ruleset_attr = {
+		.handled_access_fs = ruleset,
+	};
+
+	/* Checks for Landlock ABI compatibility. */
+	abi = landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION);
+	if (abi < 0)
+		return LANDLOCK_COMPAT_ERR;
+	if (abi < 2)
+		ruleset_attr.handled_access_fs &= ~LANDLOCK_ACCESS_FS_REFER;
+
+	*ruleset_fd = landlock_create_ruleset(&ruleset_attr, sizeof(ruleset_attr), 0);
+	if (*ruleset_fd < 0) {
+		D(("Failed to create a ruleset"));
+		return LANDLOCK_FATAL_ERR;
+	}
+	return 0;
+}
+
+static int _pam_landlock_apply_restrictions(const int ruleset_fd) {
+	if (prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0)) {
+		D(("Failed to restrict privileges"));
+		close(ruleset_fd);
+		return LANDLOCK_FATAL_ERR;
+	}
+	if (landlock_restrict_self(ruleset_fd, 0)) {
+		D(("Failed to enforce ruleset"));
+		close(ruleset_fd);
+		return LANDLOCK_FATAL_ERR;
+	}
+	close(ruleset_fd);
+	return 0;
+}
+
+#endif /* _SECURITY_PAM_LANDLOCK_H */
--- a/modules/pam_unix/unix_chkpwd.c
+++ b/modules/pam_unix/unix_chkpwd.c
@@ -32,6 +32,10 @@
 #include <security/_pam_types.h>
 #include <security/_pam_macros.h>
 
+#ifdef WITH_LANDLOCK
+#include <security/pam_landlock.h>
+#endif
+
 #include "passverify.h"
 #include "pam_inline.h"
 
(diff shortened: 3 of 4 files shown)


ldv-alt:
This binary uses `getspnam_r` interface to obtain the shadow entry, this doesn't necessarily imply access to `/etc/shadow`.
Would it be possible to enforce a read-only access to the whole filesystem?

ldv-alt:
By the way, your patch results to a code that doesn't even compile, this is not a good sign at all.

l0kod:
I guess it misses a check for the header files (e.g. [Suricata](https://github.com/OISF/suricata/pull/7697/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R416)).

libpam/include/security/pam_sandbox.h
```diff
@@ -0,0 +1,136 @@
...
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#ifndef landlock_create_ruleset
+
+	static inline int
+landlock_create_ruleset(const struct landlock_ruleset_attr *const attr,
+		const size_t size, const __u32 flags)
+{
+	return syscall(__NR_landlock_create_ruleset, attr, size, flags);
```
l0kod, line 20: `__NR_landlock_create_ruleset` may not be provided by the libc headers if it is an old version.

libpam/include/security/pam_sandbox.h
```diff
@@ -0,0 +1,136 @@
...
+}
+#endif
+
+#define ACCESS_FS_ROUGHLY_READ ( \
+		LANDLOCK_ACCESS_FS_EXECUTE | \
+		LANDLOCK_ACCESS_FS_READ_FILE | \
+		LANDLOCK_ACCESS_FS_READ_DIR)
+
+#ifdef LANDLOCK_ACCESS_FS_REFER
+#define ACCESS_FS_ROUGHLY_WRITE ( \
```
l0kod, line 49: As a general note, we should only have one `ACCESS_FS_ROUGHLY_WRITE`, always including `LANDLOCK_ACCESS_FS_REFER`, but removing it at runtime if required. See https://git.kernel.org/pub/scm/linux/kernel/git/mic/linux.git/commit/?h=next&id=c773739d5362a8a3fc595c9851119562cb26f2c7

However, for this PAM use case, let's make it simple and just remove `LANDLOCK_ACCESS_FS_REFER` everywhere. Indeed, this access right is denied by default even if the ruleset doesn't handle it, and because this sandboxed PAM will not use rename(2) nor link(2), it should be fine.

libpam/include/security/pam_sandbox.h
```diff
@@ -0,0 +1,136 @@
...
+		LANDLOCK_ACCESS_FS_MAKE_BLOCK | \
+		LANDLOCK_ACCESS_FS_MAKE_SYM)
+#endif
+
+#define ACCESS_FILE_RW ( \
+		LANDLOCK_ACCESS_FS_WRITE_FILE | \
+		LANDLOCK_ACCESS_FS_READ_FILE)
+
+#define ACCESS_FILE_RO (LANDLOCK_ACCESS_FS_READ_FILE)
+
```
l0kod, line 83: extra newline

libpam/include/security/pam_sandbox.h
```diff
@@ -0,0 +1,136 @@
...
+	__u64 access_fs_rw = ACCESS_FS_ROUGHLY_READ | ACCESS_FS_ROUGHLY_WRITE;
+
+	struct landlock_ruleset_attr ruleset_attr = {
+		.handled_access_fs = access_fs_rw,
+	};
+	struct landlock_path_beneath_attr path_beneath = {
+		.allowed_access = ACCESS_FILE_RO,
+	};
+	int ruleset_fd, err;
+#ifdef LANDLOCK_ACCESS_FS_REFER
```
l0kod, line 96: `LANDLOCK_ACCESS_FS_REFER` may be defined or not by the libc headers, but what matter is what the running kernel supports. See the Suricata patches which defines `LANDLOCK_ACCESS_FS_REFER` if not already defined.

In a general case, this kind of ABI check should be performed all the time to make the code less complex and more maintainable.

Anyway, as said just before, for now there is no need to handle `LANDLOCK_ACCESS_FS_REFER`. This kind of ABI version check will be useful when Landlock will handle more interesting access rights (e.g. the `LANDLOCK_ACCESS_FS_TRUNCATE` one). The only thing required for now is to check that the running kernel support Landlock (i.e. `if (abi < 0) { return without error }`).

libpam/include/security/pam_sandbox.h
```diff
@@ -0,0 +1,136 @@
...
+
+	abi = landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION);
+	if (abi < 2) {
+		ruleset_attr.handled_access_fs &= ~ACCESS_ABI_2;
+		access_fs_rw &= ~ACCESS_ABI_2;
+	}
+#endif
+	ruleset_fd = landlock_create_ruleset(&ruleset_attr, sizeof(ruleset_attr), 0);
+	if (ruleset_fd < 0) {
+		perror("Failed to create a ruleset");
```
l0kod, line 107: I don't know where such error message goes but it may be more appropriate in a log (as an error because it should not happen).

libpam/include/security/pam_sandbox.h
```diff
@@ -0,0 +1,136 @@
...
+#endif
+
+#define ACCESS_FILE_RW ( \
+		LANDLOCK_ACCESS_FS_WRITE_FILE | \
+		LANDLOCK_ACCESS_FS_READ_FILE)
+
+#define ACCESS_FILE_RO (LANDLOCK_ACCESS_FS_READ_FILE)
+
+
+static int _restrict_to_file_read(const char * const file_path)
```
l0kod, line 85: This `_restrict_to_file_read()` doesn't look like a good design because I think it will lead to calling it multiple times, which will create extra nested sandboxes, which are limited to 16 layers for the whole system. Only one layer is needed here, and the API should follow the builder pattern (as do the Landlock syscalls), i.e. create a ruleset, add some rules, and finally enforce them.

modules/pam_unix/unix_chkpwd.c
```diff
@@ -103,6 +104,13 @@ int main(int argc, char *argv[])
 	 */
 	setup_signals();
 
+	/*
+	 * Use landlock to reduce the file scope of unix_chkpwd to /etc/shadow
+	 */
+
+	if (_restrict_to_file_read("/etc/shadow") > 0)
```
l0kod, line 111: Is it possible to open `/etc/shadow` before restricting the current thread and use this file descriptor after? This will simplify the whole thing by only restricting access to all files (opened FD are not taken into account).

modules/pam_unix/unix_chkpwd.c
```diff
@@ -103,6 +104,13 @@ int main(int argc, char *argv[])
...
+		return PAM_SYSTEM_ERR;
```
l0kod, line 112: I guess such error should not be returned if Landlock is not supported by the running kernel, only if some legitimate Landlock calls failed unexpectedly.

nbouchinet-anssi:
> By the way, your patch results to a code that doesn't even compile, this is not a good sign at all.

Hi, Thanks for the reply, sorry I indeed missed to add check for the header files

t8m:
> Hi, Thanks for the reply, sorry I indeed missed to add checks for the header files

It should be also optional as presence of the headers should not automatically mean that the support should be compiled in.

l0kod:
> It should be also optional as presence of the headers should not automatically mean that the support should be compiled in.

I'd suggest to enable sandboxing whenever possible to protect users. This could be an enabled-by-default configuration though, but I'm not sure this complexity is worth it because Landlock support must not break anything anyway.

libpam/include/security/pam_sandbox.h
nbouchinet-anssi: Seem's fair for this case but as the `pam_sandbox.h` header could be used in other modules shouldn't it integrate `LANDLOCK_ACCESS_FS_REF` checks anyway ?

libpam/include/security/pam_sandbox.h
```diff
@@ -0,0 +1,136 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <sys/prctl.h>
+#include <sys/stat.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#ifndef landlock_create_ruleset
+
+	static inline int
```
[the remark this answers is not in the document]
nbouchinet-anssi: ACK

libpam/include/security/pam_sandbox.h
nbouchinet-anssi: Right, will split to match the builder pattern.

libpam/include/security/pam_sandbox.h
nbouchinet-anssi: I've locally moved the libpam error prints to calls to the `D()` debug macro, the error is then handled in `unix_chkpwd` using the `libaudit` and `helper_log_err()` functions.

Maybe @t8m and @ldv-alt could point me to a more appropriate way to handle those error messages ?

libpam/include/security/pam_sandbox.h
nbouchinet-anssi: Thank's I now use the suricata way to handle this.

libpam/include/security/pam_sandbox.h
nbouchinet-anssi: Anyway I'll be able to handle error more finely after this https://github.com/linux-pam/linux-pam/pull/486#discussion_r953669996 recommendation will be applied.

nbouchinet-anssi:
> This binary uses `getspnam_r` interface to obtain the shadow entry, this doesn't necessarily imply access to `/etc/shadow`. Would it be possible to enforce a read-only access to the whole filesystem?

Sorry I didn't saw this interface before the PR,  I'll read more about getspnam_r to see if a more fine tuned ruleset is feasible as it will be more restrictive than a read-only access to the whole filesystem.

modules/pam_unix/unix_chkpwd.c
nbouchinet-anssi: As the landlock support might be optional and thus explicitly asked by the user at configuration time this could lead to a hard dependency to it, which I think is a valid reason to return a `PAM_SYSTEM_ERR`, but I might be wrong.

l0kod: `PAM_SYSTEM_ERR` is legitimate but only when an unexpected Landlock error occurs, not when Landlock is not supported by the running kernel (best-effort security approach).

configure.ac
```diff
@@ -492,6 +492,21 @@ if test -n "$LIBSELINUX" ; then
     LIBS=$BACKUP_LIBS
 fi
 
+AC_ARG_ENABLE([sandbox],
+    AS_HELP_STRING([--enable-sandbox],[Enable landlock sandbox]),
```
l0kod, line 496: I think it should be `--disable-sandbox` instead (i.e. secure by default), otherwise nobody will use sandboxing.

l0kod: BTW, **L**andlock is a proper noun. ;)

t8m: No! It should not be enabled by default. At least not currently.

ldv-alt: Anyway,  `sandbox` is too generic, please use `landlock` instead.

l0kod: At least it needs to be part of the CI, with a kernel supporting Landlock (Linux >= 5.13, and potentially >= 5.19 to check `LANDLOCK_ACCESS_FS_REFER`).

libpam/include/security/pam_landlock.h
```diff
@@ -0,0 +1,148 @@
...
+#define ACCESS_FILE_RW ( \
+		LANDLOCK_ACCESS_FS_WRITE_FILE | \
+		LANDLOCK_ACCESS_FS_READ_FILE)
+
+#define ACCESS_FILE_RO (LANDLOCK_ACCESS_FS_READ_FILE)
+
+static int _pam_landlock_create_ruleset(int * const ruleset_fd) {
+	int abi;
+
+	__u64 access_fs_rw = ACCESS_FS_ROUGHLY_READ | ACCESS_FS_ROUGHLY_WRITE;
```
l0kod, line 81: The `access_fs_rw` variable is not useful.

libpam/include/security/pam_landlock.h
```diff
@@ -0,0 +1,148 @@
...
+	 * Checks for Landlock abi compatibility.
+	 */
+	abi = landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION);
+	if (abi < 0) {
+		D(("Failed to check Landlock compatibility,"
+		   "Landlock might be either disabled or not supported by the current kernel."));
+		return 1;
+	}
+	if (abi < 2) {
+		ruleset_attr.handled_access_fs &= ~ACCESS_ABI_2;
```
l0kod, line 97: As for the linked example (on my next branch), `ACCESS_ABI_2` is not useful, we can directly use `LANDLOCK_ACCESS_FS_REFER` instead.

libpam/include/security/pam_landlock.h
```diff
@@ -0,0 +1,148 @@
...
+	};
+
+	/*
+	 * Checks for Landlock abi compatibility.
+	 */
+	abi = landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION);
+	if (abi < 0) {
+		D(("Failed to check Landlock compatibility,"
+		   "Landlock might be either disabled or not supported by the current kernel."));
+		return 1;
```
l0kod, line 94: This should not be an error but a warning, and the sandboxing code following the `_pam_landlock_create_ruleset()` call should be ignored.

nbouchinet-anssi: Yup, I planned to change it but didn't have time yesterday, thanks.

libpam/include/security/pam_landlock.h
nbouchinet-anssi: Removed

libpam/include/security/pam_landlock.h
nbouchinet-anssi: Done

libpam/include/security/pam_landlock.h
nbouchinet-anssi: Done

configure.ac
nbouchinet-anssi: Done, the flat has been renamed and is disabled by default.

modules/pam_unix/unix_chkpwd.c
nbouchinet-anssi: Done, a LOG_NOTICE message is sent to the user and the program continues if Landlock is not supported by the running kernel.

libpam/include/security/pam_sandbox.h
nbouchinet-anssi: Done

libpam/include/security/pam_sandbox.h
nbouchinet-anssi: Those macros are detected at configuration time thanks to the configure.ac file, hence this code can't be reached if the libc headers are too old.

nbouchinet-anssi: Oh, I missuses the AC_CHECK_DECLS, will have to tweak it to exclude Landlock if one of the Landlock symbols is missing.

nbouchinet-anssi:
Hi, I will not be available for the next 3 weeks.

nbouchinet-anssi:
Hi, sorry it took far more than 3 weeks for me to reply.

I've change the Landlock rule to enforce a read-only access to the whole filesystem [suggested](https://github.com/linux-pam/linux-pam/pull/486#issuecomment-1224554904) by @ldv-alt .

The `_pam_landlock_restrict_to_file_read()` function has thus been removed and could be sent in a later patch permitting a more fine grained Landlock configuration, unused macros has also been removed.

l0kod (commented):
Looks good to me but I didn't test it.

libpam/include/security/pam_landlock.h
```diff
@@ -0,0 +1,106 @@
...
+
+static int _pam_landlock_create_ruleset(int * const ruleset_fd, const __u64 ruleset) {
+	int abi;
+
+	struct landlock_ruleset_attr ruleset_attr = {
+		.handled_access_fs = ruleset,
+	};
+
+	/*
+	 * Checks for Landlock abi compatibility.
```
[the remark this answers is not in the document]
nbouchinet-anssi: Done

modules/pam_unix/unix_chkpwd.c
```diff
@@ -103,6 +107,30 @@ int main(int argc, char *argv[])
...
+	 */
+	int ruleset_fd, err;
+
+	err = _pam_landlock_create_ruleset(&ruleset_fd, ACCESS_FS_ROUGHLY_WRITE);
+	if (err) {
+		if (err == LANDLOCK_FATAL_ERR) {
+			helper_log_err(LOG_ERR ,"Failed to create base ruleset");
+			return PAM_SYSTEM_ERR;
+		} else {
+			helper_log_err(LOG_NOTICE , "Failed to check Landlock compatibility,"
```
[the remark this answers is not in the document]
nbouchinet-anssi: Done

modules/pam_unix/unix_chkpwd.c
```diff
@@ -103,6 +107,30 @@ int main(int argc, char *argv[])
...
+						    "Landlock is either disabled or not supported by the kernel");
```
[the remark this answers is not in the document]
nbouchinet-anssi: Done

chripo:
What is the current status of this pull request?

ldv-alt:
> What is the current status of this pull request?

If anyone is still interested in this PR, please update it and check whether it sill works as expected.

Issue #390, opened 2021-09-24
Faillock default directory usage is not consistent


Konrad-Ziarko:
Hi
My issue is that faillock does not read `dir` value from `/etc/security/faillock.conf` when invoking `faillock` command without arguments.
I would expect that value `dir` in configuration is the default value.

Right now faillock is working a little unexpectedly:  
- On failed login, faillock uses `dir` from `/etc/security/faillock.conf` to store data.  
- When I invoke command `faillock` without arguments, it uses `dir=/var/run/faillock`.  
I have to specify `--dir` which is logical but I would expect `dir` from config file to be set as default for operations.

Is there any logical reason why it works this way?
> pam-1.3.1-14.el8.x86_64

Konrad-Ziarko:
As I see in code `opts->dir = FAILLOCK_DEFAULT_TALLYDIR;` is set whenever argument `--dir` was not provided.
So it seems like this change would require inserting `read_config_file()` into `args_parse` function.

t8m:
Yes, would you like to submit a PR?

ldv-alt:
> As I see in code `opts->dir = FAILLOCK_DEFAULT_TALLYDIR;` is set whenever argument `--dir` was not provided. So it seems like this change would require inserting `read_config_file()` into `args_parse` function.

`args_parse` invokes ``read_config_file` since commit fc867a9e2, this issue seems to be resolved.

Konrad-Ziarko:
it seems so; waiting for new pam release

PR #491, opened 2022-09-19
pam_faillock: set config dir to the default value

dengbo11:
* modules/pam_faillock/main.c (args_parse): opts->dir need to be set the default value(FAILLOCK_DEFAULT_TALLYDIR) when invoking faillock command without arguments.

Resolves: https://github.com/linux-pam/linux-pam/issues/390

### DIFF
--- a/modules/pam_faillock/main.c
+++ b/modules/pam_faillock/main.c
@@ -106,6 +106,8 @@ args_parse(int argc, char **argv, struct options *opts)
 		}
 	}
 
+	opts->dir = get_tally_dir(opts);
+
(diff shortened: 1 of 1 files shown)


modules/pam_faillock/main.c
```diff
@@ -106,6 +106,8 @@ args_parse(int argc, char **argv, struct options *opts)
 		}
 	}
 
+	opts->dir = get_tally_dir(opts);
+
```
t8m, line 110: This is not correct. Also I am not sure the #390 is still relevant issue. There were changes in pam_faillock in regards to reading the file recently.

ldv-alt:
I also don't see why this PR is needed after commit fc867a9e2.

Issue #466, opened 2022-05-29
Undefined variable 'exec_prefix' in Pkgconfig file pam.pc and others


lechner:
Thanks for providing Pkgconfig files per #369. On Guix, their current [build recipe produces](https://github.com/guix-mirror/guix/blob/master/gnu/packages/linux.scm#L1601-L1613) a defective `pam.pc` in that the placeholder `${exec_prefix}` is used but not defined:
```
libdir=${exec_prefix}/lib
includedir=/gnu/store/bxzpnbpss9raf0gcbyv67x640s31il1p-linux-pam-1.5.2/include/security

Name: PAM
Description: The primary Linux-PAM library. It is used by PAM modules and PAM-aware applications.
URL: http://www.linux-pam.org/
Version: 1.5.2
Cflags: -I${includedir}
Libs: -L${libdir} -lpam
```
In Guix, the issue can probably be mitigated via additional command-line options to `./configure` but somehow there still seemed to be a bug in the interaction between [AC_PREFIX_DEFAULT here](https://github.com/linux-pam/linux-pam/blob/master/configure.ac#L17), the [m4/shell test here](https://github.com/linux-pam/linux-pam/blob/master/configure.ac#L28-L36) and the substitution of shell (and perhaps Makefile) [variable here](https://github.com/linux-pam/linux-pam/blob/master/libpam/Makefile.am#L45-L46).

The same issue occurs in the other two Pkgconfig files, `pamc.pc` and `pam_misc.pc`.

Thank you for taking a look!

ldv-alt:
The `exec_prefix` variable is maintained by [autoconf](https://www.gnu.org/software/autoconf/manual/autoconf-2.71/html_node/Installation-Directory-Variables.html).  The documentation says:
```By default it’s the same as prefix.```

lechner:
Is the issue perhaps that Autoconf is not in charge of Pkgconfig files? The variable is apparently no longer available when `pkg-config` runs at a later time—long after PAM was configure and installed—unless the variable is also declared in Pkgconfig via `exec_prefix=[LITERAL EXEC PREFIX]`.

ldv-alt:
`libpam/pam.pc`, `libpam_misc/pam_misc.pc`, and `libpamc/pamc.pc` are listed in `AC_CONFIG_FILES` and therefore autoconf _is_ responsible for `exec_prefix` substitution.

[...]

PR #493, opened 2022-09-28
libpam*: For uncommon prefixes, provide substitution variables in pkgconfig files.

lechner:
Fix undefined references to ${exec_prefix} in pkgconfig files on Guix. The subsequent declarations of ${libdir} and ${includedir} in the same files require this commit when ${prefix} is set to something other than /usr.

When the pkgconfig files were initially provided, the two lines added here were dropped for what seemed like a good reason. [1] In the common case of a /usr prefix, 'configure.ac' sets ${libdir} and possibly ${includedir} explicitly [2] so the additional lines were then not needed.

Guix and probably Nix too, however, depart from the Filesystem Hierarchy Standard and require the missing lines. Without those lines, the pkgconfig files are defective on Guix. [3]

Since working systems are not affected, the lines are added for all. The fix was confirmed for Guix. One of the files looked like this:

prefix=/gnu/store/3mcmjilqrivrpb3hvps32lnbnyrxrzr8-linux-pam-1.5.2-1.dc2f566 exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=/gnu/store/3mcmjilqrivrpb3hvps32lnbnyrxrzr8-linux-pam-1.5.2-1.dc2f566/include/security

Name: PAM
Description: The primary Linux-PAM library. It is used by PAM modules and PAM-aware applications. URL: http://www.linux-pam.org/
Version: 1.5.2
Cflags: -I${includedir}
Libs: -L${libdir} -lpam

* libpam/pam.pc.in, libpamc/pamc.pc.in, libpam_misc/pam_misc.pc.in: Add @prefix@ and @exec_prefix@.

Resolves: https://github.com/linux-pam/linux-pam/issues/466

[1] https://github.com/linux-pam/linux-pam/pull/369#discussion_r650557756
[2] https://github.com/linux-pam/linux-pam/blob/40c271164dbcebfc5304d0537a42fb42e6b6803c/configure.ac#L28-L36
[3] https://github.com/linux-pam/linux-pam/issues/466

### DIFF
--- a/libpam/pam.pc.in
+++ b/libpam/pam.pc.in
@@ -1,3 +1,5 @@
+prefix=@prefix@
+exec_prefix=@exec_prefix@
 libdir=@libdir@
 includedir=@includedir@
 
--- a/libpam_misc/pam_misc.pc.in
+++ b/libpam_misc/pam_misc.pc.in
@@ -1,3 +1,5 @@
+prefix=@prefix@
+exec_prefix=@exec_prefix@
 libdir=@libdir@
 includedir=@includedir@
 
--- a/libpamc/pamc.pc.in
+++ b/libpamc/pamc.pc.in
@@ -1,3 +1,5 @@
+prefix=@prefix@
+exec_prefix=@exec_prefix@
 libdir=@libdir@
 includedir=@includedir@
 


ldv-alt (approved):

ldv-alt:
Merged, thank you for your contribution.

Issue #498, opened 2022-10-21
Support for vendor locations (pam_shells and /etc/shells)


dvzrv:
Hi! I'm currently looking into the integration of /etc/shells in the packaging ecosystem on Arch Linux.
Our current way of dealing with /etc/shells is as follows:

* some shells (e.g. sh and bash) are in the file by default (probably for historical reasons)
* other shells sed themselves in and out of the file in package addition/ removal hooks

This means, that /etc/shells provides a form of state, which needs to be maintained by the package management system or the user, or by other means.
When looking at the concept of immutable systems with vendor locations such as /usr, it becomes non-trivial to deal with files such as /etc/shells which are really an accumulation of several inputs though.
Would it be possible to support a vendor location (e.g. /usr/etc/shells.d/ or /usr/lib/shells.d/), which is trumped by /etc/shells and may contain drop-in files?
That way providers of shell packages on distributions can just add their respective files containing the shell path(s) and pam_shells would just consume them from that vendor location.

The upside to this is, that a system can be run with a removed /etc and be bootstrapped from the vendor location and that this would simplify the packaging of shells on downstream distributions. I'm aware, that other projects would need to buy in on this as well (e.g. shadow-utils, [util-linux](https://github.com/util-linux/util-linux/issues/1857))

ldv-alt:
Some people also suggested in the past that packages could maintain their `shell.d` files and the package management system could maintain `/etc/shells` based on the data in these `shell.d` files, see e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1704972#c3, this is likely to be a simpler approach because it doesn't require coordination between all kinds of projects that might be reading `/etc/shells`.

dvzrv:
> Some people also suggested in the past that packages could maintain their shell.d files and the package management system could maintain /etc/shells based on the data in these shell.d files

The problem with this (current) approach is, that `/etc/shells` is the result of a function of an arbitrary process and all downstreams have to maintain this type of boilerplate. If `/etc/shells` is accumulated from a list of `shell.d` files, this means a process needs to write the file after triggering of a package hook and potentially also on every boot (when thinking of immutable systems). While this can be done (e.g. in the context of systemd using [systemd-generators](https://man.archlinux.org/man/core/systemd/systemd.generator.7.en)), it means even more overhead and a write operation that could be circumvented.
[...]

PR #509, opened 2022-11-18
pam_shells: Use the vendor directory as fallback for a distribution provided default config if there is no one in /etc.

schubi2:
Use the vendor directory as fallback for a distribution provided default config if there is no one in /etc.

If pam will be compiled with the option --enable-vendordir=<vendor_dir> and NOT defined 
--disable-econf, the files which define valid login shells will be parsed in following order:

- <vendor_dir>/shells
- <vendor_dir>/shells.d/*
- /etc/shells.d/shells

But all files in <vendor_dir> will be ingnored if the user has defined his
own file /etc/shells.

This PR should solve: https://github.com/linux-pam/linux-pam/issues/498

- Makefile.am: Add libeconf setting.
- pam_shells.c: Take care about the fallback configuration in the vendor directory.
- pam_shells.8.xml: Added description for the vendor directory.
- configure.ac: Added ECONF settings for building man pages.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -520,8 +520,7 @@ AC_ARG_ENABLE([econf],
   AS_HELP_STRING([--disable-econf], [do not use libeconf]),
   [WITH_ECONF=$enableval], WITH_ECONF=yes)
 if test "$WITH_ECONF" = "yes" ; then
-  PKG_CHECK_MODULES([ECONF], [libeconf], [],
-  [AC_CHECK_LIB([econf],[econf_readDirs],[ECONF_LIBS="-leconf"],[ECONF_LIBS=""])])
+  AC_CHECK_LIB([econf],[econf_readDirsWithCallback],[ECONF_LIBS="-leconf"],[ECONF_LIBS=""])
   if test -n "$ECONF_LIBS" ; then
     ECONF_CFLAGS="-DUSE_ECONF=1 $ECONF_CFLAGS"
   fi
@@ -535,7 +534,11 @@ if test -n "$enable_vendordir"; then
 		     [Directory for distribution provided configuration files])
   AC_DEFINE_UNQUOTED([VENDOR_SCONFIGDIR], ["$enable_vendordir/security"],
 		     [Directory for PAM modules distribution provided configuration files])
-  STRINGPARAM_VENDORDIR="--stringparam vendordir '$enable_vendordir' --stringparam profile.condition 'with_vendordir'"
+  if test "$WITH_ECONF" = "yes" ; then
+    STRINGPARAM_VENDORDIR="--stringparam vendordir '$enable_vendordir' --stringparam profile.condition 'with_vendordir;with_vendordir_and_with_econf'"
+  else
+    STRINGPARAM_VENDORDIR="--stringparam vendordir '$enable_vendordir' --stringparam profile.condition 'with_vendordir;with_vendordir_and_without_econf"
+  fi
 else
   STRINGPARAM_VENDORDIR="--stringparam profile.condition 'without_vendordir'"
 fi
--- a/modules/pam_shells/pam_shells.c
+++ b/modules/pam_shells/pam_shells.c
@@ -13,27 +13,47 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <sys/stat.h>
 #include <syslog.h>
 #include <unistd.h>
+#if defined (USE_ECONF)	&& defined (VENDORDIR)
+#include <libeconf.h>
+#endif
 
 #include <security/pam_modules.h>
 #include <security/pam_modutil.h>
 #include <security/pam_ext.h>
 
 #define SHELL_FILE "/etc/shells"
-
+#define SHELLS "shells"
+#define ETCDIR "/etc"
 #define DEFAULT_SHELL "/bin/sh"
+
+static bool check_file(const char *filename, const void *pamh)
+{
+    struct stat sb;
+
+    if (stat(filename, &sb)) {
+	pam_syslog(pamh, LOG_ERR, "Cannot stat %s: %m", filename);
+	return false;		/* must have /etc/shells */
+    }
+
+    if ((sb.st_mode & S_IWOTH) || !S_ISREG(sb.st_mode)) {
+	pam_syslog(pamh, LOG_ERR,
+		   "%s is either world writable or not a normal file",
+		   filename);
+	return false;
+    }
+    return true;
+}
 
 static int perform_check(pam_handle_t *pamh)
 {
     int retval = PAM_AUTH_ERR;
     const char *userName;
     const char *userShell;
-    char shellFileLine[256];
-    struct stat sb;
     struct passwd * pw;
-    FILE * shellFile;
 
     retval = pam_get_user(pamh, &userName, NULL);
     if (retval != PAM_SUCCESS) {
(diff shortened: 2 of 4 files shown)


modules/pam_shells/pam_shells.c
```diff
@@ -48,6 +50,85 @@ static int perform_check(pam_handle_t *pamh)
...
+	    free(file_path);
+	    free(begin_key_files);
+	    return PAM_AUTH_ERR;
+	}
+	free(file_path);
+        key_files++;
+    }
+    free(begin_key_files);
+
+    error = econf_readDirs(&key_file,
```
ldv-alt, line 98: Here `econf_readDirs` is called, which in turn would invoke `econf_readDirsHistory` again, but this time without all the checks implemented above, which is obviously a TOCTOU race condition.
Regretfully, econf doesn't provide a race-free API for this task yet.
What do you think about extending econf API, e,g, with a variant of `econf_readDirs` with a callback function?

ldv-alt (changes requested):

ldv-alt:
Stefan, please note that this time again the commit message doesn't contain the information listed in the PR.

schubi2:
> Stefan, please note that this time again the commit message doesn't contain the information listed in the PR.

Yes, I have missed it again.....

modules/pam_shells/pam_shells.c
```diff
@@ -48,6 +50,85 @@ static int perform_check(pam_handle_t *pamh)
...
 	userShell = DEFAULT_SHELL;
 
+#if defined (USE_ECONF)	&& defined (VENDORDIR)
+    size_t size = 0;
+    econf_file **key_files, **begin_key_files;
+    econf_err error;
+    char **keys;
+    econf_file *key_file;
+
+    error = econf_readDirsHistory(&key_files,
```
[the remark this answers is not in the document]
schubi2: I have added a check in the configure.ac

modules/pam_shells/pam_shells.c
schubi2: That's an good idea! I have implemented it.

modules/pam_shells/pam_shells.c
```diff
@@ -13,27 +13,51 @@
...
 
 #define SHELL_FILE "/etc/shells"
-
+#define SHELLS "shells"
+#define ETCDIR "/etc"
 #define DEFAULT_SHELL "/bin/sh"
 
+static pam_handle_t *log_pamh = NULL;
+
+static bool check_file(const char *filename)
```
ldv-alt, line 35: Stefan, would it be possible to let the callback function have an extra opaque argument that could be specified to the function that calls this callback?
If yes, this would help to avoid adding a static variable to the module - something we strive not  to do nowadays.

ldv-alt (changes requested):

modules/pam_shells/pam_shells.c
schubi2: Yes, that's a good idea. I have implemented it.

modules/pam_shells/pam_shells.c
```diff
@@ -13,27 +13,48 @@
...
 #define SHELL_FILE "/etc/shells"
-
+#define SHELLS "shells"
+#define ETCDIR "/etc"
 #define DEFAULT_SHELL "/bin/sh"
 
+static bool check_file(const char *filename, const void *data)
+{
+    struct stat sb;
+    const pam_handle_t *log_pamh = (const pam_handle_t *)data;
```
ldv-alt, line 36: I don't think you need an explicit cast from `void *` to `pam_handle_t *`.

modules/pam_shells/pam_shells.c
```diff
@@ -48,18 +69,50 @@ static int perform_check(pam_handle_t *pamh)
...
+    econf_file *key_file;
+
+    error = econf_readDirsWithCallback(&key_file,
+				       VENDORDIR,
+				       ETCDIR,
+				       SHELLS,
+				       NULL,
+				       "", /* key only */
+				       "#", /* comment */
+				       check_file, (void *)pamh);
```
ldv-alt, line 85: Do you need an explicit cast from `pam_handle_t *` to `void *`?

modules/pam_shells/pam_shells.c
```diff
@@ -48,18 +69,50 @@ static int perform_check(pam_handle_t *pamh)
...
+	retval = strcmp(keys[i], userShell);
+        if (!retval)
+	   break;
+    }
+    econf_free (key_file);
+#else
+    char shellFileLine[256];
+    FILE * shellFile;
+
+    if (!check_file(SHELL_FILE, (void *)pamh))
```
ldv-alt, line 113: Do you need an explicit cast from `pam_handle_t *` to `void *`?

ldv-alt (changes requested):
Looks good, but I don't see why do you add these explicit casts.

configure.ac
```diff
@@ -523,6 +523,7 @@ if test "$WITH_ECONF" = "yes" ; then
   PKG_CHECK_MODULES([ECONF], [libeconf], [],
   [AC_CHECK_LIB([econf],[econf_readDirs],[ECONF_LIBS="-leconf"],[ECONF_LIBS=""])])
   if test -n "$ECONF_LIBS" ; then
+    AC_CHECK_LIB([econf],[econf_readDirsWithCallback], [], [AC_MSG_ERROR([Please update libeconf])])
     ECONF_CFLAGS="-DUSE_ECONF=1 $ECONF_CFLAGS"
   fi
```
ldv-alt, line 528: This configure snippet requires that `libeconf` either doesn't exist or is fresh enough, which is somewhat unusual.
Wouldn't it be better to just check for `econf_readDirsWithCallback` instead?

modules/pam_shells/pam_shells.c
schubi2: Yes, you are right. I am wondering why the compiler does not complain it.

modules/pam_shells/pam_shells.c
schubi2: Yes, you are right. I am wondering why the compiler does not complain it.

modules/pam_shells/pam_shells.c
schubi2: Yes, you are right. I am wondering why the compiler does not complain it.

configure.ac
schubi2: Yes, it is much more simple.

ldv-alt:
Stefan, the commit message contains just a single line of text, could you add some descriptive text from this PR to the commit message, please?

schubi2:
> Stefan, the commit message contains just a single line of text, could you add some descriptive text from this PR to the commit message, please?

I have added a little description how it is working now with vendor dirs.

ldv-alt (approved):

ReillyBrogan:
So I've been working on integrating support for this into our own distribution (Solus). One hangup I'm finding is that accountsservice uses libc getusershell which without patches only supports `/etc/shells`. Since we want to be able to ship `shells.d` files alongside the shell packages they are part of this seems to be a blocker to us since accountsservice-using DEs (GNOME/Budgie/MATE/etc) won't detect that such shells are valid shells in that case. 

I'm curious what the best path forward there is, it doesn't look like SUSE is carrying any patches to glibc or accountsservice to enable that so I assume the `shells.d` approach is currently not utilized on SUSE distributions? 

Would it be a good idea to add a getusershell-like function to libpam that other applications like accountsservice could be patched to use instead of getusershell? This would prevent duplication of the vendordir/libeconf functionality across all applications that use that function and such patches would be relatively easy to maintain (at least until they were upstreamed of course). Or would it be better to try to get the functionality into glibc natively?

thkukuk:
@ReillyBrogan you could try to use https://github.com/thkukuk/libusershell_econf

ReillyBrogan:
> Better would be a separate library providing a `getusershell()` function and linking all applications against it. But will be a lot of work to identify this applications and patch the link process.

Well, luckily Solus has ABI scanning built into our package building so I can say that the users of getusershell are (at least in our smaller repo, and not accounting for anything that builds with a static libc):
- [accountsservice](https://gitlab.freedesktop.org/accountsservice/accountsservice/-/blob/23.13.9/src/user-classify.c?ref_type=tags#L94)
- GNU CLISP (they use a Sourceforge-hosted Mercurial repository which doesn't support web search of the code so I'm not going to bother getting a code link right now)
- [kerberos](https://github.com/krb5/krb5/blob/krb5-1.21.2-final/src/clients/ksu/main.c#L974C1-L974C42)
- [sudo](https://github.com/sudo-project/sudo/blob/SUDO_1_9_15p2/plugins/sudoers/pwutil_impl.c#L466)
- texlive - Not actually sure which of their source repos corresponds to the tarball
- [util-linux](https://github.com/util-linux/util-linux/blob/v2.39.2/login-utils/chsh.c#L93) - Perhaps this was a spot where econf support was missed?

> @ReillyBrogan you could try to use https://github.com/thkukuk/libusershell_econf

I'll give that a try, looks easy enough to make some test patches with. I do think that perhaps creating separate libs for each piece of functionality might not be the best approach (I see you have a NSS module as well). Perhaps creating a libhermetic/libeconf-helper/libNameCanBeWorkshopped which acts as a "glue" library to wrap common functionality around libeconf might be a better approach? It would be easier from a distro integration standpoint to only have to package one library and then to be able to read the documentation/release notes to see all the various "things" that it supports.

thkukuk:
A NSS module is no "library", you cannot combine this. I was long enough glibc developer to know what you can combine and what not. Best would be if application developers would start to support hermetic /usr native.
But this all has nothing to do anymore with this PR.

PR #510, opened 2022-11-18
pam_shells: Added xtest test case

schubi2:
Test case for checking pam_authenticate in pam_shells

xtests/Makefile.am: Added additional test.
xtests/run-xtests.sh: Using own "shells" file
xtests/shells.conf: valid shell definitions which will be used for /etc/shells
xtests/tst-pam_shells.c: Calling pam_authenticate
xtests/tst-pam_shells.pamd: pam section description for pam_namespace
xtests/tst-pam_shells.sh: starting script


xtests/Makefile.am
```diff
@@ -43,7 +44,8 @@ XTESTS = tst-pam_dispatch1 tst-pam_dispatch2 tst-pam_dispatch3 \
 	tst-pam_access1 tst-pam_access2 tst-pam_access3 \
 	tst-pam_access4 tst-pam_limits1 tst-pam_succeed_if1 \
 	tst-pam_group1 tst-pam_authfail tst-pam_authsucceed \
-	tst-pam_pwhistory1 tst-pam_time1 tst-pam_motd
+	tst-pam_pwhistory1 tst-pam_time1 tst-pam_motd \
+        tst-pam_shells
```
ldv-alt, line 48: Note that tst-pam_shells is indented with spaces while the lines above it are undented with tabs.

xtests/tst-pam_shells.c
```diff
@@ -0,0 +1,126 @@
...
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/time.h>
+#include <sys/resource.h>
```
ldv-alt, line 51: Why would you need to include these headers?

xtests/tst-pam_shells.c
```diff
@@ -0,0 +1,126 @@
...
+  for (count = 0; count < num_msg; ++count)
+    {
+      reply[count].resp_retcode = 0;
+      reply[count].resp = strdup ("!!");
+    }
+
+  /* Set the pointers in the response structure and return. */
+  *response = reply;
+  return PAM_SUCCESS;
+}
```
ldv-alt, line 83: Why would you need such a non-trivial conversation function?

xtests/tst-pam_shells.pamd
```diff
@@ -0,0 +1,5 @@
+#%PAM-1.0
+auth     required       pam_shells.so
+account  required       pam_permit.so
+password required       pam_permit.so
+session  required       pam_permit.so
```
ldv-alt, line 5: Why would you need to use non-auth types?

ldv-alt (changes requested):

xtests/tst-pam_shells.c
```diff
@@ -0,0 +1,92 @@
...
+  if (argc > 1 && strcmp (argv[1], "-d") == 0)
+    debug = 1;
+
+  retval = pam_start("tst-pam_shells", user, &conv, &pamh);
+  if (retval != PAM_SUCCESS)
+    {
+      if (debug)
+	fprintf (stderr, "pam_shells: pam_start returned %d\n", retval);
+      return 1;
+    }
```
ldv-alt, line 73: Feel free to use test_assert.h framework like most of `modules/*/tst-*.c` tests do.

xtests/tst-pam_shells.pamd
```diff
@@ -0,0 +1,2 @@
+#%PAM-1.0
+auth     required       pam_shells.so
```
ldv-alt, line 2: git warns that there is no newline at the end of this file.

ldv-alt (changes requested):
While the current edition of the test checks that the shell listed in `/etc/shells` is indeed accepted, I suggest also to check that the shell not listed in `/etc/shells` is rejected.

For example, you can add an option to `tst-pam_shells` instructing it whether `pam_authenticate` is expected to return `PAM_SUCCESS` or `PAM_AUTH_ERR`.

xtests/tst-pam_shells.c
schubi2: Yes, I have removed it.

xtests/tst-pam_shells.pamd
schubi2: fixed

schubi2:
> While the current edition of the test checks that the shell listed in `/etc/shells` is indeed accepted, I suggest also to check that the shell not listed in `/etc/shells` is rejected.
> 
> For example, you can add an option to `tst-pam_shells` instructing it whether `pam_authenticate` is expected to return `PAM_SUCCESS` or `PAM_AUTH_ERR`.
I have added this use case

xtests/tst-pam_shells.c
```diff
@@ -0,0 +1,92 @@
...
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <security/pam_appl.h>
+
+static struct pam_conv conv = {
+    NULL,
+    NULL
+};
```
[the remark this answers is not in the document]
schubi2: Yes, I have removed it.

xtests/tst-pam_shells.c
```diff
@@ -0,0 +1,72 @@
...
+  shells.conf:
+  /bin/testbash
+
+*/
+
+#include "test_assert.h"
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
```
ldv-alt, line 46: The rule of thumb is that if <config.h> has to be included, it should be the first included file.
Fortunately, "test_assert.h" already includes <config.h>, so the order of inclusion doesn't cause any issues here.
I could suggest removing `#include <config.h>` here.

ldv-alt (changes requested):
Stefan, please note that this time again the commit message doesn't contain the information listed in the PR.

xtests/tst-pam_shells.c
schubi2: Yes, I have removed it and it is compiling :+1:

schubi2:
> Stefan, please note that this time again the commit message doesn't contain the information listed in the PR.

Thorsten has explained me, how it should look like. I hope I have done it right.

ldv-alt (approved):
Ideally, we need to find a way how to run xtests in CI, but it's not a part of this PR anyway.

PR #511, opened 2022-11-23
pam_listfile: fix pointer misuse leading to data corruption

fayak:
pam module listfile assume the group being tested will be written at the end of the argument list by carrying only a pointer to the value being examined in `myval` to `apply_val`

Therefore example
```
auth    required       pam_listfile.so \
        onerr=succeed apply=ftp item=user sense=deny file=/etc/ftpusers
```

modified from https://linux.die.net/man/8/pam_listfile is not working because `apply_val` will point to the latest value of `myval` which in this case will be `/etc/ftpusers` instead of `ftp`

This PR fixes this issue by copying the value of `myval` to `apply_val` instead of just taking a reference pointer

### DIFF
--- a/modules/pam_listfile/pam_listfile.c
+++ b/modules/pam_listfile/pam_listfile.c
@@ -53,17 +53,16 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
     const char *citemp;
     char *ifname=NULL;
     char aline[256];
-    char mybuf[256],myval[256];
+    char mybuf[256],myval[256],apply_val[256];
     struct stat fileinfo;
     FILE *inf;
-    const char *apply_val;
     int apply_type;
 
     /* Stuff for "extended" items */
     struct passwd *userinfo;
 
     apply_type=APPLY_TYPE_NULL;
-    apply_val="";
+    apply_val[0] = '\0';
 
     for(i=0; i < argc; i++) {
 	{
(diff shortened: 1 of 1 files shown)


modules/pam_listfile/pam_listfile.c
```diff
@@ -53,17 +53,16 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
...
     FILE *inf;
-    const char *apply_val;
     int apply_type;
 
     /* Stuff for "extended" items */
     struct passwd *userinfo;
 
     apply_type=APPLY_TYPE_NULL;
-    apply_val="";
+    memset(apply_val,'\0',sizeof(apply_val));
```
ldv-alt, line 65: Looks like zeroing out the whole `apply_val` buffer is superfluous, `apply_val[0] = '\0'` should be sufficient.

ldv-alt (changes requested):
I agree with your analysis of the bug.
I wouldn't call it a `bad pointer`, though: there is nothing wrong with the pointer itself,
it's rather data clobbering or or data corruption or something like that.

Could you copy your description from the PR to the commit message itself, please?

fayak:
I've rephrased the title of the commit/PR, and added the description to the commit. Is it ok for you ?

ldv-alt (approved):

PR #514, opened 2022-12-06
doc/man/Makefile.am: fix XMLS list

thkukuk:
The XMLS list of xml sources for the manual pages missed some xml files and instead contained some nroff sources.


ldv-alt (approved):
I wonder how did you find out what files were missing.

thkukuk:
> I wonder how did you find out what files were missing.

I tried to apply pull requests to the original tar archive, which failed since files where missing. So I sorted the list and compared them one by one with what we have in git...

PR #515, opened 2022-12-09
Add module for pam_limit_ssh

pschou:
Often limits are needed only to prevent excessive connections remotely.  Local limits may be needed, but they don't have to be as strict.   This module enables SSH limits to be applied at the PAM level.

### DIFF
--- a/libpamc/include/security/pam_client.h
+++ b/libpamc/include/security/pam_client.h
@@ -110,7 +110,7 @@ char **pamc_list_agents(pamc_handle_t pch);
 
 #define PAM_BP_RENEW(old_p, cntrl, data_length)                            \
 do {                                                                       \
-    if ((void*)(old_p) != NULL) {                                                 \
+    if ((old_p) != NULL) {                                                 \
 	if (*(old_p)) {                                                    \
 	    uint32_t __size;                                              \
             __size = PAM_BP_SIZE(*(old_p));                                \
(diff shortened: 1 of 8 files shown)


ldv-alt:
I suppose the most reliable way of implementing this feature is within the sshd itself.
Anyway, I very much doubt that reading `/proc/[0-9]*/cmdline` is a good idea.

pschou:
> I suppose the most reliable way of implementing this feature is within the sshd itself. Anyway, I very much doubt that reading `/proc/[0-9]*/cmdline` is a good idea.

I thought the same thing, but how would / should ssh do this when the process has been restarted?

It's been a challenge for us as the pam_limits has been problematic due to how the limits are imposed.

thkukuk (changes requested):
This PR contains many changes which are partly highly questionable and have nothing to do with pam_limit_ssh itself.
Please don't mix this but create separate PRs for this.

thkukuk:
I don't like this approach at all, it heavily depends on the used ssh implementation (and there are several), kernel configuration and security measurements.
If we would accept the module in this form, we will be flooded with bug reports.

I see two options:
* You maintain the module yourself as standalone package. There are many PAM modules outside Linux-PAM doing this already, there is no need to get them all merged here.
* use a more stable and compatible interface. If systemd is in use, querying logind could be an option: go through all open sessions and count the remote ones by this user.

PR #518, opened 2023-01-16
pam_selinux: treat getenforce failures as enforcing

cgzones:
security_getenforce(3) can return -1 on error; either because the selinuxfs is not mounted or reading from /sys/fs/selinux/enforce failed.

Since security_getenforce(3) is either called after an approving call to is_selinux_enabled(3) in create_context() or with populated module data in restore_context(), which requires a previous pass of create_context(), the selinuxfs should be mounted. Reading from /sys/fs/selinux/enforce should never fail (except being prohibited by the SElinux policy itself) since it is a public interface.

In the unlikely case of security_getenforce(3) nevertheless failing continue execution as if the result was enforcing (likewise to pam_sepermit and pam_rootok).


ldv-alt (approved):
I agree with the analysis.
Please rebase so that it could be merged.

Issue #477, opened 2022-07-15
pam_dynamic.c:79 syntax error


AndrewChe7:
I found syntax error with cppcheck in 79 and 85 line of file libpam/pam_dynamic.c. Unmatched '(':
```c
if( shl_findsym(&handle, symbol, (short) TYPE_PROCEDURE, &ret ){
		_symbol = malloc( strlen(symbol) + sizeof(SHLIB_SYM_PREFIX) + 1 );
...
if( shl_findsym(&handle, _symbol,
				(short) TYPE_PROCEDURE, &ret ){
```
Seems like this code is never used, cause I didn't found any PAM_DYDL or PAM_SHL definition. Maybe there is no need in these part of code or syntax error should be fixed.

PR #521, opened 2023-01-23
libpam: remove dead code in pam_dynamic.c

ldv-alt:
Apparently, the PAM_SHL variant cannot be compiled since the very first commit back in 2005 when it was introduced, and another variant uses PAM_DYLD which is virtually unknown to search engines.

* libpam/pam_dynamic.c [PAM_SHL || PAM_DYLD]: Remove.

Resolves: https://github.com/linux-pam/linux-pam/issues/477

### DIFF
--- a/libpam/pam_dynamic.c
+++ b/libpam/pam_dynamic.c
@@ -32,107 +32,25 @@
  */
 
 #include "pam_private.h"
-
-#ifdef PAM_SHL
-# include <dl.h>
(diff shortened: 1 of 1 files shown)


thkukuk:
DYLD seems to be used by MacOS, don't know if that is really from relevance for us. 
Else I'm fine with dropping that support.

thkukuk (approved):

PR #524, opened 2023-01-25
pam_pwhistory: use vendor specific pwhistory.conf as fallback

schubi2:
Use the vendor directory defined by --enable-vendordir=DIR configure option as fallback for the distribution provided default config file if there is no configuration in /etc.

* modules/pam_pwhistory/pam_pwhistory.8.xml: Describes pwhistory.conf
* modules/pam_pwhistory/pwhistory_config.c: (parse_config_file) [VENDOR_PWHISTORY_DEFAULT_CONF] Try to open
   VENDOR_PWHISTORY_DEFAULT_CONF if PWHISTORY_DEFAULT_CONF file does not exist.
* modules/pam_pwhistory/tst-pam_pwhistory-retval.c: testcase for pam_pwhistory.
* modules/pam_pwhistory/Makefile.am: define testcase

### DIFF
--- a/modules/pam_pwhistory/pwhistory_config.c
+++ b/modules/pam_pwhistory/pwhistory_config.c
@@ -39,6 +39,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <syslog.h>
+#include <sys/stat.h>
 
 #include <security/pam_modutil.h>
 
@@ -46,6 +47,10 @@
 #include "pwhistory_config.h"
 
 #define PWHISTORY_DEFAULT_CONF SCONFIGDIR "/pwhistory.conf"
+
+#ifdef VENDOR_SCONFIGDIR
+#define VENDOR_PWHISTORY_DEFAULT_CONF (VENDOR_SCONFIGDIR "/pwhistory.conf")
+#endif
 
 void
 parse_config_file(pam_handle_t *pamh, int argc, const char **argv,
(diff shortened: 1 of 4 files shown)


thkukuk:
Looks good to me.

modules/pam_pwhistory/pwhistory_config.c
```diff
@@ -67,6 +72,19 @@ parse_config_file(pam_handle_t *pamh, int argc, const char **argv,
...
 
+#ifdef VENDOR_PWHISTORY_DEFAULT_CONF
+    if (!strcmp(fname, PWHISTORY_DEFAULT_CONF)) {
+        /*
+         * Check whether PWHISTORY_DEFAULT_CONF file is available.
+         * If it does not exist, fall back to VENDOR_PWHISTORY_DEFAULT_CONF file.
+         */
+        struct stat buffer;
+        if (stat(fname, &buffer) != 0 && errno == ENOENT) {
+            fname = VENDOR_PWHISTORY_DEFAULT_CONF;
```
ldv-alt, line 83: I think we shouldn't fall back to the vendor config file if a `conf=` option was specified and its argument was equal to the default config file name.
That is, I suggest moving the fall back code inside the `fname == NULL` branch, this would also make the `strcmp` check unnecessary.

ldv-alt (changes requested):

ldv-alt (approved):
The test is really very basic, and it also doesn't test this vendordir change, so I moved it into a separate commit, please have a look.  Also, I amended the commit message a bit, also please have a look.
Besides that, looks fine.

PR #528, opened 2023-01-30
overhaul pam_overwrite

cgzones:
Avoid compiler optimizations to elide the memory erasure by using a secure method: either `memset_explicit()` [C23], `bzero_explicit()` [glibc 2.25] or a manual memory barrier.

Also introduce a `_pam_overwrite_array()` convenience wrapper.

### DIFF
--- a/libpam/include/pam_inline.h
+++ b/libpam/include/pam_inline.h
@@ -161,7 +161,7 @@ pam_read_passwords(int fd, int npass, char **passwords)
 				if (npass > 0) {
 					memcpy(passwords[i], pptr, rbytes);
 				}
-				memset(pptr, '\0', rbytes);
+				pam_overwrite_n(pptr, rbytes);
 			}
 		}
 		offset += rbytes;
@@ -169,7 +169,7 @@ pam_read_passwords(int fd, int npass, char **passwords)
 
 	/* clear up */
 	if (offset > 0 && npass > 0) {
-		memset(passwords[i], '\0', offset);
+		pam_overwrite_n(passwords[i], offset);
 	}
 
 	return i;
--- a/libpam/include/pam_inline.h
+++ b/libpam/include/pam_inline.h
@@ -9,6 +9,7 @@
 #define PAM_INLINE_H
 
 #include "pam_cc_compat.h"
+#include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
 #include <errno.h>
@@ -33,6 +34,12 @@
  * 0, otherwise.
  */
 #define PAM_MUST_BE_ARRAY(a_)		PAM_FAIL_BUILD_ON_ZERO(!PAM_IS_NOT_ARRAY(a_))
+/*
+ * Evaluates to
+ * - a syntax error if the argument is an array,
+ * 0, otherwise.
+ */
+#define PAM_MUST_NOT_BE_ARRAY(a_)	PAM_FAIL_BUILD_ON_ZERO(PAM_IS_NOT_ARRAY(a_))
 
 /* Evaluates to the number of elements in the specified array.  */
 #define PAM_ARRAY_SIZE(a_)		(sizeof(a_) / sizeof((a_)[0]) + PAM_MUST_BE_ARRAY(a_))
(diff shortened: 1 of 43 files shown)


t8m:
This is a nice work. Is there any concern about targets where none of the secure overwrite functions aren't available and asm intrinsics aren't either?

libpam/include/security/_pam_macros.h
```diff
@@ -7,6 +7,8 @@
  * Organized by Cristian Gafton <gafton@redhat.com>
  */
 
+#include "config.h"
```
ldv-alt, line 10: Exported headers (those that are listed in `include_HEADERS`) are not allowed to `#include "config.h"` because the latter is not exported.

libpam/include/security/_pam_macros.h
```diff
@@ -17,22 +19,39 @@
...
-            __xx__[__i__] = 0; \
-} while (0)
+#define _pam_overwrite(x)                     \
+do {                                          \
+    char *xx__ = x;                           \
+    if (xx__)                                 \
+        _pam_overwrite_n(xx__, strlen(xx__)); \
+} while(0)
+
+#define _pam_overwrite_array(x) _pam_overwrite_n(x, sizeof(x) * sizeof(*x) + PAM_MUST_BE_ARRAY(x))
```
ldv-alt, line 54: If `x` is an array, `sizeof(x)` would be the size of the array in bytes, wouldn't it?

modules/pam_selinux/pam_selinux.c
```diff
@@ -393,7 +393,6 @@ free_module_data(module_data_t *data)
   freecon(data->prev_exec_context);
   if (data->exec_context != data->default_user_context)
     freecon(data->exec_context);
-  memset(data, 0, sizeof(*data));
```
ldv-alt, line 396: This removal is not obvious.

ldv-alt (changes requested):
It's unfortunate that `_pam_macros.h` with all its macros is exported at all:
all these new #ifdef's aren't that useful without `config,h` which is not exported.

t8m:
> It's unfortunate that `_pam_macros.h` with all its macros is exported at all:
> all these new #ifdef's aren't that useful without `config,h` which is not exported.

Yeah, one option could be to just keep the existing _pam_overwrite() macros as is and add internal macros in pam_inline.h instead that would be used throughout the code instead of _pam_overwrite().

libpam/include/security/_pam_macros.h
cgzones: Somehow I got confused about `sizeof()` and added the nested one; dropped.

modules/pam_selinux/pam_selinux.c
cgzones: The struct `module_data_t` consists of only pointers (so no sensitive data) and the memory is freed on the next line, so there is no further access (e.g. potential double free...).

cgzones:
See https://github.com/linux-pam/linux-pam/pull/531 for a Draft of replacement macros defined in `"pam_inline.h"`.

libpam/include/pam_inline.h
```diff
@@ -66,6 +67,61 @@ pam_str_skip_icase_prefix_len(const char *str, const char *prefix, size_t prefix
 #define pam_str_skip_icase_prefix(str_, prefix_)	\
 	pam_str_skip_icase_prefix_len((str_), (prefix_), sizeof(prefix_) - 1 + PAM_MUST_BE_ARRAY(prefix_))
 
+
+/*
+ * Macros to securly erase memory
+ */
+
+#ifdef HAVE_MEMSET_EXPLICIT
+# define pam_override_n(x, n)            \
```
t8m, line 76: Please use pam_overwrite* not override.

modules/pam_unix/md5.c
```diff
@@ -149,7 +151,7 @@ void MD5Name(MD5Final)(unsigned char digest[16], struct MD5Context *ctx)
 	MD5Name(MD5Transform)(ctx->buf.i, ctx->in.i);
 	byteReverse(ctx->buf.i, 4);
 	memcpy(digest, ctx->buf.c, 16);
-	memset(ctx, 0, sizeof(*ctx));	/* In case it's sensitive */
+	pam_overwrite_n(ctx, sizeof(*ctx));	/* In case it's sensitive */
```
ldv-alt, line 154: `pam_overwrite_object(ctx);`

modules/pam_unix/passverify.c
```diff
@@ -147,7 +147,7 @@ PAMH_ARG_DECL(int verify_pwd_hash,
 			if (cdata != NULL) {
 				cdata->initialized = 0;
 				pp = x_strdup(crypt_r(p, hash, cdata));
-				memset(cdata, '\0', sizeof(*cdata));
+				pam_overwrite_n(cdata, sizeof(*cdata));
```
ldv-alt, line 150: `pam_overwrite_object(cdata);`

modules/pam_unix/passverify.c
```diff
@@ -166,8 +166,10 @@ PAMH_ARG_DECL(int verify_pwd_hash,
 		}
 	}
 
-	if (pp)
+	if (pp) {
+		pam_overwrite(pp);
 		_pam_delete(pp);
```
ldv-alt, line 171: I suppose `_pam_delete` calls `pam_overwrite`.

modules/pam_unix/passverify.c
```diff
@@ -494,18 +496,21 @@ PAMH_ARG_DECL(char * create_password_hash,
...
+		pam_overwrite_n(cdata, sizeof(*cdata));
 		free(cdata);
 #endif
 		return NULL;
 	}
-	sp = x_strdup(sp);
+	ret = strdup(sp);
+	pam_overwrite(sp);
 #ifdef HAVE_CRYPT_R
+	pam_overwrite_n(cdata, sizeof(*cdata));
```
ldv-alt, line 510: `pam_overwrite_object(cdata);`

ldv-alt (changes requested):

libpam/include/pam_inline.h
```diff
@@ -66,6 +73,59 @@ pam_str_skip_icase_prefix_len(const char *str, const char *prefix, size_t prefix
...
+#define pam_overwrite_object(x) pam_overwrite_n(x, sizeof(*(x)) + PAM_MUST_NOT_BE_ARRAY(x))
+
+static inline void
+pam_drop_response(struct pam_response *reply, int replies)
+{
+	int reply_i;
+
+	for (reply_i = 0; reply_i < replies; ++reply_i) {
+		if (reply[reply_i].resp) {
+			pam_overwrite_string(reply[reply_i].resp);
```
ldv-alt, line 121: This function uses `pam_overwrite` instead of `pam_overwrite_string` in commit 32581eb7c006fa8dd646f1283ad001178fd7d8d0 which makes it uncompilable, please fix.

modules/pam_unix/passverify.c
```diff
@@ -166,8 +166,9 @@ PAMH_ARG_DECL(int verify_pwd_hash,
 		}
 	}
 
-	if (pp)
+	if (pp) {
 		_pam_delete(pp);
+	}
```
ldv-alt, line 171: This hunk doesn't seem to be relevant.

ldv-alt (approved):

ldv-alt:
I think this is ready to be merged, but the change is so vast that I could have easily missed something, so I'd rather wait a bit longer for some comments.

libpam/pam_get_authtok.c
```diff
@@ -174,6 +175,14 @@ pam_get_authtok_internal (pam_handle_t *pamh, int item,
...
     {
       /* We want to abort */
+      if (resp[0]) {
+        pam_overwrite_string (resp[0]);
+        _pam_drop (resp[0]);
+      }
+      if (resp[1]) {
+        pam_overwrite_string (resp[1]);
+        _pam_drop (resp[1]);
+      }
```
ldv-alt, line 185: Since both `pam_overwrite_string` and `_pam_drop` check their arguments, let's not duplicate the check here.

PR #531, opened 2023-02-01
overhaul pam_overwrite alternative

cgzones:


libpam/include/pam_inline.h
```diff
@@ -66,6 +66,60 @@ pam_str_skip_icase_prefix_len(const char *str, const char *prefix, size_t prefix
 #define pam_str_skip_icase_prefix(str_, prefix_)	\
 	pam_str_skip_icase_prefix_len((str_), (prefix_), sizeof(prefix_) - 1 + PAM_MUST_BE_ARRAY(prefix_))
 
+
+/*
+ * Macros to securly erase memory
+ */
+
+#ifdef HAVE_MEMSET_EXPLICIT
+# define _pam_override_n(x, n)           \
```
t8m, line 75: Rather than override I'd use _pam_cleanup() or even pam_overwrite() without the underscore - given this is internal macro we do not need to particularly care about conflicts.

t8m:
IMO this is the way to go.

libpam/include/pam_inline.h
```diff
@@ -66,6 +66,60 @@ pam_str_skip_icase_prefix_len(const char *str, const char *prefix, size_t prefix
...
+#define _pam_override(x)                             \
+do {                                                 \
+	char *xx__ = x;                              \
+	if (xx__)                                    \
+		_pam_override_n(xx__, strlen(xx__)); \
+} while(0)
+
+#define _pam_override_array(x) _pam_override_n(x, sizeof(x) + PAM_MUST_BE_ARRAY(x))
+
+#define _pam_drop_response(/* struct pam_response * */ reply, /* int */ replies) \
```
ldv-alt, line 108: Maybe we could have an actual static inline function instead of a macro.
Also, I don't think we really need underscore prefixes for new internal functions and macros.

PR #538, opened 2023-02-14
libpam: use getlogin() from libc and not utmp

thkukuk:
utmp uses 32bit time_t for compatibility with 32bit userland on some
 64bit systems and is thus not Y2038 safe. Use getlogin() from libc
 which avoids using utmp.

 * libpam/pam_modutil_getlogin.c: Use getlogin() instead of parsing utmp

### DIFF
--- a/libpam/pam_modutil_getlogin.c
+++ b/libpam/pam_modutil_getlogin.c
@@ -10,7 +10,6 @@
 
 #include <stdlib.h>
 #include <unistd.h>
-#include <utmp.h>
 
 #define _PAMMODUTIL_GETLOGIN "_pammodutil_getlogin"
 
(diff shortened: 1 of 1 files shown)


ldv-alt:
Given that nowadays `getlogin()` uses /proc/self/loginuid and therefore more safe than both the old `getlogin()` implementation and our utmp-based `pam_modutil_getlogin`, I'm in favor of this change.  I suggest adding to the commit message an comment explaining that modern `getlogin()` is now safe to use.

libpam/pam_modutil_getlogin.c
```diff
@@ -19,62 +18,33 @@ pam_modutil_getlogin(pam_handle_t *pamh)
...
+    curr_user_len = strlen(logname)+1;
+    curr_user = calloc(curr_user_len, 1);
     if (curr_user == NULL) {
-	goto clean_up_and_go_home;
+      return NULL;
     }
 
-    strncpy(curr_user, ut->ut_user, sizeof(ut->ut_user));
-    /* calloc already zeroed the memory */
+    strncpy(curr_user, logname, curr_user_len);
```
ldv-alt, line 40: The compiler seems to be too picky about this, let's use memcpy instead.

thkukuk: Not only too picky, but also wrong...

ldv-alt (approved):

PR #539, opened 2023-02-14
pam_lastlog: deprecate it and disable by default

thkukuk:
pam_lastlog uses utmp, wtmp, btmp and lastlog. None of them is Y2038 safe, even on 64bit architectures. Most 64bit architectures uses 32bit time_t for compat reasons with 32bit userland.
Additional, all relevant tools for which pam_lastlog would make sense have already own support for all four files, so this module will most likely only create duplicate entries.

* configure.ac: don't build pam_lastlog by default.

### DIFF
--- a/ci/run-build-and-tests.sh
+++ b/ci/run-build-and-tests.sh
@@ -5,7 +5,7 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-DISTCHECK_CONFIGURE_FLAGS='--disable-dependency-tracking --enable-Werror'
+DISTCHECK_CONFIGURE_FLAGS='--disable-dependency-tracking --enable-Werror --enable-lastlog'
 export DISTCHECK_CONFIGURE_FLAGS
 
 case "${TARGET-}" in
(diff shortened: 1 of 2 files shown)


ikerexxe:
Is there any replacement for pam_lastlog?

thkukuk:
> Is there any replacement for pam_lastlog?

As written: all tools for which pam_lastlog makes sense have the functionality already built-in. For what do you need a replacement?
Else your glibc developers wrote already that the corresponding support in glibc will be deprecated and not fixed, so 2038 at latest is the end anyways.

ldv-alt (dismissed):
It's fine to deprecate the module, but disabling it by default without any prior warning doesn't look very nice, so I suggest delaying the disablement to the next release.

Also, the are several grammar typos in the commit message:

- Most 64bit architectures uses -> Most 64bit architectures use
- Additional -> Additionally
- have already own -> already have their own

Could you fix them, please?

thkukuk:
> It's fine to deprecate the module, but disabling it by default without any prior warning doesn't look very nice, so I suggest delaying the disablement to the next release.

I disagree here, nobody will read the deprecation note if the module still builds. And you did the same with pam_cracklib, pam_tally and pam_tally2 (see commit f49166c7d8f3ae2c9d337154f7e5dc50d41ab6bf).
We don't remove the module yet, but people will only notice that something may change in the future if they need to get active and enable it again.

The wording is fixed.

ldv-alt (dismissed):
OK, there is a precedent indeed, although there is no hurry in removing the module completely.
Let's do it, but please add `--enable-cracklib` to `DISTCHECK_CONFIGURE_FLAGS` in ci/run-build-and-tests.sh script, to let the module be tested in CI as before.

ldv-alt (approved):

PR #541, opened 2023-02-22
Y2038: use logind instead of utmp

thkukuk:
The struct utmp from glibc uses on many 64bit architectures a 32bit time_t for compatibility with a 32bit userland, which means utmp will not survive the year 2038 (32bit time_t overflow). Use the data from logind instead of utmp.

* configure.ac: Add option --enable-logind
* modules/pam_issue/Makefile.am: Add CFLAGS/LIBS for logind support
* modules/pam_issue/pam_issue.c: Use sd_get_sessions instead of utmp
* modules/pam_timestamp/Makefile.am: Add CFLAGS/LIBS for logind support
* modules/pam_timestamp/pam_timestamp.c: query logind for login time

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -526,6 +526,17 @@ if test -n "$LIBSELINUX" ; then
     LIBS=$BACKUP_LIBS
 fi
 
+LOGIND_CFLAGS=
+SYSTEMD_LIBS=
+AC_ARG_ENABLE([logind],
+  AS_HELP_STRING([--disable-logind], [Disable logind support]),
+  [WITH_LOGIND=$enableval], [WITH_LOGIND=yes])
+if test "$WITH_LOGIND" = "yes"; then
+  PKG_CHECK_MODULES([SYSTEMD], [libsystemd >= 254], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $SYSTEMD_CFLAGS"], [:])
+fi
+AC_SUBST([LOGIND_CFLAGS])
+AC_SUBST([SYSTEMD_LIBS])
+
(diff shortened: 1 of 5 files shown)


thkukuk:
The PR isn't ready yet as there is an open discussion: add two missing functions to systemd/sd-login.h or write a library which uses the DBUS interface to get the missing informations from logind. But it shows already how this could be done.

@ldv-alt @t8m what do you think about this idea to query logind instead of parsing utmp to solve Y2038 problem?

ldv-alt:
Makes sense.  FWiW, I'm for adding these two functions to libsystemd.

thkukuk:
All necessary PRs for systemd are accepted, this PR builds and works with current systemd from git.

ldv-alt:
The systemd from git currently claims that it's version is 253, while the configure check expects libsystemd >= 254,
so in its current form the logind support won't be activated until systemd 254-rc1 is out.

thkukuk:
> The systemd from git currently claims that it's version is 253, while the configure check expects libsystemd >= 254, so in its current form the logind support won't be activated until systemd 254-rc1 is out.

I think that's acceptable. If we require only 253, we would need additional checks to find out if that is an official 253 without the functions or an "inofficial" build from git with the functions for about 3-5 month. If somebody wants to use that already today, he can backport the functions and change the Linux-PAM configure check or use current git and change that version to 254 already (I personally would go the first option).

ldv-alt (approved):

kilobyte:
Instead of #ifdef .. #else, the code should try systemd if running, and if that fails, fall back to utmp.

PR #550, opened 2023-03-30
pam_timestamp: fix build failure

ikerexxe:
bcba17939e1b1a568cd4a764534cde74d37078cc started using pam_overwrite_n() without providing the definition to this function, which causes a build failure.

modules/pam_timestamp/hmac_openssl_wrapper.c: include pam_inline.h


ldv-alt (approved):
Ideally we should have this build scenario covered by CI tests.

ikerexxe:
I agree, but how can I do it? Should I enable `--enable-openssl` in the CI checks? That would leave the other HMAC implementation untested.

ldv-alt:
We should probably add something with `--enable-openssl` to the CI test matrix.

PR #554, opened 2023-04-04
pam_unix: fix compiling if HAVE_NIS is not set

thkukuk:
modules/pam_unix/yppasswd_xdr.c: Include config.h and don't compile the XDR functions until HAVE_NIS is set

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -459,7 +459,6 @@ AC_SUBST(LIBDB)
 
 AC_ARG_ENABLE([nis],
         AS_HELP_STRING([--disable-nis], [Disable building NIS/YP support in pam_unix]))
-AM_CONDITIONAL([HAVE_NIS], [test "x$enable_nis" != "xno"])
 
 AS_IF([test "x$enable_nis" != "xno"], [
   old_CFLAGS=$CFLAGS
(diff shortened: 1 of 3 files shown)


t8m:
Hmm why the:
```
if HAVE_NIS
  pam_unix_la_SOURCES += yppasswd_xdr.c
endif
```

snippet of Makefile.am does not work as expected?

thkukuk:
> snippet of Makefile.am does not work as expected?

Good question, all the other "if" do work...

ldv-alt:
btw, I added a couple of --disable-nis jobs to the ci matrix, and it passed (without this PR):
https://github.com/linux-pam/linux-pam/actions/runs/4608495662

thkukuk:
> btw, I added a couple of --disable-nis jobs to the ci matrix, and it passed (without this PR):

The error I'm getting is:
```
In file included from yppasswd_xdr.c:6:0:
yppasswd.h:9:10: fatal error: rpc/rpc.h: Datei oder Verzeichnis nicht gefunden
 #include <rpc/rpc.h>
          ^~~~~~~~~~~
compilation terminated.
```

So if your build environment contains rpc/rpc.h, it will build, independent of --disable-nis or not.
If I look at the generated Makefile: some if get's ignored, others not. But I didn't understand yet why.
Was there some change in automake?

ldv-alt:
In logs of those --disable-nis CI jobs yppasswd_xdr was mentioned only once:
```
rm -f ./.deps/yppasswd_xdr.Plo
```

I wonder how the relevant faulty part of the generated Makefile.in looks like, and what is the version of automake that produces that oddness.

thkukuk:
Ok:
configure --disable-nis works and with no rpc/rpc.h installed works.
configure without --disable-nis and with no rpc/rpc.h installed fails.

Seems the check if NIS is available or not in configure.in is broken.

ldv-alt:
Apparently, all those configure checks have no effect on HAVE_NIS, its value depends on --enable-nis/--disable-nis only.

ldv-alt (approved):

Issue #532, opened 2023-02-03
libpam rejects pam config files containing long lines


DDDas7:
Hi,
     
    Greetings!
     
    I would like to  highlight an issue related to the "pam_tty_audit.so" module. Login issue is observed if the number of comma separated user  lists is very large. If the complete string is larger than 1024 bytes  then "Permission denied" error is observed during the login example (su - <username>).
     
    ---------------------------------------------------------------------------
    session     required      pam_tty_audit.so disable=* enable=abc,xyz, ............, zzz
    ---------------------------------------------------------------------------
     
    the 1024 bytes limitation is hardcoded and  is coming from below file.
     
     - libpam/pam_handlers.c
     
       ---------------------------------------
       #define BUF_SIZE                  1024
       ---------------------------------------
     
    I would like to write a patch for the above issue  but would like to confirm the following.
     
    1) Will it be possible to increase BUF_SIZE to 4096  or dynamically to the default BUF_SIZE value to "getconf  ARG_MAX" which is 1/4th the size of stack size (uimit -s)?
     
    2) If solution in point 1 is not possible will it be fine to create a separate config file similar to faillock (/etc/security/faillock.conf) to customize the value.
     
    Any  suggestion / feedback will be helpful to resolve the above issue.
     
    Regards,
    Deepak Das

DeepakDas7:
Hi, 

Greetings!

Anyway update is available regarding  the above query?

Regards,
Deepak Das

ikerexxe:
@ldv-alt can you guide us on the best solution for this issue?

thkukuk:
By looking at the code, I'm against just increasing the buffer, as this does not solve the fundamental problem here, that _pam_assemble_line returns that the buffer is too small, but the calling function aborts on this.

Either _pam_assemble_line increases the buffer dynamically, or the calling function tries it again with a bigger buffer. But hardcoded buffer sizes should not be used anymore today.

DeepakDas7:
Hi Thorsten (@thkukuk),  

Thank you for your feedback!

As per your suggestion, I have  make the necessary changes and created  PR (https://github.com/linux-pam/linux-pam/pull/561)

Can you please review the same .

Thanks and Regards,
Deepak Das

ldv-alt:
This is not related to pam_tty_audit.so in any way, one could achieve the same effect with pam_permit.so

ldv-alt:
[...]

PR #561, opened 2023-04-13
libpam: raise line buffer size limit

DDDas7:
Resolves EACCES error due to 1024 buffer size while reading individual lines in pam files.

Resolves: https://github.com/linux-pam/linux-pam/issues/532

### DIFF
--- a/libpam/pam_handlers.c
+++ b/libpam/pam_handlers.c
@@ -25,7 +25,7 @@
 #define _PAM_ISA "."
 #endif
 
-static int _pam_assemble_line(FILE *f, char *buf, int buf_len);
+static int _pam_assemble_line(FILE *f, char **buf, size_t buf_len);
 
 static void _pam_free_handlers_aux(struct handler **hp);
 
@@ -62,12 +62,18 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 #endif /* PAM_READ_BOTH_CONFS */
     )
 {
-    char buf[BUF_SIZE];
     int x;                    /* read a line from the FILE *f ? */
+    char *buf = malloc(BUF_SIZE * sizeof(char));
+
+    if (buf == NULL) {
+        pam_syslog(pamh, LOG_CRIT, "malloc failed");
+        return PAM_BUF_ERR;
+    }
+
     /*
      * read a line from the configuration (FILE *) f
      */
-    while ((x = _pam_assemble_line(f, buf, BUF_SIZE)) > 0) {
+    while ((x = _pam_assemble_line(f, &buf, BUF_SIZE)) > 0) {
 	char *tok, *nexttok=NULL;
 	const char *this_service;
 	const char *mod_path;
@@ -204,6 +210,7 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 		    if (res != PAM_SUCCESS) {
 			pam_syslog(pamh, LOG_ERR, "error adding substack %s", tok);
 			D(("failed to load module - aborting"));
+			free(buf);
 			return PAM_ABORT;
 		    }
 		}
@@ -278,9 +285,14 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 	    if (res != PAM_SUCCESS) {
 		pam_syslog(pamh, LOG_ERR, "error loading %s", mod_path);
 		D(("failed to load module - aborting"));
+		free(buf);
 		return PAM_ABORT;
 	    }
 	}
+    }
+
+    if (buf != NULL) {
+        free(buf);
     }
 
     return ( (x < 0) ? PAM_ABORT:PAM_SUCCESS );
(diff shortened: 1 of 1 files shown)


ldv-alt:
This just raises the line buffer size limit from 1024 to 8192.

DDDas7:
Hi  Dimitry ( @ldv-alt ) ,

Greetings!

I have added a new patch. Can you please review the updated  patch .

Thank you!

Regards,
Deepak

libpam/pam_handlers.c
```diff
@@ -17,7 +17,7 @@
 #include <fcntl.h>
 #include <unistd.h>
 
-#define BUF_SIZE                  1024
+#define BUF_SIZE                  4096
```
ldv-alt, line 20: Since you're introducing a dynamic memory allocation, what's the reason of changing default?

libpam/pam_handlers.c
```diff
@@ -59,12 +59,14 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 #endif /* PAM_READ_BOTH_CONFS */
     )
 {
-    char buf[BUF_SIZE];
+    char *buf = (char *) malloc(BUF_SIZE * sizeof(char));
```
ldv-alt, line 62: malloc return values do not require a cast.

libpam/pam_handlers.c
```diff
@@ -76,6 +78,23 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 	char **argv;
 	int argvlen;
 
+	if (x == (mod_buf_size - 1)) {
+	    mod_buf_size *= 2;
+	    buf = (char *) realloc(buf, mod_buf_size * sizeof(char));
```
ldv-alt, line 83: This introduces a memory leak on error path.

libpam/pam_handlers.c
```diff
@@ -76,6 +78,23 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
...
+
+	    if (buf == NULL) {
+		pam_syslog(pamh, LOG_ERR,
+			   "Error Allocation Memory. Aborting! "
+		           "in file %s at line %d", __FILE__, __LINE__);
+                return PAM_ABORT;
```
ldv-alt, line 89: This should rather return `PAM_BUF_ERR`, see examples in this file.

libpam/pam_handlers.c
```diff
@@ -76,6 +78,23 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
...
+            }
+
+	    memset(buf, '\0', mod_buf_size * sizeof(char));
```
ldv-alt, line 92: Why `memset` is needed here?

libpam/pam_handlers.c
```diff
@@ -76,6 +78,23 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
...
+            file_pos -= x;
+            fseek(f, file_pos, SEEK_SET);
```
ldv-alt, line 94: Sorry, but this is not acceptable, there shouldn't be any re-reading of the file.

ldv-alt (changes requested):
I suggest changing `_pam_assemble_line` instead, for example, by using `getline` instead of `fgets` there.

DDDas7:
Hi Dimitry ( @ldv-alt ) ,

Greetings!

I have added a new patch. Can you please review the updated patch .

Thank you!

Regards,
Deepak

DDDas7:
Hi Dimitry ( @ldv-alt ) ,

Greetings!

I have added a new patch and committed the change..  Some tests are failing during checks. I am presently reviewing it and will update once the issue is fixed.

Thank you!

libpam/pam_handlers.c
```diff
@@ -278,11 +279,15 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
...
 		pam_syslog(pamh, LOG_ERR, "error loading %s", mod_path);
 		D(("failed to load module - aborting"));
+		free(buf);
 		return PAM_ABORT;
 	    }
 	}
     }
 
+    if (buf != NULL)
+        free(buf);
```
BenBE, line 289: `free(NULL);` is a no-op on any non-broken C library implementation.
– C99 standard, section 7.20.3.2/2

libpam/pam_handlers.c
```diff
@@ -576,37 +588,28 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
-	if (p >= endp - 1) {
-	    /* Overflow */
-	    D(("overflow"));
-	    return -1;
-	}
-	if (fgets(p, endp - p, f) == NULL) {
+	if (getline(buffer, &buf_len, f) <= 0) {
 	    if (used) {
 		/* Incomplete read */
+		free(*buffer);
```
stoeckmann, line 604: This leads to a double free when the while-loop in _pam_parse_conf_file is left. The free is performed after the while as well.

stoeckmann:
Also I'm not entirely sure if the code base is ready for arbitrarily long lines as it is right now. There are some cases where strlen's return value is stored in int, which would overflow with 2 GB lines.

I'm well aware that it's very very unlikely to encounter such lines, but to me it seems that not that many persons had problems with the 1024 character limit so far either.

stoeckmann:
Please see https://github.com/linux-pam/linux-pam/pull/658 for my attempt of solving it. It also fixes some issues with current configuration parsing.

PR #566, opened 2023-04-20
man: clarify PAM_DATA_SILENT in pam_set_data.3

bluca:
pam_end.3 has additional and important information about PAM_DATA_SILENT, copy it to pam_set_data.3 since that describes the cleanup callback where it will be set.

### DIFF
--- a/doc/man/pam_set_data.3.xml
+++ b/doc/man/pam_set_data.3.xml
@@ -113,7 +113,13 @@
           <para>
             Which indicates that the process would prefer to perform the
             <function>cleanup()</function> quietly. That is, discourages
-            logging/messages to the user.
+            logging/messages to the user. It is generally used to indicate that
+            the current closing of the library is in a
+            <citerefentry>
+              <refentrytitle>fork</refentrytitle><manvolnum>2</manvolnum>
+            </citerefentry>ed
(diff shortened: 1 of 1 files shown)


t8m (approved):

poettering:
Google pointed me to this piece of documentation first btw:

https://github.com/linux-pam/linux-pam/blob/master/doc/adg/Linux-PAM_ADG.xml#L232

Probably should receive similar treatment

bluca:
> Google pointed me to this piece of documentation first btw:
> 
> https://github.com/linux-pam/linux-pam/blob/master/doc/adg/Linux-PAM_ADG.xml#L232
> 
> Probably should receive similar treatment

That's `PAM_SILENT`, not `PAM_DATA_SILENT`, I think they are different things?

poettering:
oh.

(that said, the fact that pam_end() is supposed to be called from the child with PAM_DATA_SILENT still deserves to be mentioned there)

bluca:
> oh.
> 
> (that said, the fact that pam_end() is supposed to be called from the child with PAM_DATA_SILENT still deserves to be mentioned there)

The pam_end manpage already says that, and it is pulled in the PAM_ADG document when it gets built. The XML in the repository is only part of it, it is assembled when it's built and published.

PR #572, opened 2023-05-08
Various PATH_MAX fixes

pinotree:
`PATH_MAX` is optional in POSIX, and not defined on GNU/Hurd; hence, do some changes to be able to build PAM more on Hurd:
- pam_timestamp: check for its existence before trying to check it
- pam_pwhistory: use `asprintf()` to allocate buffers
- tests: define a fallback `PATH_MAX` if not available

### DIFF
--- a/libpam/include/test_assert.h
+++ b/libpam/include/test_assert.h
@@ -11,6 +11,7 @@
 #  include <config.h>
 # endif
 
+# include <limits.h>
 # include <stdio.h>
 # include <stdlib.h>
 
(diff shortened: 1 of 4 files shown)


modules/pam_env/tst-pam_env-retval.c
```diff
@@ -33,6 +33,10 @@ static const char usr_conf[] = TEST_NAME_DIR VENDOR_SCONFIGDIR "/pam_env.conf";
 
 static struct pam_conv conv;
 
+#ifndef PATH_MAX
+# define PATH_MAX 4096
+#endif
```
t8m, line 38: Maybe it could be added to test_assert.h so it is not necessary to repeat it everywhere?

pinotree: yeah, that was another possibility, and I wanted to send at least a version of fixes; changed, thanks

t8m (approved):

ldv-alt (approved):

PR #573, opened 2023-05-08
pam_nologin: explicitly fail when nologin path is a directory

pinotree:
On some systems (e.g. GNU/Hurd), `read()` succeeds on the fd of a directory; since the module assumes that `read()` fails (and thus `pam_modutil_read()` as well), manually fail in case the open fd refers to a directory.

### DIFF
--- a/modules/pam_nologin/pam_nologin.c
+++ b/modules/pam_nologin/pam_nologin.c
@@ -98,6 +98,16 @@ static int perform_check(pam_handle_t *pamh, struct opt_s *opts)
 	    goto clean_up_fd;
 	}
 
+	/*
+	 * on some OSes (e.g. Hurd) reading a directory succeeds,
+	 * instead of failing with EISDIR; hence, work as if
+	 * pam_modutil_read later on would fail
+	 */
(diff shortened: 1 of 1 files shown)


ldv-alt (commented):
After this change, a `fstat` error is treated as if there is no file.
I don't mind, but maybe add a comment or mention it somehow?

pinotree:
> After this change, a `fstat` error is treated as if there is no file. I don't mind, but maybe add a comment or mention it somehow?

ah I see what you mean: I moved the `fstat` block up as optimization, and that can give a different `retval`; moved it back to where it was, thanks!

ldv-alt (commented):
Looks good.

As the rest of the file is indented with tabs, could you reindent the new code with tabs, please?

pinotree:
> As the rest of the file is indented with tabs, could you reindent the new code with tabs, please?

Oops, done, thanks for noticing. The mixup of tabs and spaces to indent around the sources makes things confusing...

ldv-alt (approved):

t8m (approved):

PR #574, opened 2023-05-08
configure.ac: add `--enable-examples` option

pinotree:
Allow the user to not build the examples through `--disable-examples` (enabled by default); this can be useful:
- when cross-compiling, as the examples are not useful
- in distribution builds, not building stuff that is not used in any way


ldv-alt (changes requested):
The commit says `--enable-doc` while the option being added is `--disable-examples`.

ldv-alt (approved):

PR #576, opened 2023-05-11
examples/tty_conv: fix build failure on musl

vi-improved:
termio.h is the old System V version of the interface header, and is only provided in glibc and dietlibc as far as I can tell. This fixes it to instead use the POSIX termios.h.


ldv-alt (changes requested):
Please copy the PR description to the commit message.

vi-improved:
Done.

ldv-alt (approved):

PR #579, opened 2023-05-15
pam_xauth: switch away from PATH_MAX

pinotree:
Allocate the path buffer in `check_acl()` dynamically using `asprintf()`, so there is no need to use a PATH_MAX-fixed size buffer. The fallback PATH_MAX definition is no more needed, thus is dropped.

Make sure that paths too long still result in `PAM_SESSION_ERR`.

### DIFF
--- a/modules/pam_xauth/pam_xauth.c
+++ b/modules/pam_xauth/pam_xauth.c
@@ -73,11 +73,6 @@
 #define XAUTHDEF ".Xauthority"
 #define XAUTHTMP ".xauthXXXXXX"
 
-/* Hurd compatibility */
-#ifndef PATH_MAX
-#define PATH_MAX 4096
-#endif
-
 /* Possible paths to xauth executable */
 static const char * const xauthpaths[] = {
 #ifdef PAM_PATH_XAUTH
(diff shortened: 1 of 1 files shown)


pinotree:
Note there is a potential behaviour change: in case the resulting `path` is longer than `PATH_MAX`, or in general too long to cause `stat()` to fail with `ENAMETOOLONG`, the return value of `check_acl()` changes from `PAM_SESSION_ERR` (because of the former `i >= (int)sizeof(path)` check) to `PAM_PERM_DENIED` (because of the error handling block at the end of the function).

In case the previous return value is needed, I can add an additional error handling block for `ENAMETOOLONG`.

ldv-alt:
Yes, according to the documentation, the return value has to be `PAM_SESSION_ERR` in case of `ENAMETOOLONG`.

pinotree:
> Yes, according to the documentation, the return value has to be `PAM_SESSION_ERR` in case of `ENAMETOOLONG`.

Thanks! Fixed accordingly.

modules/pam_xauth/pam_xauth.c
```diff
@@ -350,12 +351,19 @@ check_acl(pam_handle_t *pamh,
 						   path);
 				}
 			}
+			free(path);
 			return noent_code;
+		case ENAMETOOLONG:
+			pam_syslog(pamh, LOG_ERR,
+				   "too long path: %s", path);
```
ldv-alt, line 358: Let's use the same wording as in the default case:
```
pam_syslog(pamh, LOG_ERR, "error opening %s: %m", path);
```

ldv-alt (changes requested):

ldv-alt (approved):

Issue #483, opened 2022-08-18
pam_shells returns PAM_AUTH_ERR for unknown users, leading to the username being logged


TheJonny:
in archllinux, pam_shells is enabled for `login` `auth`. When accidentally typing the password as username, it is judged as PAM_AUTH_ERR by this module, leading to a log entry by login.

pam_shells is behaving exactly like documented.
But other modules return PAM_USER_UNKNOWN, so that login can hide the username in the log entry.

ldv-alt:
Maybe `pam_shells` could be taught to to return `PAM_USER_UNKNOWN` in that case.

TheJonny:
I grepped for pam_modutil_getpwnam and summarized the occurrences here: [getpwnam.txt](https://github.com/linux-pam/linux-pam/files/11482140/getpwnam.txt)

If I see it correctly, also
- [ ] pam_timestamp,
- [ ] pam_sepermit
- [ ] pam_listfile 

are affected. (as is pam_succede_if in #559, but that one is a bit harder and already has an issue)

PR #580, opened 2023-05-15
pam_shells: return PAM_USER_UNKNOWN if getpwnam fails

TheJonny:
Until before, in this case PAM_AUTH_ERR was returned. This leads to unknown users being logged with the unknown username.
Now it resembles the behavior of other modules like pam_unix in this case.

see #483

### DIFF
--- a/modules/pam_shells/pam_shells.8.xml
+++ b/modules/pam_shells/pam_shells.8.xml
@@ -75,6 +75,14 @@
         </listitem>
       </varlistentry>
       <varlistentry>
+        <term>PAM_USER_UNKNOWN</term>
+        <listitem>
+          <para>
+            The user does not exist or the user's login shell could not be determined.
+          </para>
+        </listitem>
+      </varlistentry>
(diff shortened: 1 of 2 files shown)


modules/pam_shells/pam_shells.8.xml
```diff
@@ -74,6 +74,14 @@
           </para>
         </listitem>
       </varlistentry>
+      <varlistentry>
+        <term>PAM_USER_UNKNOWN</term>
+        <listitem>
+          <para>
+            The does not exist or the user's login shell could not be determined.
```
ldv-alt, line 81: Did you mean "The _user_ does not exist"?

ldv-alt (changes requested):

modules/pam_shells/pam_shells.8.xml
TheJonny: you are right

ldv-alt (approved):

PR #584, opened 2023-06-02
pam_unix: improve fallback values for "rounds" for yescrypt and blowfish

kranzj:
This change improves the fallback values for the "rounds" parameter for yescrypt and blowfish by using the smallest reasonable value if the users sets a too low value and by using the highest reasonable value if the user sets a too high value. This better realizes user intent and is consistent with the approach taken for SHA256.


ldv-alt (approved):
Looks good, although this would be backwards incompatible in case of unreasonably large values specified for `rounds` parameter.

PR #586, opened 2023-06-07
Fix build if crypt_r isn't available

berolinux:
retval was being defined only in #ifdef HAVE_CRYPT_R, but used unconditionally.


ldv-alt:
Could you add
```
Fixes: bcba17939e1b ("modules: make use of secure memory erasure")
```
to the commit message, please

berolinux:
done

ldv-alt (approved):

PR #588, opened 2023-07-10
pam_userdb: enable GDBM support

ikerexxe:
* configure.ac: add `gdbm` option to `enable-db`
* modules/pam_userdb/pam_userdb.c: conditionally provide database access depending on the database technology

### DIFF
--- a/modules/pam_userdb/pam_userdb.c
+++ b/modules/pam_userdb/pam_userdb.c
@@ -25,6 +25,8 @@
 
 #ifdef HAVE_NDBM_H
 # include <ndbm.h>
+#elif defined(HAVE_GDBM_H)
+# include <gdbm.h>
 #else
 # ifdef HAVE_DB_H
 #  define DB_DBM_HSEARCH    1 /* use the dbm interface */
@@ -39,6 +41,12 @@
 #include <security/pam_ext.h>
 #include <security/_pam_macros.h>
 #include "pam_inline.h"
+
+#ifndef HAVE_GDBM_H
+# define COND_UNUSED UNUSED
+#else
+# define COND_UNUSED
+#endif /* HAVE_GDBM_H */
 
 /*
  * Conversation function to obtain the user's password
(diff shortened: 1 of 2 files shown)


modules/pam_userdb/pam_userdb.c
```diff
@@ -129,6 +131,76 @@ _pam_parse (pam_handle_t *pamh, int argc, const char **argv,
...
+    return key;
+}
+
+static datum
+db_nextkey(void *dbm, datum key UNUSED)
+{
+    datum nextkey;
+
+#ifdef HAVE_GDBM_H
+    nextkey = gdbm_nextkey(dbm, key);
```
ldv-alt, line 171: Sorry, how can `key` be `UNUSED` if it is passed down to `gdbm_nextkey`?

ikerexxe: It's unused for the non-gdbm option, where `key` isn't used. If I wouldn't have set the `UNUSED` keyword, then the CI would have failed. However, I'm open to other solutions.

ldv-alt: This means that the use of `UNUSED` keyword also has to be conditional.

ikerexxe: Done

modules/pam_userdb/pam_userdb.c
```diff
@@ -129,6 +137,76 @@ _pam_parse (pam_handle_t *pamh, int argc, const char **argv,
...
+    datum key;
+
+#ifdef HAVE_GDBM_H
+    key = gdbm_firstkey(dbm);
+#else
+    key = dbm_firstkey(dbm);
+#endif /* HAVE_GDBM_H */
+
+    return key;
+}
```
ldv-alt, line 169: In all these new wrapper functions, do they really need that level of complexity?
For example, `db_firstkey` could be simplified to

```
static datum
db_firstkey(void *dbm)
{
#ifdef HAVE_GDBM_H
    return gdbm_firstkey(dbm);
#else
    return dbm_firstkey(dbm);
#endif
}
```

ikerexxe: Updated

modules/pam_userdb/pam_userdb.c
```diff
@@ -129,6 +137,62 @@ _pam_parse (pam_handle_t *pamh, int argc, const char **argv,
...
+db_open(const char *database, mode_t file_mode)
+{
+#ifdef HAVE_GDBM_H
+    void *dbm = gdbm_open(database, 4096, GDBM_READER, file_mode, NULL);
+#else
+    void *dbm = dbm_open(database, O_RDONLY, file_mode);
+#endif /* HAVE_GDBM_H */
+
+    return dbm;
+}
```
ldv-alt, line 153: Do you really need `dbm` variable here?

ikerexxe: Not at all. Thanks for catching that.

ldv-alt (approved):

PR #591, opened 2023-07-27
Fix up formatting of audit messages

stevegrubb:
Pam_faillock uses audit_log_user_message to write to the audit system. It does not take an op argument, so you have to add one yourself. Otherwise the pam_faillock part of the message is lost because it's not in key=value format.

Also, we can't use uid in that event because the kernel already adds that field. What we normally do is use 'suid' (meaning sender uid) as the field name.


ldv-alt:
Apparently, the message logged by pam_selinux also doesn't quite follow the key=value format:
```
pam: default-context=%s selected-context=%s
```
Should it be changed to something like
```
op=pam_selinux default-context=%s selected-context=%s
```
?

stevegrubb:
Sure. I pushed another commit to this PR.

ldv-alt (approved):

PR #593, opened 2023-07-29
pam_pwhistory: fix passing NULL filename argument to pwhistory helper

zainhasib:
This bug was observed in an environment where SELinux was enabled and no file override was passed to `pam_pwhistory.so` in pam config. That causes a `NULL` value to be passed as an argument when calling `pwhistory_helper` binary, short-circuiting the arguments to just three.

The changes introduced handle the default case when the file is not overridden and passed as an empty string rather than a `NULL` value so that `pwhistory_helper` binary when called via `execve` can read all the arguments passed.


ldv-alt:
Fixes the regression introduced by commit 11c35109a67fa82b75f8d427fcb5ab41f61550f7.

ldv-alt (approved):

PR #597, opened 2023-08-05
Remove memory leaks in tests

cgzones:
* tests: free handles via pam_end()
  Destroy the pam handles via pam_end() to release all associated resources.  This allows to run the test-suite with sanitizers and validates the resource cleanup in pam_end() and callees.

* pam_start: free handlers on handler init failure
  If the pam handlers fail to initialize halfway, clean them up afterwards.  Since we set the handle to NULL callers can't clean them.

* pam_faillock: free handle in test

* ci: add GCC 13 job with sanitzers

### DIFF
--- a/tests/tst-pam_fail_delay.c
+++ b/tests/tst-pam_fail_delay.c
@@ -67,6 +67,8 @@ main (void)
       return 1;
     }
 
+  pam_end (pamh, retval);
+
   /* 2: use NULL pam handle */
   retval = pam_fail_delay (NULL, 60);
   if (retval == PAM_SUCCESS)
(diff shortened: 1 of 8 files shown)


tests/tst-pam_mkargv.c
```diff
@@ -50,5 +50,7 @@ int main(void)
 	return 1;
     }
 
+  free(myargv);
+
```
ldv-alt, line 54: This looks like a different fix not related to `pam_end()`.

tests/tst-pam_fail_delay.c
```diff
@@ -67,6 +67,8 @@ main (void)
       return 1;
     }
 
+  pam_end (pamh, retval | PAM_DATA_SILENT);
```
ldv-alt, line 70: Is there any particular reason to use `PAM_DATA_SILENT` in these added `pam_end()` invocations?

PR #186, opened 2020-03-11
pam_access: Support UID and GID in access.conf

blueskycs2c:
pam_access: Support UID and GID in access.conf fix [#114](https://github.com/linux-pam/linux-pam/issues/114).

eg:access.conf
-:ALL EXECPT (1000) 1002 :LOCAL

Issue #114, opened 2019-05-20
pam_access: Support UID and GID in access.conf


kevinptt0323:
It would be great if pam_access supports UID and GID just like sudoers. For sudoers, we can use `#1000` to represent a user with UID 1000.

```
User ::= '!'* user name |
         '!'* #uid |
         '!'* %group |
         '!'* %#gid |
         '!'* +netgroup |
         '!'* %:nonunix_group |
         '!'* %:#nonunix_gid |
         '!'* User_Alias
```

Ref: [Sudoers Manual](https://www.sudo.ws/man/1.8.13/sudoers.man.html#Aliases)

PR #601, opened 2023-08-08
pam_access: Support UID and GID in access.conf

mluckam:
Following up on #114 and #186.  It appears that the original author, @blueskycs2c, has been inactive for several years.  The change has been [rebased as requested](https://github.com/linux-pam/linux-pam/pull/186#issuecomment-1650565201) by @t8m.

### DIFF
--- a/modules/pam_access/pam_access.c
+++ b/modules/pam_access/pam_access.c
@@ -605,7 +605,30 @@ netgroup_match (pam_handle_t *pamh, const char *netgroup,
   return retval;
 }
 
-/* user_match - match a username against one token */
+/* user_name_or_uid_match - match a username or user uid against one token */
+static int
+user_name_or_uid_match(pam_handle_t *pamh, const char *tok,
+		       const struct login_info *item)
+{
+    /* ALL or exact match of username */
+    int rv = string_match(pamh, tok, item->user->pw_name, item->debug);
+    if (rv != NO)
+	return rv;
+
+    if (tok[strspn(tok, "0123456789")] != '\0')
+	return NO;
+
+    char buf[sizeof(long long) * 3 + 1];
+    snprintf(buf, sizeof(buf), "%llu",
+	     zero_extend_signed_to_ull(item->user->pw_uid));
+    if (item->debug)
+	pam_syslog(pamh, LOG_DEBUG, "user_match: tok=%s, uid=%s", tok, buf);
+
+    /* check for exact match of uid */
+    return string_match (pamh, tok, buf, item->debug);
+}
+
+/* user_match - match a user against one token */
 
 static int
 user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
@@ -656,7 +679,7 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
 		hostname = item->hostname;
 	}
         return (netgroup_match (pamh, tok + 1, hostname, string, item->debug));
-    } else if ((rv=string_match (pamh, tok, string, item->debug)) != NO) /* ALL or exact match */
+    } else if ((rv=user_name_or_uid_match(pamh, tok, item)) != NO) /* ALL or exact match */
       return rv;
     else if (item->only_new_group_syntax == NO &&
 	     pam_modutil_user_in_group_nam_nam (pamh,
@@ -667,6 +690,36 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
     return NO;
 }
 
+
+/* group_name_or_gid_match - match a group name or group gid against one token */
+static int
+group_name_or_gid_match(pam_handle_t *pamh, const char *tok,
+			const char *usr, int debug)
+{
+    /* check for exact match of group name */
+    if (pam_modutil_user_in_group_nam_nam(pamh, usr, tok) != NO)
+	return YES;
+
+    if (tok[strspn(tok, "0123456789")] != '\0')
+	return NO;
+
+    char *endptr = NULL;
+    errno = 0;
+    unsigned long int ul = strtoul(tok, &endptr, 10);
+    gid_t gid = (gid_t) ul;
+    if (errno != 0
+	|| tok == endptr
+	|| *endptr != '\0'
+	|| (unsigned long) zero_extend_signed_to_ull(gid) != ul) {
+	return NO;
+    }
+
+    if (debug)
+	pam_syslog(pamh, LOG_DEBUG, "group_match: user=%s, gid=%s", usr, tok);
+
+    /* check for exact match of gid */
+    return pam_modutil_user_in_group_nam_gid(pamh, usr, gid);
+}
 
 /* group_match - match a username against token named group */
 
@@ -684,10 +737,10 @@ group_match (pam_handle_t *pamh, char *tok, const char* usr, int debug)
     tok++;
     tok[strlen(tok) - 1] = '\0';
 
-    if (pam_modutil_user_in_group_nam_nam(pamh, usr, tok))
+    if (group_name_or_gid_match (pamh, usr, tok, debug))
         return YES;
 
-  return NO;
+    return NO;
 }
 
 
--- a/libpam/include/pam_inline.h
+++ b/libpam/include/pam_inline.h
@@ -45,6 +45,26 @@
 #define PAM_ARRAY_SIZE(a_)		(sizeof(a_) / sizeof((a_)[0]) + PAM_MUST_BE_ARRAY(a_))
 
 /*
+ * Zero-extend a signed integer type to unsigned long long.
+ */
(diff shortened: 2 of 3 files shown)


modules/pam_access/pam_access.c
```diff
@@ -599,14 +599,25 @@ netgroup_match (pam_handle_t *pamh, const char *netgroup,
...
+
+     memset(buf, 0, BUFSIZ);
+     /* tok is uid */
+     if (strspn(tok, "0123456789") == strlen(tok)) {
+	 sprintf(buf, "%d", item->user->pw_uid);
+	 if (item->debug)
+		pam_syslog(pamh, LOG_DEBUG, "user_match: tok=%s, uid=%d", tok, item->user->pw_uid);
+	 string = buf;
+     } else
+	 string = item->user->pw_name;
```
ldv-alt, line 620: This adds a code that will be executed in all cases including those where it won't be used.
Let's create a new function, e.g. user_or_uid_match, and use it in instead of
```
if ((rv=string_match (pamh, tok, string, item->debug)) != NO)
```

ldv-alt (changes requested):

modules/pam_access/pam_access.c
mluckam: The line in question above `if ((rv=string_match (pamh, tok, string, item->debug)) != NO)`
https://github.com/linux-pam/linux-pam/blob/a3c70778c49f1b0ebb93ea2c51df8fc72fe076eb/modules/pam_access/pam_access.c#L659
is not modified as part of this PR.  Can you clarify on this comment?

ldv-alt: The only place in `user_match()` where uid support is applicable is exactly that `string_match()` invocation I'm talking about, so I suggest changing that invocation instead of adding the code unconditionally in the beginning of `user_match()`.

mluckam: Thank you for the clarification.  Is the uid support not also applicable to the call to `group_match()`?  If a uid is provided, the uid is passed instead of the name.
https://github.com/linux-pam/linux-pam/blob/a3c70778c49f1b0ebb93ea2c51df8fc72fe076eb/modules/pam_access/pam_access.c#L633

mluckam: @ldv-alt thoughts on the above comment?

mluckam: @t8m @ldv-alt thoughts on the above comment?

ldv-alt: Yes, numeric uid support is also applicable to that call to `group_match`, but this patch doesn't add such support.

mluckam: Added new method user_name_or_uid_match.  I'll resolve this conversation, let me know if there is additional feedback.

mluckam:
@ldv-alt the requested new function was created.

mluckam:
The requested changes have been made, mind taking a look?
@t8m 
@ldv-alt
@BenBE 
@thkukuk 
@stoeckmann

modules/pam_access/pam_access.c
```diff
@@ -605,7 +605,31 @@ netgroup_match (pam_handle_t *pamh, const char *netgroup,
   return retval;
 }
 
-/* user_match - match a username against one token */
+/* user_name_or_uid_match - match a username or user uid against one token */
+static int
+user_name_or_uid_match (pam_handle_t *pamh, const char *tok, const struct login_info *item)
```
BenBE, line 610: For consistency this should keep its name or `group_match` below should be ranemed accordingly too.

mluckam: The method `user_match` was not renamed to `user_name_or_uid_match`.  Instead within `user_match`, the call to `string_match` was replaced with `user_name_or_uid_match`.  Since `group_match` and `user_match` already existed I did not want to rename these functions and potentially make breaking changes elsewhere.

The existing function `group_match` could be renamed to `group_name_or_gid_match`, but the new function `user_name_or_uid_match` cannot be named `user_match` due to a function of that name already existing.  Welcome thoughts on how to proceed.

BenBE: Ah, missed that indirect call/split of `user_match`. Would a similar change for `group_match` be reasonable?

mluckam: Like the idea.  It provides a standard approach for both methods.  Pushed the updates, let me know what you think.

modules/pam_access/pam_access.c
```diff
@@ -605,7 +605,31 @@ netgroup_match (pam_handle_t *pamh, const char *netgroup,
...
+{
+    int rv;
+
+    //check for exact match of username
```
ldv-alt, line 614: There were no c++ style comments in this file, let's not start using them here.

modules/pam_access/pam_access.c
```diff
@@ -605,7 +605,31 @@ netgroup_match (pam_handle_t *pamh, const char *netgroup,
...
+    if((rv=string_match (pamh, tok, item->user->pw_name, item->debug)) != NO) {
+        return rv;
+    }
+
+    //check for exact match of uid
+    if (strspn(tok, "0123456789") == strlen(tok)) {
```
ldv-alt, line 620: I'd prefer
```
if (tok[strspn(tok, "0123456789")] == '\0') {
```

modules/pam_access/pam_access.c
```diff
@@ -605,7 +605,31 @@ netgroup_match (pam_handle_t *pamh, const char *netgroup,
...
+        char buf[BUFSIZ];
+        sprintf(buf, "%d", item->user->pw_uid);
```
ldv-alt, line 622: Is it correct to treat uid as a signed integer here?

modules/pam_access/pam_access.c
```diff
@@ -684,10 +732,11 @@ group_match (pam_handle_t *pamh, char *tok, const char* usr, int debug)
     tok++;
     tok[strlen(tok) - 1] = '\0';
 
-    if (pam_modutil_user_in_group_nam_nam(pamh, usr, tok))
+    if(group_name_or_gid_match(pamh, usr, tok, debug)) {
```
ldv-alt, line 735: Please keep the space between `if` and `(`.

ldv-alt (changes requested):

modules/pam_access/pam_access.c
```diff
@@ -668,6 +692,30 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
...
+{
+    int rv;
+
+    if ((rv=pam_modutil_user_in_group_nam_nam(pamh, usr, tok)) != NO) {
+        return rv;
+    }
+
+    /* tok is GID */
+    if (strspn(tok, "0123456789") == strlen(tok)) {
+        gid_t grpgid = atoi(tok);
```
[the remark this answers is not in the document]
mluckam: I used `sscanf`.  Let me know if you have a preferred strategy.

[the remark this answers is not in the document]
ldv-alt: I suggest using `strtoul` here, it allows proper error checking.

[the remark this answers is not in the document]
mluckam: Updated to use `strtoul`.  Used [pam_sepermit](https://github.com/linux-pam/linux-pam/blob/master/modules/pam_sepermit/pam_sepermit.c#L199-L200) as an example.

modules/pam_access/pam_access.c
```diff
@@ -605,7 +606,30 @@ netgroup_match (pam_handle_t *pamh, const char *netgroup,
...
 
-/* user_match - match a username against one token */
+/* user_name_or_uid_match - match a username or user uid against one token */
+static int
+user_name_or_uid_match (pam_handle_t *pamh, const char *tok, const struct login_info *item)
+{
+    /* check for exact match of username */
+    if(string_match (pamh, tok, item->user->pw_name, item->debug) != NO) {
+        return YES;
+    }
```
ldv-alt, line 616: `string_match` can return other values besides `YES` and `NO`, so this would be a change of behavior.

modules/pam_access/pam_access.c
```diff
@@ -668,6 +692,40 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
...
+        return NO;
+    }
+
+    char *endptr = NULL;
+    errno = 0;
+    unsigned long int grpgid = strtoul(tok, &endptr, 10);
+    if(errno != 0
+    || tok == endptr
+    || grpgid > UINT_MAX
+    ) {
```
ldv-alt, line 716: Something went wrong with the indentation.
Also, `*endptr != '\0'` check is missing.

ldv-alt (changes requested):

ldv-alt (approved):

Issue #602, opened 2023-08-15
pam_strerror() fail to localize strings when linux-pam is installed into non-standard prefix (missing bindtextdomain() call)


trofi:
In https://github.com/NixOS/nixpkgs/issues/249010 rypervenche nitoced that in `nixpkgs` `su` lacks expected translation of a login prompt.

`nixpkgs` installs every package into it's own unique prefix and never overlays directoryes from different packages. As a result for `su` we have at least 3 locale directories:
- `shadow`: /nix/store/ymcln55n92xm25sk2pipqbcp5xghwc3j-shadow-4.13/share/locale/
- `glibc`: /nix/store/ibp4camsx1mlllwzh32yyqcq2r2xsy1a-glibc-2.37-8/share/locale
- `linux-pam`: /nix/store/k1lvsb3vyw2ijk9naamnay5nwc4mibda-linux-pam-1.5.2/share/locale

To find the locale one need to call `bindtextdomain()` or rely on a default.
- `shadow` calls `(void) bindtextdomain (PACKAGE, LOCALEDIR);` in `src/su.c` (ok)
- `glibc` uses default textdomain search path within `glibc` itself, no need to bind it explicitly (ok)
- `linux-pam`: does not bind it's directory anywhere (bad)

At least `glibc` runs `bindtextdomain ()` under a lock, thus should be safe to run it in `pam_start()` or `pam_strerror()`.

WDYT of adding a form of `(void) bindtextdomain (PACKAGE, LOCALEDIR);` call for `linux-pam`-specific `LOCALEDIR` as well? Other libraries do it in library initialization routines in static constructor or at first session start.

Thanks!

trofi:
Possible naive solution: https://github.com/linux-pam/linux-pam/pull/604

ldv-alt:
Implemented in #604

PR #604, opened 2023-08-16
pam_start.c: call bindtextdomain() to expose Linux-PAM locales

trofi:
Without the change locales are not visible to applications using PAM if PAM library is installed into a --prefix= different from the default one.


trofi:
https://github.com/linux-pam/linux-pam/issues/602 describes an issue in more detail if more context needed.

libpam/pam_start.c
```diff
@@ -25,6 +25,21 @@ static int _pam_start_internal (
     D(("called pam_start: [%s] [%s] [%p] [%p]"
        ,service_name, user, pam_conversation, pamh));
 
+    /* Bind text domain to pull in PAM transpations for a case where
```
ldv-alt, line 28: typo: `transpations`

libpam/pam_start.c
```diff
@@ -25,6 +25,21 @@ static int _pam_start_internal (
...
+       has a chance to have some overhead. Let's try to do it once (or a
+       small number of times as `bound_text_domain` is not protected by
+       a lock. */
+    static int bound_text_domain = 0;
+    if (!bound_text_domain) {
+	bound_text_domain = 1;
+#ifdef HAVE_BINDTEXTDOMAIN
+	bindtextdomain(PACKAGE, LOCALEDIR);
+#endif
+    }
```
ldv-alt, line 41: Let's move the whole hunk under `#ifdef HAVE_BINDTEXTDOMAIN`.

ldv-alt (changes requested):

libpam/pam_start.c
trofi: Fixed typo.

libpam/pam_start.c
trofi: Moved all under `#ifdef HAVE_BINDTEXTDOMAIN`

ldv-alt (approved):

Issue #605, opened 2023-08-16
pam_env: Allow parsing env for auth stage


mitfik:
Hi, 

I was trying to follow documentation and make `pam_env` working with `auth` method:

as per manpage:
> MODULE TYPES PROVIDED         [top](https://man7.org/linux/man-pages/man8/pam_env.8.html#top_of_page)
       The auth and session module types are provided.

but following configuration does not seems to work:

> auth required pam_env.so 

The use cases which I am trying to achieve with this is to allow user to provide `ENV Variable` before authentication process would start, so module can send push notification to the user without knowing him up front and ask for approval from his second device to get in (basically sign challenge with private key).

It seems that even some distro provide `auth required pam_env.so` as default configuration: 
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/4/html/reference_guide/s1-pam-sample-simple

But looking into the code it does not seems that this is supported: [L1021](https://github.com/linux-pam/linux-pam/blob/81d5593e32ef06e749ab8c591111d37b95c42e01/modules/pam_env/pam_env.c#L1021)

Could not find any reasonable answer and concluded with one of the below:
- distros have some magic patch which I can't find 
- everyone is wrong about support for `auth required pam_env` and this was never implemented
- I don't understand how to enable it ...

To be sure I create a patch to test that on my own and seems that with code modification behavior is correct.

Is there any reason why authentication was never implemented? Any security risks?

ldv-alt:
The `auth` stack is not just `pam_authenticate` but also `pam_setcred`. That is, `auth required pam_env` doesn't affect `pam_authenticate` but does affect `pam_setcred`.

thkukuk:
@mitfik I don't understand this request:

`The use cases which I am trying to achieve with this is to allow user to provide ENV Variable before authentication process would start, so module can send push notification to the user without knowing him up front and ask for approval from his second device to get in (basically sign challenge with private key).`

So the user sets the environment variable upfront and it is consumed by PAM modules? Beside that this is a huge security problem, you don't need pam_env for this.
If it is for communication between PAM modules: never use environment variables for this, please look at `pam_set_data()`.

Please specify your use case more precisely.

PR #606, opened 2023-08-16
pam_env: enable for pam_authenticate

mitfik:
Addressing #605


ldv-alt (changes requested):
The auth stack is not just `pam_authenticate` but also `pam_setcred`. That is, `auth required pam_env` doesn't currently affect `pam_authenticate` but does affect `pam_setcred`.  Since changing this would affect backwards compatibility, I suggest adding an option instead of changing the default behavior.

PR #608, opened 2023-09-21
examples/tty_conv: fix build on musl

skychengzhixing:
Using__ GLIBC__ This macro distinguishes between calling different termios.h for compiling with Musl

### DIFF
--- a/examples/tty_conv.c
+++ b/examples/tty_conv.c
@@ -6,7 +6,7 @@
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
-#include <termios.h>
+#include <asm/termios.h>
 #include <security/pam_appl.h>
 #include <sys/ioctl.h>
 


skychengzhixing:
@ldv-alt Could you please help me check this submission?

examples/tty_conv.c
```diff
@@ -6,7 +6,11 @@
 #include <string.h>
 #include <errno.h>
 #include <unistd.h>
+#ifdef __GLIBC__
 #include <termios.h>
+#else
+#include <asm-generic/termios.h>
+#endif
```
ldv-alt, line 13: Could you elaborate what is the build issue you are fixing, please?

By the way, it seems to me that it should rather be
```
#include <asm/termios.h>
```
unconditionally.

ldv-alt (changes requested):

examples/tty_conv.c
thkukuk: I assume musl libc has no termios.h header.

In any case, the logic is wrong. termios.h is not glibc specific but a standard. If an implementation does not support it, we need a workaround for this specific implementation and not break everything else.

thkukuk (changes requested):

examples/tty_conv.c
```diff
@@ -17,7 +21,7 @@
  ***************************************/
 static void echoOff(int fd, int off)
 {
-    struct termios tty;
+    struct termio tty;
```
thkukuk, line 24: termio is deprecated, termios is the successor.
If you have an implementation, which does not support current standards, make an exception for this implementation, but not for everything else.

thkukuk (changes requested):

skychengzhixing:
When I compile the libpam package using Musl's cross compilation toolchain, the following error will be reported:
| ../../Linux-PAM-1.5.3/examples/tty_conv.c:9:10: fatal error: termio.h: No such file or directory
|     9 | #include <termio.h>
|       |          ^~~~~~~~~~
| compilation terminated.
| make[2]: *** [Makefile:491: tty_conv.o] Error 1

I add #include <asm-generic/termios.h>，I added this # include<asm generic/termios. h> because the struct termio is in this header file of the Musl cross toolchain.

When I compile using the latest branch: master, I will report the following error:
../../Linux-PAM-1.5.3/examples/tty_conv.c: In function 'echoOff':
| ../../Linux-PAM-1.5.3/examples/tty_conv.c:20:9: warning: implicit declaration of function 'ioctl' [-Wimplicit-function-declaration]
|    20 |     if (ioctl(fd, TCGETA, &tty) < 0)
|       |         ^~~~~
| ../../Linux-PAM-1.5.3/examples/tty_conv.c:20:19: error: 'TCGETA' undeclared (first use in this function)
|    20 |     if (ioctl(fd, TCGETA, &tty) < 0)
|       |                   ^~~~~~
| ../../Linux-PAM-1.5.3/examples/tty_conv.c:20:19: note: each undeclared identifier is reported only once for each function it appears in
| ../../Linux-PAM-1.5.3/examples/tty_conv.c:29:23: error: 'TCSETAF' undeclared (first use in this function)
|    29 |         if (ioctl(fd, TCSETAF, &tty) < 0)
|       |                       ^~~~~~~
| ../../Linux-PAM-1.5.3/examples/tty_conv.c:37:23: error: 'TCSETAW' undeclared (first use in this function)
|    37 |         if (ioctl(fd, TCSETAW, &tty) < 0)
|       |                       ^~~~~~~
| make[2]: *** [Makefile:491: tty_conv.o] Error 1

If we use the latest struct termios, we still need to modify the path of the header file. I have made the changes on PR and the compilation is successful.

examples/tty_conv.c
ldv-alt: The code uses a few ioctls, namely, TCGETA, TCSETAF, and TCSETAW, which makes the whole program non-portable.
I suppose when the code uses a syscall interface like this, it should rather use kernel headers and the data types defined there.  Or better use a standard interface like termios(3) instead.

ldv-alt:
I'm pretty sure that
```
#include <termios.h>
```
compiles on every properly configured musl system, so it looks like you have a local misconfiguration, please fix that first.

ldv-alt:
I think this is a non-issue.

PR #609, opened 2023-09-21
pam_wheel: introduce auth_self mode

lnussel:
Some Linux vendors use the wheel group to make users in that group authenticate with their own password as root. Ie sudo mode.

### DIFF
--- a/modules/pam_wheel/pam_wheel.c
+++ b/modules/pam_wheel/pam_wheel.c
@@ -51,6 +51,7 @@
 #define PAM_TRUST_ARG       0x0004
 #define PAM_DENY_ARG        0x0010
 #define PAM_ROOT_ONLY_ARG   0x0020
+#define PAM_AUTH_SELF       0x0040
 
 static int
 _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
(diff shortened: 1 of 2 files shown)


modules/pam_wheel/pam_wheel.c
```diff
@@ -76,13 +77,20 @@ _pam_parse (const pam_handle_t *pamh, int argc, const char **argv,
...
+               ctrl |= PAM_AUTH_SELF;
 	  else if ((str = pam_str_skip_prefix(*argv, "group=")) != NULL)
 	       strncpy(use_group, str, group_length - 1);
           else {
                pam_syslog(pamh, LOG_ERR, "unknown option: %s", *argv);
           }
      }
 
+     if ((ctrl & (PAM_DENY_ARG|PAM_TRUST_ARG)) && (ctrl & PAM_AUTH_SELF)) {
+	     pam_syslog(pamh, LOG_ERR, "can't combine 'auth_self' with 'trust' or 'deny' args %x", ctrl);
```
ldv-alt, line 90: `%#x` suits best for printing hexadecimals.

modules/pam_wheel/pam_wheel.c
```diff
@@ -199,6 +207,16 @@ perform_check (pam_handle_t *pamh, int ctrl, const char *use_group)
 
 	} else {
 	    retval = PAM_IGNORE;
+
+	    if (ctrl & PAM_AUTH_SELF && pwd->pw_uid == 0) {
+		if (ctrl & PAM_DEBUG_ARG) {
+		    pam_syslog(pamh, LOG_NOTICE, "changing authentication target to %s", fromsu);
+		}
+		if (pam_set_item(pamh, PAM_USER, fromsu) != PAM_SUCCESS) {
```
ldv-alt, line 215: Changing `PAM_USER` is a big hammer, its effects go far beyond the authentication.  In particular, session stack modules would use this new `PAM_USER`, for example, pam_loginuid would set loginuid to the caller uid instead of the target uid, pam_limits would apply limits for the caller uid instead of the target uid, and so on.  In other words, I don't see how to use this new feature safely.

ldv-alt (changes requested):

modules/pam_wheel/pam_wheel.c
lnussel: ok, good point. So that mean support for this kind of authentication would have to go directly into pam_unix

PR #276, opened 2020-10-12
Revert "libpam/pam_modutil_sanitize.c: optimize the way to close fds"

t8m:
This reverts commit 1b087edc7f05237bf5eccc405704cd82b848e761.

Fixes #266

PR #614, opened 2023-10-25
libpam: use close_range() to close file descriptors

ikerexxe:
* configure.ac: check whether close_range() is available in the system.
* libpam/pam_modutil_sanitize.c: use close_range() to close all file descriptors. If the interface isn't available use the previous approach.

Link: https://github.com/linux-pam/linux-pam/pull/276
Resolves: https://issues.redhat.com/browse/RHEL-5099

### DIFF
--- a/libpam/pam_modutil_sanitize.c
+++ b/libpam/pam_modutil_sanitize.c
@@ -11,6 +11,10 @@
 #include <syslog.h>
 #include <sys/resource.h>
 
+#ifndef CLOSE_RANGE_UNSHARE
+#define CLOSE_RANGE_UNSHARE	(1U << 1)
+#endif /* CLOSE_RANGE_UNSHARE */
+
 /*
  * Creates a pipe, closes its write end, redirects fd to its read end.
  * Returns fd on success, -1 otherwise.
@@ -84,9 +88,8 @@ redirect_out(pam_handle_t *pamh, enum pam_modutil_redirect_fd mode,
 	return fd;
 }
 
-/* Closes all descriptors after stderr. */
 static void
-close_fds(void)
+close_fds_iteratively(void)
 {
 	/*
 	 * An arbitrary upper limit for the maximum file descriptor number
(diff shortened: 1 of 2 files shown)


libpam/pam_modutil_sanitize.c
```diff
@@ -107,8 +107,12 @@ close_fds(void)
 	else
 		fd = (int)rlim.rlim_max - 1;
 
+#ifdef HAVE_CLOSE_RANGE
+	close_range(STDERR_FILENO+1, fd, CLOSE_RANGE_UNSHARE);
```
ldv-alt, line 111: `close_range` can fail so its return code has to be checked and a fallback is needed.

When using `close_range`, no need to calculate `fd`, just use `-1U` instead.

To be on the safe side, please check that `CLOSE_RANGE_UNSHARE` is defined.

ldv-alt (changes requested):

libpam/pam_modutil_sanitize.c
ikerexxe: Done

libpam/pam_modutil_sanitize.c
```diff
@@ -111,6 +114,21 @@ close_fds(void)
...
+{
+	int ret;
+
+#ifdef HAVE_CLOSE_RANGE
+	ret = close_range(STDERR_FILENO+1, -1U, CLOSE_RANGE_UNSHARE);
+	if (ret == -1)
+		close_fds_iteratively();
+#else
+	close_fds_iteratively();
+#endif /* HAVE_CLOSE_RANGE */
```
ldv-alt, line 129: Let's simplify it a bit:

```
#ifdef HAVE_CLOSE_RANGE
        if (close_range(STDERR_FILENO+1, -1U, CLOSE_RANGE_UNSHARE) == 0)
                return;
#endif /* HAVE_CLOSE_RANGE */

        close_fds_iteratively();
```

ldv-alt (changes requested):

ldv-alt (approved):

PR #617, opened 2023-11-09
pam_canonicalize_user: new module to canonicalize user name

ldv-alt:
This module uses the name of the user obtained via pam_get_user(3)
as a key to query the password database, and replaces PAM_USER
with the pw_name value that has been returned.

The main usage scenario is systems where a user name is used in several
distinct authentication systems, some of them being case sensitive while
others are not.

The issue this module deals with pops up from time to time.
For example, 9 years ago a similar user name canonicalization was implemented in pam_selinux:
https://bugzilla.redhat.com/show_bug.cgi?id=1071010
https://github.com/linux-pam/linux-pam/commit/da695db7a453b1e2a5ef63fcca21d59a2ed75dda
https://lists.fedorahosted.org/pipermail/pam-developers/2014-March/thread.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -768,7 +768,8 @@ AC_CONFIG_FILES([Makefile libpam/Makefile libpamc/Makefile libpamc/test/Makefile
 	Make.xml.rules \
 	modules/Makefile \
 	modules/pam_access/Makefile \
-        modules/pam_debug/Makefile modules/pam_deny/Makefile \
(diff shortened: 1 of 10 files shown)


ldv-alt:
@ikerexxe, do you think this could be useful?

ikerexxe:
I am not sure. This is the first I've heard of this kind of problem, however I haven't been working on this project that long. From what I see it was a problem several years ago.

PR #622, opened 2023-11-11
pam_rootok: call va_end only once

stoeckmann:
The amount of va_start and va_end calls should be identical. Since va_end is called right after vasprintf, don't call it later again.


ldv-alt (approved):
Fixes: a6a1b9f788a79b2a09827c72a755f471c2e05100

PR #627, opened 2023-11-11
Various fixes related to handling debug output

BenBE:
This PR includes a few fixes related to generating debug output.

The first set of patches includes code changes to address warnings related to the debug output functions themselves, while the later part of this PR fixes several issues related to caller sites for these functions.


libpam/include/security/_pam_types.h
```diff
@@ -166,6 +166,12 @@ typedef struct pam_handle pam_handle_t;
 # define PAM_DEPRECATED
 #endif
 
+#if PAM_GNUC_PREREQ(2,7)
+# define PAM_UNUSED __attribute__((__unused__))
+#else
+# define PAM_UNUSED
+#endif
```
ldv-alt, line 173: Given that m4/attribute.m4 already defines `UNUSED`, does it worth exporting a new macro just for use in PAM_DEBUG case?

Wouldn't it be sufficient to use
```
#ifdef UNUSED
UNUSED
#endif
```
in _pam_macros.h?

BenBE:
Noticed a few more instances of issues; fixes incoming.

BenBE:
Okay, now includes several fixes for warnings that didn't show up for me previously.

ldv-alt:
I suggest commit reordering so that commits that enable warnings are preceded by commits that fix them.

I also suggest adding to CI a job with --enable-debug.

BenBE:
> I suggest commit reordering so that commits that enable warnings are preceded by commits that fix them.

There are no commits in this patch set that enable any additional warnings. The ordering of the commits, while somewhat chaotic, is ordered already to introduce the warning for format issues and subsequently fix the issues detected by this.

> I also suggest adding to CI a job with --enable-debug.

I'd suggest to split this into a separate PR (or suggest for you to add a commit for this as I do not have an overview where I'd need to start to do this properly).

ldv-alt:
> > I suggest commit reordering so that commits that enable warnings are preceded by commits that fix them.
> 
> There are no commits in this patch set that enable any additional warnings. The ordering of the commits, while somewhat chaotic, is ordered already to introduce the warning for format issues and subsequently fix the issues detected by this.

Since new function decorations enable additional warnings, I suggest moving the commits adding these decorations to the end of the commit list.

> > I also suggest adding to CI a job with --enable-debug.
> 
> I'd suggest to split this into a separate PR (or suggest for you to add a commit for this as I do not have an overview where I'd need to start to do this properly).

OK, I added a commit that tests --enable-debug builds, and it shows that there are more warnings that have to be addressed.

ldv-alt (approved):

PR #629, opened 2023-11-11
pam_echo: avoid heap overflow on huge files

stoeckmann:
The module might overflow heap on 32 bit systems if a 4 GB file is supplied as argument.

I have used the size limit which is also in place for pam_motd already.

Proof of Concept (on a 32 bit system):

1. Create a 4 GB file
```
dd if=/dev/zero of=/tmp/huge bs=1 count=1 seek=4294967295
```
2. Create a PAM configuration with pam_echo
```
session                optional        pam_echo.so file=/tmp/huge
```
3. Use the PAM tool, e.g. su (if the line is added to your su configuration)

### DIFF
--- a/modules/pam_echo/pam_echo.c
+++ b/modules/pam_echo/pam_echo.c
@@ -41,6 +41,7 @@
 #include <stdio.h>
 #include <fcntl.h>
 #include <string.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <unistd.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


modules/pam_echo/pam_echo.c
```diff
@@ -182,6 +183,12 @@ pam_echo (pam_handle_t *pamh, int flags, int argc, const char **argv)
 	  return PAM_IGNORE;
 	}
 
+      if ((uint64_t) st.st_size >= (uint64_t) SIZE_MAX)
```
ldv-alt, line 186: Let's not assume that uint64_t is big enough and use uintmax_t instead, or cast SIZE_MAX to __typeof__(st.st_size).

ldv-alt (changes requested):

modules/pam_echo/pam_echo.c
stoeckmann: SIZE_MAX is very large anyway. During our reviews we've encountered a sane limit in pam_motd, so I have copied it now.

ldv-alt: SIZE_MAX is way too large indeed, but returning PAM_BUF_ERR on sizes greater than 0x10000 is a change of behavior.

stoeckmann: Adjusted as recommended.

ldv-alt (approved):

PR #630, opened 2023-11-11
pam_exec: fix stack overflow on \0 output

stoeckmann:
If an executed program prints \0 at the beginning of a line, then pam_exec triggers an out of boundary read (and possible) write on the stack.

Proof of Concept (compile with -fsanitize=address):
```
auth optional pam_exec.so stdout /bin/echo -ne \x00
```


stoeckmann:
Also add the same check to pam_securetty (which requires an invalid configuration).

ldv-alt (approved):

PR #632, opened 2023-11-11
pam_shells: improve /etc/shells handling

stoeckmann:
- Only allow absolute paths for login shells
- Handle truncated lines
- Use getusershell if available (dropped after @ldv-alt pointed out possible MT issues)

Some distributions have comments in /etc/shells. These should never be considered valid login shells.

See also https://github.com/shadow-maint/shadow/pull/837 in shadow.

### DIFF
--- a/modules/pam_shells/pam_shells.c
+++ b/modules/pam_shells/pam_shells.c
@@ -8,6 +8,7 @@
 
 #include "config.h"
 
+#include <limits.h>
 #include <pwd.h>
 #include <stdarg.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


ldv-alt:
Given that these functions are MT-Unsafe, I'm not sure we should introduce them, unless the module is also inherently MT-Unsafe.

stoeckmann:
Oh yes, thanks for pointing it out. I've dropped the getusershell part.

ldv-alt (approved):

PR #633, opened 2023-11-12
libpam: avoid infinite recursion with includes

BenBE:
When there's a loop of configuration files consisting solely of includes the recursion depth level is never incremented and thus no upper limit is enforced. This leads to a crash caused by a stack overflow.

This patch updates the logic to track both the number of includes as well as the number of substacks we are on; ultimately adding a new parameter to track this information.

### DIFF
--- a/libpam/pam_handlers.c
+++ b/libpam/pam_handlers.c
@@ -44,6 +44,7 @@ static int _pam_add_handler(pam_handle_t *pamh
 static int _pam_load_conf_file(pam_handle_t *pamh, const char *config_name
 				, const char *service /* specific file */
 				, int module_type /* specific type */
+				, int include_level /* level of include */
 				, int stack_level /* level of substack */
 #ifdef PAM_READ_BOTH_CONFS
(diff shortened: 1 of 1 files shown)


libpam/pam_handlers.c
```diff
@@ -352,7 +355,7 @@ static int _pam_load_conf_file(pam_handle_t *pamh, const char *config_name
 
     D(("_pam_load_conf_file called"));
 
-    if (stack_level >= PAM_SUBSTACK_MAX_LEVEL) {
+    if (include_level >= PAM_SUBSTACK_MAX_LEVEL) {
 	D(("maximum level of substacks reached"));
 	pam_syslog(pamh, LOG_ERR, "maximum level of substacks reached");
```
ldv-alt, line 360: With this change, the diagnostic message becomes misleading.

ldv-alt (changes requested):

ldv-alt (approved):

PR #636, opened 2023-11-13
libpam: fix _pam_mkargv return value on error path

ldv-alt:
* libpam/pam_misc.c (_pam_mkargv): Return 0 in case of memory allocation failure.


ldv-alt:
@stoeckmann this is related to your #628.

stoeckmann:
Nice spotting!

PR #638, opened 2023-11-15
pam_mkhomedir: switch to *at functions

stoeckmann:
Avoid link dereferences, since they cannot occur with /etc/skel iterations used in the code. If such links are encountered, fail instead.


modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -40,8 +40,10 @@ create_homedir(const struct passwd *pwd, mode_t dir_mode,
    int retval = PAM_SESSION_ERR;
 
    /* Create the new directory */
-   if (mkdir(dest, 0700) && errno != EEXIST)
+   if (mkdir(dest, 0700))
    {
+      if (errno = EEXIST)
```
ldv-alt: `errno = EEXIST` is an assignment, not a comparison.

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -153,15 +155,17 @@ create_homedir(const struct passwd *pwd, mode_t dir_mode,
 	 if (pointedlen >= 0) {
             if(symlink(pointed, newdest) != 0)
             {
-               pam_syslog(NULL, LOG_DEBUG,
-			  "unable to create link %s: %m", newdest);
+               retval = errno = EEXIST ? PAM_SUCCESS : PAM_PERM_DENIED;
```
ldv-alt: `errno = EEXIST` is an assignment, not a comparison.

ldv-alt (changes requested):
Even though all these accidental assignments are fixed in the last commit, please fix them in the first place as we strive to keep the code bisectable.

I'd also like to note that the new approach uses about twice as much file descriptors in the environment where RLIMIT_NOFILE is inherited and could be legitimately set e.g. by pam_limits to a low value. This is not an objection, though.

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: Fixed, thanks for noticing!

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: Fixed as well

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -213,16 +217,17 @@ create_homedir(const struct passwd *pwd, mode_t dir_mode,
       }
 
       /* Open the dest file */
-      if ((destfd = open(newdest, O_WRONLY | O_TRUNC | O_CREAT, 0600)) < 0)
+      if ((destfd = open(newdest, O_WRONLY | O_CREAT | O_EXCL, 0600)) < 0)
       {
-         pam_syslog(NULL, LOG_DEBUG,
-		    "unable to open dest file %s: %m", newdest);
+	 retval = errno = EEXIST ? PAM_SUCCESS : PAM_PERM_DENIED;
```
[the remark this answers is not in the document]
stoeckmann: And also here

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -26,275 +26,346 @@
...
-         continue;
-      }
-
-      /* If it's a symlink, create a new link. */
-      if (S_ISLNK(st.st_mode))
-      {
-	 int pointedlen = 0;
+	    if (retval != PAM_SUCCESS)
+	       pam_syslog(NULL, LOG_DEBUG,
+			  "unable to create link %s: %m", dent->d_name);
```
ldv-alt, line 163: Shouldn't it be
```
"unable to create link %s/%s: %m",
parent->path, dent->d_name);
```
?

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -26,275 +26,346 @@
...
-               if (lchown(newdest, pwd->pw_uid, pwd->pw_gid) != 0)
-               {
-                   pam_syslog(NULL, LOG_DEBUG,
-			      "unable to change perms on link %s: %m", newdest);
-                   closedir(d);
+	 if (fchownat(parent->fd, dent->d_name, pwd->pw_uid, pwd->pw_gid,
+		      AT_SYMLINK_NOFOLLOW) != 0)
+	 {
+	    pam_syslog(NULL, LOG_DEBUG,
+		       "unable to change perms on link %s: %m", dent->d_name);
```
ldv-alt, line 174: Shouldn't it be
```
"unable to change perms on link %s/%s: %m",
parent->path, dent->d_name);
```
?

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -26,275 +26,346 @@
...
+			O_WRONLY | O_CREAT | O_EXCL | O_NOFOLLOW, 0600)) < 0)
+   {
+      retval = errno == EEXIST ? PAM_SUCCESS : PAM_PERM_DENIED;
+      if (retval != PAM_SUCCESS)
          pam_syslog(NULL, LOG_DEBUG,
-		    "unable to open or stat src file %s: %m", newsource);
-         if (srcfd >= 0)
-            close(srcfd);
-         closedir(d);
+		    "unable to open dest file %s: %m", dent->d_name);
```
ldv-alt, line 213: Shouldn't it be
```
"unable to open dest file %s/%s: %m",
parent->path, dent->d_name);
```
?

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -26,275 +26,346 @@
...
-         free(newsource);
-         free(newdest);
+   /* Set the proper ownership and permissions for the module. We make
+      the file a+w and then mask it with the set mask. This preserves
+      execute bits */
+   if (fchmod(destfd, (st.st_mode | 0222) & (~u_mask)) != 0 ||
+       fchown(destfd, pwd->pw_uid, pwd->pw_gid) != 0)
+   {
+      pam_syslog(NULL, LOG_DEBUG,
+		 "unable to change perms on copy %s: %m", dent->d_name);
```
ldv-alt, line 224: Shouldn't ut be
```
"unable to change perms on copy %s: %m",
parent->path, dent->d_name);
```
?

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -26,275 +26,346 @@
...
-   if (chmod(dest, 0777 & (~u_mask)) != 0 ||
-       chown(dest, pwd->pw_uid, pwd->pw_gid) != 0)
+   if (fchmodat(parent->fd, dest, dir_mode, AT_SYMLINK_NOFOLLOW) != 0 ||
+       fchownat(parent->fd, dest, pwd->pw_uid, pwd->pw_gid,
+		AT_SYMLINK_NOFOLLOW) != 0)
    {
       pam_syslog(NULL, LOG_DEBUG,
-		 "unable to change perms on directory %s: %m", dest);
-      return PAM_PERM_DENIED;
+		 "unable to change perms on directory %s: %m", base.path);
```
ldv-alt, line 329: Looks like `base.path` can be NULL here.

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -26,275 +26,346 @@
...
-create_homedir_helper(const struct passwd *_pwd,
+create_homedir_helper(const struct passwd *_pwd, mode_t home_mode,
 		      const char *_skeldir, const char *_homedir)
 {
    int retval = PAM_SESSION_ERR;
+   struct dir_spec base;
+   char *cp = strrchr(_homedir, '/');
 
-   retval = create_homedir(_pwd, _skeldir, _homedir);
+   if (cp == NULL)
```
ldv-alt, line 346: Apparently, `cp` cannot be NULL here because of `pwd->pw_dir[0] != '/'` check added in the calling function.
Maybe moving that check and the code that follows it into this function would result to a more readable code.

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -26,275 +26,346 @@
...
+   {
+      pam_syslog(NULL, LOG_ERR,
+		 "Relative user home directory: %s", _homedir);
+      return PAM_SESSION_ERR;
+   }
 
-   if (chmod(_homedir, home_mode) != 0)
+   *cp = '\0';
+   retval = dir_spec_open(&base,
+			  cp == NULL ? "." : cp == _homedir ? "/" : _homedir);
```
ldv-alt, line 355: `cp` cannot be NULL here as well.

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: Yes, adjusted

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: Adjusted as well

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: And adjusted here as well

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: And also here, thanks for your careful review!

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: Yes, and it's even nicer to use parent->path and destdir so it looks like the other ones.

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: That's true. We've kept the check here just to be cautious with strrchr return value dereference without NULL check. It can be removed and I've done so.

modules/pam_mkhomedir/mkhomedir_helper.c
stoeckmann: Oh yes, thanks

ldv-alt (approved):

PR #644, opened 2023-12-04
Treat INT_MAX limitations in pam_modutil_ioloop.c

stoeckmann:
The size arguments to pam_modutil_read and pam_modutil_write are of type int. I would like to change these to size_t to make them look like size arguments of system calls read and write, but since they are exported through header files, this would imply an API change.

A negative size value should not lead to a 0 return value, but -1 instead. Otherwise this can be interpreted as a short read, which leads to issues in pam_echo.

Since module pam_timestamp did not properly check for negative return values, I have adjusted the code there as well to avoid any side-effects due to my changes.

Proof of Concept for pam_echo:
1. Add pam_echo to a tool of your choice, e.g. /etc/pam.d/su
```
session   optional    pam_echo.so   file=/tmp/poc.txt
```
2. Create a file larger than INT_MAX
```
echo "hello world" > /tmp/poc.txt
dd if=/dev/zero of=/tmp/poc.txt bs=1 count=1 seek=2147483648 conv=notrunc
```
3. Trigger the line, e.g. run su

You will most likely encounter an empty line after the password prompt. If you skip the dd command, you will see "hello world" instead. The empty line would imply that the malloc-returned memory started with '\0' which is most likely the case.

### DIFF
--- a/modules/pam_timestamp/hmac_openssl_wrapper.c
+++ b/modules/pam_timestamp/hmac_openssl_wrapper.c
@@ -84,7 +84,7 @@ static int
 generate_key(pam_handle_t *pamh, char **key, size_t key_size)
 {
     int fd = 0;
-    size_t bytes_read = 0;
+    ssize_t bytes_read = 0;
     char * tmp = NULL;
 
     fd = open("/dev/urandom", O_RDONLY);
@@ -103,7 +103,7 @@ generate_key(pam_handle_t *pamh, char **key, size_t key_size)
     bytes_read = pam_modutil_read(fd, tmp, key_size);
     close(fd);
 
-    if (bytes_read < key_size) {
+    if (bytes_read < 0 || (size_t)bytes_read < key_size) {
         pam_syslog(pamh, LOG_ERR, "Short read on random device");
         free(tmp);
         return PAM_AUTH_ERR;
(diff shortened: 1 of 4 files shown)


ldv-alt (approved):

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -184,7 +190,7 @@ write_file(pam_handle_t *pamh, const char *file_name, char *text,
     bytes_written = pam_modutil_write(fd, text, text_length);
     close(fd);
 
-    if (bytes_written < text_length) {
+    if (bytes_written < 0 || (size_t)bytes_written < text_length) {
         pam_syslog(pamh, LOG_ERR, "Short write on %s", file_name);
         free(text);
```
BenBE, line 195: Inconsistent handling of `text` buffer. All other locations have `pam_overwrite_n(text, text_length);`, thus not having it here looks strange. That aside: Overwriting an input buffer on error looks strange AND is at least "unexpected behaviour". Overwriting buffer contents should be tasked to the caller, as changing a buffer in a write function is at least confusing.

BenBE (commented):
The code changes LGTM, the overall functions where changes are made not so much. Probably best to have these remarks addressed in another PR though …

PR #648, opened 2023-12-06
pam_pwhistory: fix memory leak on error path

ldv-alt:
* modules/pam_pwhistory/pwhistory_config.c (parse_config_file): Free the result returned by pam_modutil_search_key when it becomes unused.


stoeckmann:
I've missed this error case. Thanks for reviewing the other callers!

PR #649, opened 2023-12-06
RFC: libpam: remove unneeded allocations

stoeckmann:
The pam_modutil functions store results of reentrant functions by calling pam_set_data, but these entries are never retrieved again.

If reentrant functions do not exist, no such storing is performed either.

### DIFF
--- a/libpam/pam_modutil_getgrgid.c
+++ b/libpam/pam_modutil_getgrgid.c
@@ -15,28 +15,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 
-static int intlen(int number)
-{
-    int len = 2;
-    while (number != 0) {
-        number /= 10;
-	len++;
-    }
-    return len;
-}
-
-static int longlen(long number)
-{
-    int len = 2;
-    while (number != 0) {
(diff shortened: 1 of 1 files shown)


stoeckmann:
This PR is merely meant as a "request for comment." Are there reasons that pam_modutil functions store their results in pamh? I could imagine that this was meant as "call reentrant functions once, then take result from pamh if called again." But I am not sure if this is still needed.

@ldv-alt, can you clarify this?

If you agree, I would add the other files to this comment as well.

ldv-alt:
> This PR is merely meant as a "request for comment." Are there reasons that pam_modutil functions store their results in pamh? I could imagine that this was meant as "call reentrant functions once, then take result from pamh if called again." But I am not sure if this is still needed.
> 
> @ldv-alt, can you clarify this?

The idea behind all these `pam_set_data` invocations is `pam_modutil_cleanup`,
this ensures that the memory allocated and returned by `pam_modutil_getgrgid` is eventually freed by `pam_end`.

stoeckmann:
Oh, that makes sense. Thanks!

PR #653, opened 2023-12-10
pam_limits: Specify maximum field width in sscanf()

standby24x7:
Add maximum field width in sscanf().
This prevents sscanf from writing past end of the
destination arrays, if the input contains a string longer than the arrays.


ldv-alt:
> Add maximum field width in sscanf(). This prevents sscanf from writing past end of the destination arrays, if the input contains a string longer than the arrays.

How could it happen given that the size of the buffer where the input string is stored is the same as the sizes of the destination string buffers?

stoeckmann:
See https://github.com/linux-pam/linux-pam/pull/662 as a replacement for this PR. It will help an eventual transition from fgets to getline.

PR #655, opened 2023-12-11
pam_timestamp: OpenSSL wrapper updates

BenBE:
This is a follow-up to my notes on #644.

The changes are mainly:
1. Ensure memory buffers with key material should always be wiped. This limits exposure of those buffers. This is mostly defense-in-depth, as those buffers are never exposed to the caller (error paths).

2. Prefer proper APIs for retrieving CSRNG data like OpenSSL's `RAND_priv_bytes` and `getrandom(5)`. As this is a wrapper for OpenSSL the call to `RAND_priv_bytes` is tried first, with `getrandom(5)` tried second. Only if both fail, a fallback to reading `/dev/urandom` is tried (which based on the implementation of `RAND_priv_bytes` is likely to fail too at this point).

I also added some checks for `PAM_NONNULL` to some of the functions to have the compiler check their inputs for valid pointers (as those checks were not performed inside those functions. If desired I can add those for the other functions too.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -640,6 +640,11 @@ AC_CHECK_FUNCS(explicit_bzero memset_explicit)
 AC_CHECK_FUNCS([ruserok_af ruserok], [break])
 AC_CHECK_FUNCS(close_range)
 
+dnl For module/pam_timestamp
+AC_CHECK_HEADERS([sys/random.h])
+dnl May require libbsd/libSystem on non-Linux platforms
+AC_CHECK_FUNCS(getrandom)
+
 AC_ARG_ENABLE([regenerate-docu],
   AS_HELP_STRING([--disable-regenerate-docu],[Don't re-build documentation from XML sources]),
   [enable_docu=$enableval], [enable_docu=yes])
--- a/modules/pam_timestamp/hmac_openssl_wrapper.c
+++ b/modules/pam_timestamp/hmac_openssl_wrapper.c
@@ -105,6 +105,7 @@ generate_key(pam_handle_t *pamh, char **key, size_t key_size)
 
     if (bytes_read < 0 || (size_t)bytes_read < key_size) {
         pam_syslog(pamh, LOG_ERR, "Short read on random device");
+        pam_overwrite_n(tmp, key_size);
         free(tmp);
         return PAM_AUTH_ERR;
     }
@@ -192,6 +193,7 @@ write_file(pam_handle_t *pamh, const char *file_name, char *text,
 
     if (bytes_written < 0 || (size_t)bytes_written < text_length) {
         pam_syslog(pamh, LOG_ERR, "Short write on %s", file_name);
+        pam_overwrite_n(text, text_length);
         free(text);
         return PAM_AUTH_ERR;
     }
--- a/modules/pam_timestamp/hmac_openssl_wrapper.c
+++ b/modules/pam_timestamp/hmac_openssl_wrapper.c
@@ -56,6 +56,10 @@
 #include "hmac_openssl_wrapper.h"
 #include "pam_inline.h"
 
+#ifdef HAVE_SYS_RANDOM_H
+#include <sys/random.h>
+#endif
+
 #define LOGIN_DEFS          "/etc/login.defs"
 #define CRYPTO_KEY          "HMAC_CRYPTO_ALGO"
 #define DEFAULT_ALGORITHM   "SHA512"
@@ -94,6 +98,15 @@ generate_key(pam_handle_t *pamh, char **key, size_t key_size)
         return PAM_AUTH_ERR;
     }
 
+#ifdef HAVE_GETRANDOM
+    /* Fallback to getrandom(2) if available */
+    if (getrandom(tmp, key_size, 0) == (ssize_t)key_size) {
+        *key = tmp;
+        return PAM_SUCCESS;
+    }
+#endif
+
+    /* Fallback to /dev/urandom */
     fd = open("/dev/urandom", O_RDONLY);
     if (fd == -1) {
         pam_syslog(pamh, LOG_ERR, "Cannot open /dev/urandom: %m");
--- a/modules/pam_timestamp/hmac_openssl_wrapper.c
+++ b/modules/pam_timestamp/hmac_openssl_wrapper.c
@@ -87,7 +87,7 @@ generate_key(pam_handle_t *pamh, char **key, size_t key_size)
     ssize_t bytes_read = 0;
     char *tmp = *key = NULL;
 
-    tmp = malloc(key_size);
+    tmp = calloc(1, key_size);
     if (!tmp) {
         pam_syslog(pamh, LOG_CRIT, "Not enough memory");
         return PAM_AUTH_ERR;
@@ -141,7 +141,7 @@ read_file(pam_handle_t *pamh, int fd, char **text, size_t *text_length)
         return PAM_AUTH_ERR;
     }
 
-    tmp = malloc(st.st_size);
+    tmp = calloc(1, st.st_size);
     if (!tmp) {
         pam_syslog(pamh, LOG_CRIT, "Not enough memory");
         close(fd);
--- a/modules/pam_timestamp/hmac_openssl_wrapper.c
+++ b/modules/pam_timestamp/hmac_openssl_wrapper.c
@@ -81,6 +81,7 @@ get_crypto_algorithm(pam_handle_t *pamh, int debug){
 }
 
 static int
+PAM_NONNULL((1, 2))
 generate_key(pam_handle_t *pamh, char **key, size_t key_size)
 {
     int fd = 0;
@@ -117,6 +118,7 @@ generate_key(pam_handle_t *pamh, char **key, size_t key_size)
 }
 
 static int
+PAM_NONNULL((1, 3, 4))
 read_file(pam_handle_t *pamh, int fd, char **text, size_t *text_length)
 {
     struct stat st;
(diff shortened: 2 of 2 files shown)


modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -49,13 +49,18 @@
...
 
 #include <security/pam_ext.h>
 #include <security/pam_modutil.h>
 
 #include "hmac_openssl_wrapper.h"
 #include "pam_inline.h"
 
+#ifdef HAVE_GETRANDOM
+#include <sys/random.h>
+#endif
```
ldv-alt, line 62: I'm not sure there is a guarantee that `<sys/random.h>` is always provided along with `getrandom`.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -81,22 +86,39 @@ get_crypto_algorithm(pam_handle_t *pamh, int debug){
 }
 
 static int
+PAM_NONNULL((1, 2))
```
ldv-alt, line 89: PAM_NONNULL decorations shouldn't be mixed with other changes.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -81,22 +86,39 @@ get_crypto_algorithm(pam_handle_t *pamh, int debug){
...
 {
     int fd = 0;
     ssize_t bytes_read = 0;
-    char * tmp = NULL;
+    char *tmp = *key = NULL;
 
-    fd = open("/dev/urandom", O_RDONLY);
-    if (fd == -1) {
-        pam_syslog(pamh, LOG_ERR, "Cannot open /dev/urandom: %m");
+    tmp = calloc(key_size, 1);
```
ldv-alt, line 96: What was the reason of replacing `malloc` with `calloc`?

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -81,22 +86,39 @@ get_crypto_algorithm(pam_handle_t *pamh, int debug){
...
+    if (!tmp) {
+        pam_syslog(pamh, LOG_CRIT, "Not enough memory");
         return PAM_AUTH_ERR;
     }
 
-    tmp = malloc(key_size);
-    if (!tmp) {
-        pam_syslog(pamh, LOG_CRIT, "Not enough memory");
-        close(fd);
+    //Try to get random data from OpenSSL first
```
ldv-alt, line 102: This project does not use `//` style comments.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -81,22 +86,39 @@ get_crypto_algorithm(pam_handle_t *pamh, int debug){
...
+        *key = tmp;
+        return PAM_SUCCESS;
+    }
+#endif
+
+    // Fallback to /dev/urandom
+    fd = open("/dev/urandom", O_RDONLY);
+    if (fd == -1) {
+        pam_syslog(pamh, LOG_ERR, "Cannot open /dev/urandom: %m");
+        pam_overwrite_n(tmp, key_size);
```
ldv-alt, line 120: This particular zeroing is applied to an area of memory where no data has been written yet.

ldv-alt (changes requested):

modules/pam_timestamp/hmac_openssl_wrapper.c
BenBE: Okay, will split into a separate commit.

modules/pam_timestamp/hmac_openssl_wrapper.c
BenBE: Will take a look at the docs, but AFAICS at least on Linux it's provided there. For *BSD I've to check.

modules/pam_timestamp/hmac_openssl_wrapper.c
BenBE: Having consistent, zero-initialized memory. This ensured a deterministic state when later operations to fill this memory only return the memory only partially filled.

No deal-breaker if skipped, but preferable IMO.

modules/pam_timestamp/hmac_openssl_wrapper.c
BenBE: k, will update.

modules/pam_timestamp/hmac_openssl_wrapper.c
BenBE: This zeroing applies to an area, where previous operations (`RAND_priv_bytes`/`getrandom`) may have partially filled the block.

modules/pam_timestamp/hmac_openssl_wrapper.c
BenBE: Looking around a bit, I found, that all systems that support `getrandom` have it in `sys/random.h` with the exceptions being OpenBSD, where it's in `unistd.h`. For some systems additional libraries have to be linked; I updated `configure.ac` accordingly.

NB: `unistd.h` is checked in `configure.ac` for its presence, but existing code already assumes it to be available. Thus no update on that front; but `sys/random.h` got a check using `HAVE_SYS_RANDOM_H` proper.

configure.ac
```diff
@@ -640,6 +640,12 @@ AC_CHECK_FUNCS(explicit_bzero memset_explicit)
 AC_CHECK_FUNCS([ruserok_af ruserok], [break])
 AC_CHECK_FUNCS(close_range)
 
+dnl For module/pam_timestamp
+AC_CHECK_HEADERS([sys/random.h unistd.h])
```
ldv-alt, line 644: unistd.h is already listed earlier, and it's indeed long assumed to be universally available.

configure.ac
```diff
@@ -640,6 +640,12 @@ AC_CHECK_FUNCS(explicit_bzero memset_explicit)
...
+AC_CHECK_LIB([bsd], [getrandom])
+AC_CHECK_LIB([System], [getrandom])
+AC_CHECK_FUNCS(getrandom)
```
ldv-alt, line 647: If anything, `AC_SEARCH_LIBS` would have to be used instead of `AC_CHECK_LIB`, but it seems to be an overkill in this case.  Let's just use `AC_CHECK_FUNCS(getrandom)`.

configure.ac
BenBE: Removed it. Had it there for completeness. That check was cached anyway though.

configure.ac
BenBE: Updated it to `AC_SEARCH_LIBS` only if `AC_CHECK_FUNCS` succeeds.

That way OpenBSD and OSX should work properly without unnecessary checks for most other platforms.

ldv-alt: It's the other way around, `AC_SEARCH_LIBS` is called first to find a library that provides the function, see e.g. the dance around `crypt_r` in `configure.ac`.

configure.ac
```diff
@@ -640,6 +640,12 @@ AC_CHECK_FUNCS(explicit_bzero memset_explicit)
 AC_CHECK_FUNCS([ruserok_af ruserok], [break])
 AC_CHECK_FUNCS(close_range)
 
+dnl For module/pam_timestamp
+AC_CHECK_HEADERS([sys/random.h])
+AC_SEARCH_LIBS([getrandom], [bsd System], [
+  AC_CHECK_FUNCS(getrandom)
+])
```
ldv-alt, line 647: The reason why I suggested not to invest time into `AC_SEARCH_LIBS` approach is that it's too complicated to worth the effort.

The side effect of this `AC_SEARCH_LIBS` is that everything will be linked with the library that provides `getrandom`.
Since `getrandom` is only needed for pam_timestamp/hmac_openssl_wrapper.c, the linking with extra libraries should be limited to the linking of that file, that is, only when openssl is enabled and only when this file is being linked.  This implies that all these new checks are needed only for `use_openssl=yes` case and the extra libraries would have to be saved into a separate variable and used only in the linking command that needs it.

I still don't see why would you want to go into all this complexity in this project which doesn't seem to be used outside Linux ecosystem nowadays.

modules/pam_timestamp/hmac_openssl_wrapper.c
```diff
@@ -81,22 +86,39 @@ get_crypto_algorithm(pam_handle_t *pamh, int debug){
...
-        pam_syslog(pamh, LOG_CRIT, "Not enough memory");
-        close(fd);
+    /* Try to get random data from OpenSSL first */
+    if (RAND_priv_bytes((unsigned char *)tmp, key_size) == 1) {
+        *key = tmp;
+        return PAM_SUCCESS;
+    }
+
+#ifdef HAVE_GETRANDOM
+    /* Fallback to getrandom(5) if available */
```
ldv-alt, line 109: Apparently, it's `getrandom(2)`.

ldv-alt (changes requested):

modules/pam_timestamp/hmac_openssl_wrapper.c
BenBE: Fixed.

configure.ac
BenBE: Reason is mainly to be friendly to the person who may encounter compile errors on non-Linux platforms. ;-) I left a note re `libbsd`/`libSystem` in a comment; apart from that, it's a plain `AC_CHECK_FUNCS` call now.

ldv-alt (approved):

PR #657, opened 2023-12-11
pam_unix: check str to integer conversions

stoeckmann:
Print an error in syslog if an integer could not be converted.

This prevents working with incorrectly parsed values from e.g. /etc/login.defs.

Proof of Concept:
1. Use following pam_unix.so configuration for passwd in e.g. /etc/pamd./system-password
```
password  required    pam_unix.so       yescrypt shadow try_first_pass minlen=4294967296
```
2. Try to change password with passwd
```
$ passwd
```
3. Notice that you can use any length of password and no error in syslog

With this patch, syslog will contain `pam_unix(passwd:chauthtok): option minlen invalid [4294967296]`.

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -5,6 +5,7 @@
 #include <security/_pam_macros.h>
 #include <security/pam_modules.h>
 #include "support.h"
+#include <limits.h>
 #include <stdio.h>
 #include <string.h>
 #include <sys/types.h>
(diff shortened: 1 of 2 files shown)


stoeckmann:
Conflicts with https://github.com/linux-pam/linux-pam/pull/656 but since it's a different issue, I keep them separated and will rebase this one or the other.

ldv-alt:
> With this patch, syslog will contain `pam_unix(passwd:chauthtok): option minlen invalid [4294967296]`.

Wouldn't it be slightly better to cite the original string in the log, e.g.
`pam_unix(passwd:chauthtok): invalid option value: minlen=4294967296`?

stoeckmann:
> Wouldn't it be slightly better to cite the original string in the log, e.g. `pam_unix(passwd:chauthtok): invalid option value: minlen=4294967296`?

It is the original string value, not the parsed integer. I have tried to keep the style in sync with the other syslog entries, e.g. `option remember not allowed for this module type` and `"unrecognized option [%s]", *argv`

ldv-alt:
> Conflicts with #656 but since it's a different issue, I keep them separated and will rebase this one or the other.

OK, please rebase then.

stoeckmann:
It's rebased.

ldv-alt (approved):

PR #658, opened 2023-12-12
libpam: support arbitrarily long configuration file lines

stoeckmann:
This PR introduces a wrapper around getline, called pam_getline, and uses it within libpam to get rid of fgets calls.

Since this is mostly used when parsing configuration files, it fixes some issues with current code.

See this weird configuration file for example:
```
# this is a comment \
and this is another line
\
 \
   \
 word

line1 \

line2

# prior module.
password  required    pam_unix.so       yescrypt shadow try_first_pass \
                minlen=0
```

- The current code considers "and this is another line" as a new line which is parsed, even though the escaped newline in the comment should indicate otherwise.
- The "word" line has preceding whitespaces because the current code does not handle them well with escaped newlines.
- "line1" and "line2" should be separate lines, yet the current code merges them into "line1 line2"

Having NUL bytes in the file makes it even worse, but since it is a special case and hard to see in a PR report, I have left them out. Please note that I took a lot of care to get them right in the new implementation. Feel free to (and please) test on your own!

I have already exported pam_getline because I would like to use it in modules as well. I have another branch for testing purposes already available. Please note that it's very much work in progress and will be split into multiple PRs (like this one), yet it might indicate what I want to achieve: https://github.com/stoeckmann/linux-pam/tree/getline

### DIFF
--- a/libpam/pam_handlers.c
+++ b/libpam/pam_handlers.c
@@ -32,7 +32,7 @@ static void _pam_free_handlers_aux(struct handler **hp);
 static int _pam_add_handler(pam_handle_t *pamh
 		     , int must_fail, int other, int stack_level, int type
 		     , int *actions, const char *mod_path
-		     , int argc, char **argv, int argvlen);
+		     , int argc, char **argv, size_t argvlen);
 
 /* Values for module type */
 
@@ -77,7 +77,7 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 	int handler_type = PAM_HT_MODULE; /* regular handler from a module */
 	int argc;
 	char **argv;
-	int argvlen;
+	size_t argvlen;
 
 	D(("LINE: %s", buf));
 	if (known_service != NULL) {
@@ -235,7 +235,7 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 	    if (nexttok != NULL) {
 		D(("list: %s",nexttok));
 	        argvlen = _pam_mkargv(nexttok, &argv, &argc);
-		D(("argvlen = %d",argvlen));
+		D(("argvlen = %zu",argvlen));
 		if (argvlen == 0) {
 		    /* memory allocation failed */
 		    D(("failed to allocate argument vector"));
@@ -247,7 +247,8 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 		}
 	    } else {               /* there are no arguments so fix by hand */
 		D(("empty argument list"));
-		argvlen = argc = 0;
+		argvlen = 0;
+		argc = 0;
 		argv = NULL;
 	    }
 
@@ -788,7 +789,7 @@ _pam_load_module(pam_handle_t *pamh, const char *mod_path, int handler_type)
 int _pam_add_handler(pam_handle_t *pamh
 		     , int handler_type, int other, int stack_level, int type
 		     , int *actions, const char *mod_path
-		     , int argc, char **argv, int argvlen)
+		     , int argc, char **argv, size_t argvlen)
 {
     struct loaded_module *mod = NULL;
     struct handler **handler_p;
--- a/libpam/pam_handlers.c
+++ b/libpam/pam_handlers.c
@@ -18,14 +18,23 @@
 #include <fcntl.h>
 #include <unistd.h>
 
-#define BUF_SIZE                  1024
 #define MODULE_CHUNK              4
 #define UNKNOWN_MODULE       "<*unknown module*>"
 #ifndef _PAM_ISA
 #define _PAM_ISA "."
 #endif
 
-static int _pam_assemble_line(FILE *f, char *buf, int buf_len);
+struct line_buffer {
+	char *assembled;
+	char *chunk;
+	size_t chunk_size;
+	size_t len;
+	size_t size;
+};
+
+static void _pam_buffer_init(struct line_buffer *buffer);
+
+static int _pam_assemble_line(FILE *f, struct line_buffer *buf);
 
 static void _pam_free_handlers_aux(struct handler **hp);
 
@@ -62,12 +71,15 @@ static int _pam_parse_conf_file(pam_handle_t *pamh, FILE *f
 #endif /* PAM_READ_BOTH_CONFS */
     )
 {
-    char buf[BUF_SIZE];
+    struct line_buffer buffer;
     int x;                    /* read a line from the FILE *f ? */
+
+    _pam_buffer_init(&buffer);
     /*
      * read a line from the configuration (FILE *) f
      */
-    while ((x = _pam_assemble_line(f, buf, BUF_SIZE)) > 0) {
+    while ((x = _pam_assemble_line(f, &buffer)) > 0) {
+	char *buf = buffer.assembled;
 	char *tok, *nexttok=NULL;
 	const char *this_service;
 	const char *mod_path;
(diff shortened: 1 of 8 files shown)


libpam/pam_misc.c
```diff
@@ -158,14 +159,14 @@ char *_pam_memdup(const char *x, int len)
...
-    int l;
+    size_t l;
     int argvlen = 0;
     char *sbuf, *sbuf_start;
     char **our_argv = NULL;
     char **argvbuf;
     char *argvbufp;
 #ifdef PAM_DEBUG
-    int count=0;
+    unsigned count=0;
```
BenBE, line 169: Why not use `size_t` here, as this is a counter for the number of some items?

BenBE (commented):
LGTM on first glance. One minor note, but that's mostly a question of style. Personally I try to avoid `int`/`long`/`unsigned` for their ambiguity when porting things.

libpam/pam_misc.c
stoeckmann: The `argc` is an `int` and cannot be changed, so the `unsigned` int counterpart will definitely be large enough. Also we have the `%u` formatter which is universally available for it.

ldv-alt:
Apparently, `getline` is being used unconditionally in `pam_namespace` and `pam_sepermit`.
In `pam_namespace` it is being used since 2006 when the module was introduced in the first place.

From my point of view, this would justify code cleanup so that `getline` would be used unconditionally in other cases, too.
What do you think?

ldv-alt:
I've created #659 to use `getline` unconditionally.

stoeckmann:
> From my point of view, this would justify code cleanup so that `getline` would be used unconditionally in other cases, too. What do you think?

In one word: Awesome!

It removes custom code from this PR and I don't have to further think about how to get pam_getline into helper binaries of the modules.

ikerexxe (commented):
At first glance it looks good, but I'll wait until the PR is rebased on top of master to do a full review.

As an enhancement to code maintainability I would create new functions to deal with leading spaces, comments, backslashes, etc. This way you can remove the comments that you included in `_pam_assemble_line()` and replace them with functions that explain properly what you are doing.

stoeckmann:
I have performed a rebase.

Also, to reduce the cyclomatic complexity of _pam_assemble_line, I have split separable aspects into own functions. Separable as in "this processes the string", "this processes the buffer", "this processes the current state".

libpam/pam_handlers.c
```diff
@@ -571,94 +583,207 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
+	/*
+	 * we are only interested in characters before the first '#'
+	 * character
+	 */
+	str[strcspn(str, "#")] = '\0';
+
+	return str;
+}
+
+static inline int _pam_str_unescnl(char *start, char **end)
```
ikerexxe, line 692: What does `unescnl` mean?

libpam/pam_handlers.c
```diff
@@ -571,94 +583,207 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
  */
 
-static int _pam_assemble_line(FILE *f, char *buffer, int buf_len)
+static int _pam_assemble_line(FILE *f, struct line_buffer *buffer)
 {
-    char *p = buffer;
-    char *endp = buffer + buf_len;
-    char *s, *os;
-    int used = 0;
+    int added = 0;
```
ikerexxe, line 735: Since you are returning this value I'd call it `ret` and I think you should also define its return value, either in macros (OK, NOK) or in a comment preceding the function definition. I know you are reusing the previously defined values, but this will help other people understand what you are doing.

libpam/pam_handlers.c
```diff
@@ -571,94 +583,207 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
+	*p++ = ' ';         /* replace backslash with ' ' */
+	*p = '\0';          /* truncate the line here */
+	*end = p;
+	esc = 1;
+    }
+
+    return esc;
+}
+
+static inline int _pam_str_prepare(char *line, ssize_t len,
```
ikerexxe, line 714: Preparing for what?

libpam/pam_handlers.c
```diff
@@ -571,94 +583,210 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
+	 * character
+	 */
+	str[strcspn(str, "#")] = '\0';
+
+	return str;
+}
+
+static inline int _pam_str_unescnl(char *start, char **end)
+{
+    int esc = 0;
```
ikerexxe, line 694: `ret` instead of `esc`? And again define the meaning of its return value?

libpam/pam_handlers.c
```diff
@@ -571,94 +583,210 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
+	esc = 1;
+    }
+
+    return esc;
+}
+
+static inline int _pam_str_prepare(char *line, ssize_t len,
+				   char **start, char **end)
+{
+    int esc;
```
ikerexxe, line 717: Same here

libpam/pam_handlers.c
```diff
@@ -571,94 +583,210 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
-	if (p >= endp - 1) {
-	    /* Overflow */
-	    D(("overflow"));
-	    return -1;
-	}
-	if (fgets(p, endp - p, f) == NULL) {
-	    if (used) {
+	char *start, *end;
+	ssize_t n;
+	int strcont;
```
ikerexxe, line 746: I don't understand what this variable does.

ikerexxe (commented):
I think you could squash the last two commits in one, but I'm not a maintainer of this project and rules may change between them.

stoeckmann:
> I think you could squash the last two commits in one, but I'm not a maintainer of this project and rules may change between them.

If the adjustments are acceptable, I will squash them.

Having two commits should help reviewers who are comfortable with old _pam_assemble_line logic (commit 2) and how we've progressed in making it more readable (commit 3).

libpam/pam_handlers.c
stoeckmann: Unescape new line. The '\\\n' at the end is replaced with ' ' for addition. I have added a comment to clarify it.

libpam/pam_handlers.c
stoeckmann: I've named it `added` so `1` shows that we are at least in the process of parsing a configuration line which spans across multiple file lines (if getline fails while parsing such a line, it implies that the file ended prematurely, which is an error).

Renamed it to `ret` now.

libpam/pam_handlers.c
stoeckmann: Added a comment. Preparing for "configuration line processing." If you think it should be _pam_line_prepare, implying that we differentiate further between C strings, file lines (technically/semantically identical to C strings), configuration lines (can span across multiple file lines), and line parsing state, we can rename it.

libpam/pam_handlers.c
stoeckmann: Added a comment and renamed.

libpam/pam_handlers.c
stoeckmann: It indicates if we have to continue adding file lines to the current configuration line stored in buffer (1) or if we have actually encountered the EOL (\n) which also completes the configuration line.

libpam/pam_handlers.c
ikerexxe: No, this is fine

libpam/pam_handlers.c
ikerexxe: Perhaps `multiline` would be a better name?

stoeckmann: I've changed the 0/1 logic so we have a boolean which indicates if the current line has an "end of line" marker (i.e. \n) in it or not. This way, the new variable name `eol` indicates just that. If it is true, we have reached the end. Otherwise we have to continue.

I gave your `multiline` approach a thought and noticed that if we reach the end of a multiline, it would be 0 (which feels a tiny bit off, given that we definitely had a multiline in the other iterations). This made me come up with `eol` instead. What do you think?

stoeckmann:
I had to adjust _pam_mkargv to also return size_t instead int, because otherwise a signed integer overflow could occur as well. This was proven during a proof of concept of pam_filter being unable to process arbitrarily long configuration file lines.

Discovered and fixed with @BenBE and @Maaxxs.

Proof of Concept:
1. Create a very long configuration file line for pam_filter in e.g. /etc/pam.d/su
```
echo -n "auth required pam_filter.so run1 filter " > /etc/pam.d/su
dd if=/dev/zero bs=1024 count=4194305 | tr '\0' 'A' >> /etc/pam.d/su
```
2. Run su
```
$ su
SUMMARY: AddressSanitizer: heap-buffer-overflow ../../../../libsanitizer/asan/asan_interceptors.cpp:440 in __interceptor_strcpy
```

stoeckmann:
Fixed pam_exec and pam_listfile for very long lines as well (argc, argv handling).

libpam/pam_handlers.c
```diff
@@ -571,94 +584,242 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
+{
+	/* skip leading spaces */
+	str += strspn(str, " \t");
+	/*
+	 * we are only interested in characters before the first '#'
+	 * character
+	 */
+	str[strcspn(str, "#")] = '\0';
+
+	return str;
```
ikerexxe, line 698: I just noticed that the indentation for this function is different

libpam/pam_handlers.c
ikerexxe: My first idea was to call it `eol`m but that would have reversed your logic and I didn't want to do that. So, now I'm very happy :smile:

libpam/pam_handlers.c
stoeckmann: Oh yes, thanks for pointing out! Adjusted now.

stoeckmann:
Further adjusted _pam_mkargv to protect against integer overflows on 32 bit systems and very long (but not too long) lines.

ikerexxe (approved):
The code looks good to me, but I haven't tested it. I know @DDDas7 has a set up to test https://github.com/linux-pam/linux-pam/issues/532 and I'd like to hear from him before merging this PR.

stoeckmann:
The split function used in pam_motd contained an unsigned int counter which could overflow with arbitrarily long configuration file lines. Fixed.

stoeckmann:
Escaped newlines in comments should be ignored. It's current behavior, is compatible with OpenPAM and also the way bash would handle lines like
```
echo a # b \
c
```
The `c` would be the next line, i.e. the command `c` would be executed.

libpam/pam_handlers.c
```diff
@@ -572,93 +584,166 @@ int _pam_init_handlers(pam_handle_t *pamh)
...
+		    used = -1;
+		    break;
+		}
+		used = 1;
+	    }                       /* there is more ... */
+	} else {
+	    /* End of the line! */
+	    if (buffer->assembled != NULL || (*os != '\0' && *os != '\n')) {
+		if (_pam_buffer_add(buffer, os, buffer->chunk + n))
+		    used = -1;
```
ldv-alt: This `used = -1` has to be followed by `break`, otherwise it's set to 0.

libpam/pam_handlers.c
```diff
@@ -572,93 +584,166 @@ int _pam_init_handlers(pam_handle_t *pamh)
     return PAM_SUCCESS;
 }
 
+static int _pam_buffer_add(struct line_buffer *buffer, char *start, char *end)
+{
+    size_t len = end - start;
+
+    D(("assembled: [%zu/%zu] '%s', adding [%zu] '%s'",
+	buffer->len, buffer->size,
+	buffer->assembled == NULL ? "" : buffer->assembled, len, start));
```
ldv-alt, line 593: This assumes that `strlen(start) == len`.

libpam/pam_handlers.c
```diff
@@ -658,14 +672,101 @@ static void _pam_buffer_shift(struct line_buffer *buffer)
...
+static inline int _pam_str_prepare(char *line, ssize_t len,
+				   char **start, char **end)
+{
+    int ret;
+
+    *start = line;
+    *end = line + len;
+
+    ret = _pam_str_unescnl(*start, end) ||
+	(*start)[strcspn(*start, "#")] == '#';
```
ldv-alt, line 753: The comment says the function returns 1 if newline is encountered,
but the function also returns 1 if `#` is encountered.

By the way, isn't `strchr(ptr, '#') != NULL` easier to read than `ptr[strcspn(ptr, '#')] == '#'`?

libpam/pam_handlers.c
stoeckmann: The `len` output in `[%zu]` does not assume that `strlen(start) == len`. Instead, if `len` is larger than `strlen(start)`, then a NUL byte has been encountered in the input line. We want to add it to our parsed line as well so the configuration line interpreter will end and this NUL byte without stopping the parser to search for escaped newlines afterwards.

libpam/pam_handlers.c
stoeckmann: Hm, this seems to be a comment for an outdated version. Is this still an issue?

libpam/pam_handlers.c
stoeckmann: Oh yes, thanks. Adjusted

libpam/pam_handlers.c
ldv-alt: The comment is for the penultimate commit in the series.  Apparently, the last commit in the series fixed that, but I'd prefer this to be fixed in the first commit that introduces it.

stoeckmann: Ah okay, thanks for clarification. Fixed. Please note that we can (and probably) should squash these two commits.

ldv-alt: Feel free to squash them if you feel they should rather be squashed.

stoeckmann: I have squashed them now.

ldv-alt:
Ideally, we need some tests for this.
Not necessarily new tests, it's OK to extend already existing tests.

stoeckmann:
> Ideally, we need some tests for this. Not necessarily new tests, it's OK to extend already existing tests.

I have added a test case

ldv-alt (approved):

ldv-alt:
I've added another test that hopefully demonstrates that the new parser really supports long lines.

stoeckmann:
> I've added another test that hopefully demonstrates that the new parser really supports long lines.

Awesome, thanks for finding a module with mandatory options and writing the test!

You could drop my commit which adds a test to pam_permit if you want to. Yours is definitely more helpful and covers everything mine did, adding more test coverage.

ldv-alt:
Some redundancy in the test coverage is IMHO a bonus, so I'd rather keep your test as well.

ldv-alt:
Given the importance and complexity of this change, I've been waiting longer than usual to let anyone contribute.
Now I think it's time to conclude the review and merge this PR.

stoeckmann:
Agreed, especially keeping this PR open for some time. Given the feedback, tests and adjustments, it was well justified. Hope that no regression is introduced anywhere...

Thanks to everyone involved!

ldv-alt:
Resolves: https://github.com/linux-pam/linux-pam/pull/561

PR #659, opened 2023-12-12
Assume getline exists

ldv-alt:
Apparently, getline is being used unconditionally in pam_namespace and pam_sepermit.  In pam_namespace, it is being used since 2006 when the module was introduced in the first place.

Let's assume getline is universally available and let's use it unconditionally in other cases, too.

* configure.ac (AC_CHECK_FUNCS): Remove getline and getdelim.
* libpam/pam_modutil_searchkey.c (pam_modutil_search_key): Use getline unconditionally.
* modules/pam_pwhistory/opasswd.c (check_old_pass, save_old_pass): Likewise.
* modules/pam_shells/pam_shells.c (perform_check): Likewise.


stoeckmann:
Fully agreed. The `getline` function is part of POSIX.1-2008. The shadow project requires C11 and POSIX.1-2008 now, too (see https://github.com/shadow-maint/shadow/issues/600 ). Would it make sense to consider these as a given for Linux-PAM as well?

PR #660, opened 2023-12-12
treewide: require existence of getline

stoeckmann:
The getline function has been used unconditionally for many years already, so make it a requirement for easier code base.


stoeckmann:
Duplicate of https://github.com/linux-pam/linux-pam/pull/659

PR #667, opened 2023-12-14
treewide: assume free(NULL) is no-op

ldv-alt:
The C standard guarantees that if the argument of free() is a null pointer, no action occurs.

### DIFF
--- a/libpam/pam_modutil_cleanup.c
+++ b/libpam/pam_modutil_cleanup.c
@@ -12,8 +12,6 @@ void
 pam_modutil_cleanup (pam_handle_t *pamh UNUSED, void *data,
 		     int error_status UNUSED)
 {
-    if (data) {
 	/* junk it */
-	(void) free(data);
-    }
+	free(data);
 }
--- a/modules/pam_listfile/pam_listfile.c
+++ b/modules/pam_listfile/pam_listfile.c
@@ -92,7 +92,7 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
 	    else if(!strcmp(myval,"fail"))
 		onerr = PAM_SERVICE_ERR;
 	    else {
-	        if (ifname) free (ifname);
+	        free(ifname);
 		return PAM_SERVICE_ERR;
 	    }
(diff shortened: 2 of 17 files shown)


libpam/pam_modutil_cleanup.c
```diff
@@ -12,8 +12,6 @@ void
 pam_modutil_cleanup (pam_handle_t *pamh UNUSED, void *data,
 		     int error_status UNUSED)
 {
-    if (data) {
 	/* junk it */
 	(void) free(data);
```
BenBE, line 16: Indentation, useless void cast …

modules/pam_listfile/pam_listfile.c
```diff
@@ -101,11 +101,11 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
...
 		sense=1;
 	    else {
-	        if (ifname) free (ifname);
+	        free(ifname);
 		return onerr;
 	    }
 	else if(!strcmp(mybuf,"file")) {
-	    if (ifname) free (ifname);
+	    free(ifname);
 	    ifname = malloc(strlen(myval)+1);
```
BenBE, line 109: Potentially use `strdup` …

BenBE (commented):
LGTM.

stoeckmann:
Another instance can be found here: https://github.com/linux-pam/linux-pam/blob/f032df68c985a80a2d994962b498ad4391dd30cc/modules/pam_xauth/pam_xauth.c#L553

ldv-alt:
> Another instance can be found here:
> 
> https://github.com/linux-pam/linux-pam/blob/f032df68c985a80a2d994962b498ad4391dd30cc/modules/pam_xauth/pam_xauth.c#L553

Yes, there are several cases of conditional free() performed along with zeroing.  In the beginning I was going to leave them alone, but now I'm inclined to include them into this change as well.

modules/pam_listfile/pam_listfile.c
stoeckmann: See https://github.com/linux-pam/linux-pam/pull/671

PR #668, opened 2023-12-14
treewide: store strlen results in size_t

stoeckmann:
Very long strings could overflow the int data type. Make sure to use the correct data type.

Spotted and fixed with @BenBE and @Maaxxs based on discovery by @ellcs.

This is also a preparation for transition from fgets to getline in pam_access.

### DIFF
--- a/modules/pam_access/pam_access.c
+++ b/modules/pam_access/pam_access.c
@@ -427,8 +427,8 @@ login_access (pam_handle_t *pamh, struct login_info *item)
 #ifdef HAVE_LIBAUDIT
     int     nonall_match = NO;
 #endif
-    int     end;
-    int     lineno = 0;		/* for diagnostics */
+    size_t  end;
+    size_t  lineno = 0;		/* for diagnostics */
     char   *sptr;
 
     if (item->debug)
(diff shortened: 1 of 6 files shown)


modules/pam_access/pam_access.c
```diff
@@ -448,9 +448,11 @@ login_access (pam_handle_t *pamh, struct login_info *item)
     if ((fp = fopen(item->config_file, "r"))!=NULL) {
 	while (!match && fgets(line, sizeof(line), fp)) {
 	    lineno++;
+	    if (line[0] == 0)
+		continue;
```
ldv-alt, line 452: This fix goes beyond the use of size_t.

modules/pam_namespace/pam_namespace.c
```diff
@@ -870,6 +871,12 @@ static char *md5hash(const char *instname, struct instance_data *idata)
      * Create MD5 hashes for instance pathname.
      */
 
+    len = strlen(instname);
+    if (len > UINT_MAX) {
+        pam_syslog(idata->pamh, LOG_CRIT, "Instance pathname too long");
+        return NULL;
+    }
```
ldv-alt, line 878: Likewise.

ldv-alt (changes requested):
Let's not mix several different fixes in a single commit.

modules/pam_access/pam_access.c
stoeckmann: Moved into its own PR, since this one is merely a clean up of int/size_t now: https://github.com/linux-pam/linux-pam/pull/676

modules/pam_namespace/pam_namespace.c
stoeckmann: In fact, this check is redundant. Even though md5hash itself does not check, its only caller makes sure that instname is not that long. Dropped.

ldv-alt (approved):

PR #672, opened 2023-12-15
libpam: reduce memory usage of _pam_mkargv

stoeckmann:
The input string "s" is duplicated into "sbuf" and tokens copied from there into target memory "our_argv".

Since "our_argv" is allocated to be always large enough to keep the whole string "s" (plus pointers) in it, we can skip "sbuf" entirely.


stoeckmann:
> This seems to break the unit tests for `pam_debug`:

Yes, the whitespace removal didn't work as before. Fixed.

ldv-alt (approved):

PR #673, opened 2023-12-15
libpam: remove function prefixes in debug messages

stoeckmann:
The D macro itself already adds the function names.

It is a follow up to 79f97b5dfddbd54942036851e49c369502689853.

### DIFF
--- a/libpam/pam_auth.c
+++ b/libpam/pam_auth.c
@@ -15,7 +15,7 @@ int pam_authenticate(pam_handle_t *pamh, int flags)
 {
     int retval;
 
-    D(("pam_authenticate called"));
+    D(("called."));
 
     IF_NO_PAMH(pamh, PAM_SYSTEM_ERR);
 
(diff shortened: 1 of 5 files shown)


libpam/pam_auth.c
```diff
@@ -36,7 +36,7 @@ int pam_authenticate(pam_handle_t *pamh, int flags)
     if (retval != PAM_INCOMPLETE) {
 	_pam_sanitize(pamh);
 	_pam_await_timer(pamh, retval);   /* if unsuccessful then wait now */
-	D(("exit"));
+	D(("exiting"));
```
BenBE, line 39: If adding a period for `called.` consider adding it for `exited.` too; or remove for both.

Looking at things, probably dropping the period on entry looks best IMO.

stoeckmann: Looking at majority of files, "called." and "exiting" are preferred by now.

ldv-alt (approved):

PR #675, opened 2023-12-15
pam_exec: allow also expose_authtok for password PAM_TYPE

rugubara:
I'm developing an integration between linux PAM and OpenZFS native encryption for datasets to automatically encrypt users' home directories. 
The pam_exec module fitted almost 100% to unlock the user's home directory dataset provided the password is used also as the key source for the dataset.
Apparently, when the password changes, the key must also be updated to keep them in sync.

However pam_exec exposes the password thru expose_authtok for auth PAM_TYPE only. This trivial patch allows pam_exec to expose the new password when the user or admin changes the password. 

please consider accepting this patch into the master branch.

Best regards
Anton Gubarkov.

### DIFF
--- a/modules/pam_exec/pam_exec.8.xml
+++ b/modules/pam_exec/pam_exec.8.xml
@@ -103,7 +103,7 @@
           </term>
           <listitem>
             <para>
-              During authentication the calling command can read
+              During authentication and password change the calling command can read
               the password from <citerefentry>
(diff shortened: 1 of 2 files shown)


ldv-alt (changes requested):
Please update `expose_authtok` description in `modules/pam_exec/pam_exec.8.xml`.

ldv-alt (approved):

Issue #677, opened 2023-12-20
pam_timestamp_check: potential null pointer dereference on error path


Fmiard:
Hi:
In the code on line 816 of the pam_timestamp.c file, if the pointer variable pwd is null, then a null pointer dereference problem occurs on lines 821 and 825.
To avoid this problem, I think we should return the value of the variable retval directly after determining that pwd is null, please review this modification to see if it makes sense, and if so I will create a commit. Thank you very much! 
![9 M)}Y0MJIL@RR@K2G96)0A](https://github.com/linux-pam/linux-pam/assets/154431208/2b2517c0-e232-4add-a16a-dfa17be04722)

stoeckmann:
Agreed. You can reproduce this by renaming `/etc/passwd` temporarily or by removing the entry of the calling user before calling `pam_timestamp_check`.

Fmiard:
@stoeckmann Thank you very much for your guidance!

Fmiard:
I have created a pull request on this issue, please review it to see if it is correct.Thanks!
https://github.com/linux-pam/linux-pam/pull/679

PR #679, opened 2023-12-21
pam_timestamp_check: fix potential null pointer dereference on error path

Fmiard:
If the pointer variable pwd is null, then a null pointer dereference problem occurs when fetching the contents of pw_uid and pw_name from the pwd variable.
To avoid this problem, I think we should return the value of the variable retval directly after determining that pwd is null.
https://github.com/linux-pam/linux-pam/issues/677

### DIFF
--- a/modules/pam_timestamp/pam_timestamp.c
+++ b/modules/pam_timestamp/pam_timestamp.c
@@ -815,7 +815,8 @@ main(int argc, char **argv)
 	/* Get the name of the invoking (requesting) user. */
 	pwd = getpwuid(getuid());
 	if (pwd == NULL) {
-		retval = 4;
+		fprintf(stderr, "unknown user\n");
+		return 4;
 	}
 #ifdef USE_LOGIND
 	uid = pwd->pw_uid;
--- a/modules/pam_timestamp/pam_timestamp.c
+++ b/modules/pam_timestamp/pam_timestamp.c
@@ -752,7 +752,7 @@ pam_sm_close_session(pam_handle_t *pamh UNUSED, int flags UNUSED, int argc UNUSE
 int
 main(int argc, char **argv)
 {
-	int i, retval = 0, dflag = 0, kflag = 0;
+	int i, retval, dflag = 0, kflag = 0;
 	const char *target_user = NULL, *user = NULL, *tty = NULL;
 	struct passwd *pwd;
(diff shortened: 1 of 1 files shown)


modules/pam_timestamp/pam_timestamp.c
```diff
@@ -816,6 +816,7 @@ main(int argc, char **argv)
 	pwd = getpwuid(getuid());
 	if (pwd == NULL) {
 		retval = 4;
+		return retval;
```
stoeckmann, line 819: If the user contains invalid characters, line 833 would print an error message.

Looking at the whole function, my guess (since I am not a maintainer) is, that it would make sense to move the whole pwd-processing into its own else-block right after line 819.

After the else, retval should be checked for value 4 and if it's true, then print the error and call return as suggested. This would lead to a central code point which covers the other `retval = 4` case in 827.

stoeckmann (commented):
Also, although not exactly relevant to this PR (but it would fit), is line 800. It sets retval to 2, prints an error, but still continues processing.

You can see this if you remove the setuid bit from the binary. I would expect a return value there as well.

So, if you want to add it to this PR (I would recommend it, but again, I'm not a maintainer), it would make sense to add it as a separate commit.

ldv-alt (approved):
It would be great if this PR could be reviewed by somebody else.

modules/pam_timestamp/pam_timestamp.c
```diff
@@ -815,7 +815,7 @@ main(int argc, char **argv)
 	/* Get the name of the invoking (requesting) user. */
 	pwd = getpwuid(getuid());
 	if (pwd == NULL) {
-		retval = 4;
+		return 4;
```
stoeckmann, line 818: How about adding "unknown user" error message here?

stoeckmann (commented):
Looks good. I think that the newly introduced check with return value 4 in case of error is okay. These additions (check result of format_timestamp_name and turning more retval-assignments into return-statements) could be separated in their own respective commits or added in this commit message, but otherwise it looks functional to me.

modules/pam_timestamp/pam_timestamp.c
ldv-alt: Added, thanks.

PR #684, opened 2024-01-01
pam_env: various cleanups

stoeckmann:
A memory leak can occur if DEFAULT or OVERRIDE are supplied multiple times in a line.

While at it, prepare a change for later pam_assemble_line port from libpam to pam_env by removing the usage of a fixed sized stack variable. It doesn't take copying actually.

Also improve debug and error messages and reduce variable visibility so it is easier to trace where variables are actually used (which makes the debug and error message adjustment also easier to verify).

### DIFF
--- a/modules/pam_env/pam_env.c
+++ b/modules/pam_env/pam_env.c
@@ -696,8 +696,8 @@ _expand_arg(pam_handle_t *pamh, char **value)
     free(*value);
     if ((*value = malloc(idx + 1)) == NULL) {
       D(("Couldn't malloc %zu bytes for expanded var", idx + 1));
-      pam_syslog (pamh, LOG_CRIT, "Couldn't malloc %lu bytes for expanded var",
-	       (unsigned long)idx+1);
+      pam_syslog (pamh, LOG_CRIT, "Couldn't malloc %zu bytes for expanded var",
+	       idx+1);
       goto buf_err;
     }
   }
--- a/modules/pam_env/pam_env.c
+++ b/modules/pam_env/pam_env.c
@@ -581,9 +581,9 @@ _expand_arg(pam_handle_t *pamh, char **value)
 		    * a constant so that the compiler will shut up when I
 		    * call pam_getenv and _pam_get_item_byname -- sigh
 		    */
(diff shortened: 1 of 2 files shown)


modules/pam_env/pam_env.c
```diff
@@ -581,9 +581,9 @@ _expand_arg(pam_handle_t *pamh, char **value)
 		    * a constant so that the compiler will shut up when I
 		    * call pam_getenv and _pam_get_item_byname -- sigh
 		    */
+  const char *tmpval;
```
ldv-alt: While this change is correct, I was a bit confused by the commit message.
If I understand correctly, getting rid of the copy is OK because during the short window when the copy is used the original string is _not_ modified, and the fact that the original string is modified afterwards doesn't look relevant for this change.

ldv-alt (approved):
Besides that commit message, the changes look good to merge.

BenBE:
The `and also later cleanups.` part in 97d04463656bd86a85c8d09a6535e6b6eb353578 doesn't compute grammatically. Does it remove later cleanups (or the need for those? Or the need for another `strncpy` in later cleanups?

modules/pam_env/pam_env.c
stoeckmann: I have simplified the commit message. I have tried to point out that the const keyword is kind of violated on purpose (which the comment in the code remotely hints at), but since it is really not relevant to the change, it is dropped.

stoeckmann:
> The `and also later cleanups.` part in [97d0446](https://github.com/linux-pam/linux-pam/commit/97d04463656bd86a85c8d09a6535e6b6eb353578) doesn't compute grammatically. Does it remove later cleanups (or the need for those? Or the need for another `strncpy` in later cleanups?

Reworded. With strncpy gone, it will be easier to switch from fgets to getline in a future commit.

stoeckmann:
I have added a commit to use `strndup` for easier review. Since it is already in use in other modules, it is not a newly introduced function call to the code base.

Also adding a commit here avoids merge conflicts due to the newly introduced free-call to prevent the memory leak.

PR #685, opened 2024-01-01
modules: prevent config truncations

stoeckmann:
The modules pam_faillock, pam_listfile and pam_wheel could silently truncate very long arguments coming from pam configuration files.

This was true for pam_listfile even before the arbitrary line length adjustment.

### DIFF
--- a/modules/pam_listfile/pam_listfile.c
+++ b/modules/pam_listfile/pam_listfile.c
@@ -181,7 +181,7 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
 	}
     }
 
-    /* Short-circuit - test if this session apply for this user */
+    /* Short-circuit - test if this session applies for this user */
     {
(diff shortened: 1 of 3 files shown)


ldv-alt (approved):
pam_listfile is indeed a messy code, thanks for cleaning it up a bit.

modules/pam_listfile/pam_listfile.c
```diff
@@ -181,7 +181,7 @@ pam_sm_authenticate (pam_handle_t *pamh, int flags UNUSED,
 	}
     }
 
-    /* Short-circuit - test if this session apply for this user */
+    /* Short-circuit - test if this session applies for this user */
```
ldv-alt, line 164: Ideally, the commit message should mention that the typo being fixed is in a comment.

stoeckmann: Adjusted

PR #687, opened 2024-01-02
pam_namespace: bugfixes, security improvements, documentation improvements

mgerstner:
I recently performed a security code audit of `pam_namespace` and noticed a couple of things that I want to address with this PR#.

### DIFF
--- a/modules/pam_namespace/pam_namespace.c
+++ b/modules/pam_namespace/pam_namespace.c
@@ -529,7 +529,7 @@ static int process_line(char *line, const char *home, const char *rhome,
     instance_prefix = config_options[1];
     if (instance_prefix == NULL) {
         pam_syslog(idata->pamh, LOG_NOTICE, "Invalid line missing instance_prefix");
-        instance_prefix = NULL;
+        dir = NULL;
         goto skipping;
     }
     method = config_options[2];
--- a/modules/pam_namespace/pam_namespace.c
+++ b/modules/pam_namespace/pam_namespace.c
@@ -1833,7 +1833,7 @@ static int cleanup_tmpdirs(struct instance_data *idata)
 		}
 	    } else if (pid < 0) {
 		pam_syslog(idata->pamh, LOG_ERR,
-			"Cannot fork to run namespace init script, %m");
+			"Cannot fork to cleanup temporary directory, %m");
 		rc = PAM_SESSION_ERR;
 		goto out;
 	    }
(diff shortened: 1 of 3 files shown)


modules/pam_namespace/pam_namespace.c
```diff
@@ -1379,6 +1381,9 @@ static int inst_init(const struct polydir_s *polyptr, const char *ipath,
 					/* ignore failures, they don't matter */
 				}
 
+				if (close_range(3, ~0U, 0) < 0)
+					_exit(1);
```
ldv-alt, line 1385: I doubt we can assume that `close_range` is universally available.
Let's use `pam_modutil_sanitize_helper_fds` instead, e.g.
```
if (pam_modutil_sanitize_helper_fds(idata->pamh, PAM_MODUTIL_IGNORE_FD,
                                    PAM_MODUTIL_IGNORE_FD,
                                    PAM_MODUTIL_IGNORE_FD) < 0)
        _exit(1);
```
```

File: modules/pam_namespace/pam_namespace.c [hunk 2]
```diff
@@ -1817,6 +1822,8 @@ static int cleanup_tmpdirs(struct instance_data *idata)
 			_exit(1);
 		}
 #endif
+		if (close_range(3, ~0U, 0) < 0)
+			_exit(1);
```
ldv-alt, line 1826: Likewise.

modules/pam_namespace/pam_namespace.c
```diff
@@ -34,6 +34,8 @@
 
 #define _ATFILE_SOURCE
 
+#include <linux/close_range.h>
+
```
ldv-alt, line 38: I doubt we can assume that this header file is universally available,
but we might not need it after all, see below.

ldv-alt (changes requested):

stoeckmann:
Changes look good. While trying to understand how `dir` is handled, I encountered another issue with invalid (and huge) configuration files. Since pam_namespace already used getline, this is not a new issue introduced with the support for arbitrarily long PAM configuration files.

@mgerstner, could you kindly look at https://github.com/linux-pam/linux-pam/pull/691?

modules/pam_namespace/pam_namespace.c
mgerstner: Do you have any guidelines regarding backwards compatibility? `pam_namespace` is Linux specific so other operating systems should not matter. `close_range()` has been available since kernel 5.9 which is about 3 years old now.

Ah I see that `pam_modutil_sanitize_helper_fds()` tries to use `close_range()` internally, if autoconf finds it. All right, I'll switch to the helper function then.

mgerstner: I pushed a rebased version using `pam_modutil_sanitize_helper_fds()`

ldv-alt (approved):

PR #691, opened 2024-01-02
pam_namespace: validate amount of uids in config

stoeckmann:
If more than INT_MAX uids are found in a configuration line, the variable `count` would trigger a signed integer overflow.

If more than UINT_MAX uids are found in a configuration line, then the `num_uids` counter is invalid, which could eventually lead to out of boundary accesses.

Also make sure that size multiplication for malloc does not overflow.

### DIFF
--- a/modules/pam_namespace/pam_namespace.c
+++ b/modules/pam_namespace/pam_namespace.c
@@ -637,7 +637,7 @@ static int process_line(char *line, const char *home, const char *rhome,
     if (uids) {
         uid_t *uidptr;
         const char *ustr, *sstr;
-        int count, i;
+        size_t count, i;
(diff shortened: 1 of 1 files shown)


ldv-alt (approved):

mgerstner:
The change looks good to me.

Luckily, since the `namespace.conf` file is only controlled by root and that many UIDs are unlikely to ever be specified, such a situation should never occur. Except if something goes very wrong on the configuration side.

stoeckmann:
Yes, fully agree. This change is not security-relevant in any way. But it's always nice to have someone look at such changes who has analyzed the module in question thoroughly. Thanks!

PR #692, opened 2024-01-02
configure.ac: remove unneeded function checks

stoeckmann:
Even if these checks reveal that a function does not exist there is no corresponding HAVE_* check in source files.

Noticed while adding `strndup` to https://github.com/linux-pam/linux-pam/pull/684.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -490,8 +490,8 @@ AS_IF([test "x$enable_nis" != "xno"], [
   CPPFLAGS="$CPPFLAGS $NSL_CFLAGS"
   LIBS="$LIBS $NSL_LIBS"
 
-  AC_CHECK_FUNCS([yp_get_default_domain yperr_string yp_master yp_bind yp_match yp_unbind])
-  AC_CHECK_FUNCS([getrpcport rpcb_getaddr])
+  AC_CHECK_FUNCS([yp_get_default_domain yp_bind yp_match yp_unbind])
+  AC_CHECK_FUNCS([rpcb_getaddr])
   AC_CHECK_HEADER([rpc/rpc.h], , [enable_nis=no])
   AC_CHECK_HEADER([rpcsvc/ypclnt.h], , [enable_nis=no])
   AC_CHECK_HEADER([rpcsvc/yp_prot.h], , [enable_nis=no])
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -628,16 +628,12 @@ dnl Checks for library functions.
...
 AC_PROG_GCC_TRADITIONAL
 AC_FUNC_MEMCMP
-AC_FUNC_VPRINTF
-AC_CHECK_FUNCS(fseeko getdomainname gethostname gettimeofday lckpwdf mkdir select)
-AC_CHECK_FUNCS(strcspn strdup strspn strstr strtol uname)
+AC_CHECK_FUNCS(getdomainname lckpwdf)
 AC_CHECK_FUNCS(getutent_r getpwnam_r getpwuid_r getgrnam_r getgrgid_r getspnam_r getmntent_r)
 AC_CHECK_FUNCS(getgrouplist)
-AC_CHECK_FUNCS(inet_ntop inet_pton innetgr)
-AC_CHECK_FUNCS(quotactl)
```
ldv-alt, line 637: $ac_cv_func_quotactl is still used to define COND_BUILD_PAM_SETQUOTA

configure.ac
```diff
@@ -628,16 +628,12 @@ dnl Checks for library functions.
...
-AC_CHECK_FUNCS(unshare)
```
ldv-alt, line 638: $ac_cv_func_unshare is still used to define COND_BUILD_PAM_NAMESPACE

configure.ac
```diff
@@ -628,16 +628,12 @@ dnl Checks for library functions.
...
+AC_CHECK_FUNCS(innetgr)
 AC_CHECK_FUNCS(explicit_bzero memset_explicit)
-AC_CHECK_FUNCS([ruserok_af ruserok], [break])
```
ldv-alt, line 640: $ac_cv_func_ruserok is still used to define COND_BUILD_PAM_RHOSTS

configure.ac
```diff
@@ -740,7 +736,6 @@ case "$enable_lastlog" in
   yes|check)
     BACKUP_LIBS=$LIBS
     LIBS="$LIBS -lutil"
-    AC_CHECK_FUNCS([logwtmp])
```
ldv-alt, line 743: $ac_cv_func_logwtmp is still used to define COND_BUILD_PAM_LASTLOG

ldv-alt (changes requested):

configure.ac
stoeckmann: It's back in.

configure.ac
stoeckmann: It's back in.

configure.ac
stoeckmann: It's back in.

configure.ac
stoeckmann: It's back in.

stoeckmann:
Thanks for pointing it out. I've mentioned the reason to keep them in commit message in case someone stumble upon them again.

No further matches for the rest of removals after looking for them with "grep -i" in configure.ac.

ldv-alt (approved):

PR #693, opened 2024-01-02
pam_unix: avoid reading uninitialized variable

stoeckmann:
The function `_unix_comesfromsource` calls `_unix_getpwnam` internally. When changing the authentication token, it is first called to read local passwd file and optionally contacting NIS. If an entry is found, `_unix_getpwnam` is called, this time definitely reading passwd file and contacting NIS (if support exists) and parsing the entry.

This is meant to check if the entry is not just available but also valid.

Since the return value of `_unix_getpwnam` is not checked and the supplied pointer is only set in case of success, the check for a NULl pointer afterwards can lead to undefined behavior.

It is easier to call `_unix_getpwnam` directly, check its return value and then check if the entry could be parsed. This in turn reduces the amount of /etc/passwd accesses (and fixes a theoretical TOCTOU race).

### DIFF
--- a/modules/pam_unix/pam_unix_passwd.c
+++ b/modules/pam_unix/pam_unix_passwd.c
@@ -601,6 +601,7 @@ pam_sm_chauthtok(pam_handle_t *pamh, int flags, int argc, const char **argv)
 	int remember = -1;
 	int rounds = 0;
 	int pass_min_len = 0;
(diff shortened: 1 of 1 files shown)


BenBE (approved):
LGTM.

ldv-alt (approved):

PR #695, opened 2024-01-02
modules: use getline instead of fgets

stoeckmann:
This PR changes most `fgets` calls to `getline`. Commits in this PR are quite simple, because we were able to prepare the affected code paths in previous commits.

The `pam_env` module is too complex to be added into this PR because it involves porting `_pam_assemble_line` and will be done in a separate PR.

The `pam_unix` module has one separate PR because of the involved complexity: https://github.com/linux-pam/linux-pam/pull/694


ldv-alt (approved):

stoeckmann:
Rebased and added the `pam_listfile` part.

PR #697, opened 2024-01-03
pam_env: use getline

stoeckmann:
This is effectively a port of libpam's _pam_assemble_line function found in pam_handlers.c.

Noteworthy differences to keep old behavior:
- The backslash of escaped newline is NOT replaced by a blank
- Clears buffer (overwrites everything with \0) in _pam_buffer_clear

This affects pam_env only if compiled without libeconf.


ldv-alt:
Hmm, massive code duplication is not something that can possibly make me happy.

stoeckmann:
> Hmm, massive code duplication is not something that can possibly make me happy.

Not a fan of it either, although there are two subtle difference. And it's been a duplicate (with two subtle differences) before as well. What would you suggest? libpam and its modules have no common library to share functions so far.

ldv-alt:
> What would you suggest? libpam and its modules have no common library to share functions so far.

We have a bunch of pam_modutil_* functions added exactly for that purpose.
Do you think it would be possible to convert `_pam_assemble_line` into a  pam_modutil function that could be used both by libpam and pam_env?

BenBE:
If the above differences are the only ones, than moving the code into a pam_modutil function should be relatively easy. You wouldn't even (strictly speaking) need to remove the overwriting (just make it the default, doesn't actually hurt). Thus the only thing which would need to be treated different would be the handling of line breaks; which could be made a parameter.

stoeckmann:
> We have a bunch of pam_modutil_* functions added exactly for that purpose. Do you think it would be possible to convert `_pam_assemble_line` into a pam_modutil function that could be used both by libpam and pam_env?

The pam_modutil_* functions are not a perfect match:
- These functions are exposed by libpam, but not used internally. This would be the first situation this would happen
- Exposing these functions implies a pretty much locked down API due to backwards compatibility from then on

I have moved the code into a header file. It actually hurts a bit even writing this (let alone implementing it), but in fact this is how another situation in linux-pam is already handled: Debug messages.

If compiled with `--enable-debug`, pretty much every single *.o file contains duplicates of the debug functions, as can be seen when analyzing them (or the final so files). If we fix that one, the same can be applied to pam_assemble_line, which justifies IMHO to split the "avoid function bodies in header files" issue into another PR.

Granted, I have no proper solution for it yet.

ldv-alt:
> I have moved the code into a header file.

Apparently, you've pushed it to #696.

stoeckmann:
Oops, fixed

ldv-alt (approved):

PR #701, opened 2024-01-04
pam_unix updates

cgzones:
See single commit descriptions for more information.

### DIFF
--- a/modules/pam_unix/pam_unix_passwd.c
+++ b/modules/pam_unix/pam_unix_passwd.c
@@ -350,7 +350,7 @@ static int check_old_password(const char *forwho, const char *newpass)
 	if (opwfile == NULL)
 		return PAM_ABORT;
 
-	while (getline(&buf, &n, opwfile) != -1) {
+	for (; getline(&buf, &n, opwfile) != -1; pam_overwrite_n(buf, n)) {
 		if (!strncmp(buf, forwho, len) && (buf[len] == ':' ||
 			buf[len] == ',')) {
 			char *sptr;
@@ -372,6 +372,7 @@ static int check_old_password(const char *forwho, const char *newpass)
 			break;
 		}
 	}
+	pam_overwrite_n(buf, n);
 	free(buf);
 	fclose(opwfile);
 
--- a/modules/pam_unix/pam_unix_passwd.c
+++ b/modules/pam_unix/pam_unix_passwd.c
@@ -286,7 +286,7 @@ static int _unix_run_update_binary(pam_handle_t *pamh, unsigned long long ctrl, 
 	DIAG_POP_IGNORE_CAST_QUAL;
 
 	/* should not get here: exit with error */
-	D(("helper binary is not available"));
+	pam_syslog(pamh, LOG_ERR, "failed to execute %s: %m", UPDATE_HELPER);
 	_exit(PAM_AUTHINFO_UNAVAIL);
     } else if (child > 0) {
 	/* wait for child */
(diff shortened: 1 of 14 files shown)


modules/pam_unix/support.c
```diff
@@ -489,6 +489,7 @@ int _unix_getpwnam(pam_handle_t *pamh, const char *name,
 		}
 	}
 
+	_pam_drop(buf);
```
stoeckmann, line 492: Thanks!

modules/pam_unix/pam_unix_passwd.c
```diff
@@ -346,11 +346,11 @@ static int check_old_password(const char *forwho, const char *newpass)
 	size_t n = 0;
 	size_t len = strlen(forwho);
 
-	opwfile = fopen(OLD_PASSWORDS_FILE, "r");
+	opwfile = fopen(OLD_PASSWORDS_FILE, "re");
 	if (opwfile == NULL)
 		return PAM_ABORT;
 
-	while (getline(&buf, &n, opwfile) != -1) {
+	for (; getline(&buf, &n, opwfile) != -1; pam_overwrite_n(buf, n)) {
```
stoeckmann, line 353: We might call `pam_overwrite_n` here, but it's not guaranteed that `realloc` within `getline` wipes the previous memory area. Not saying that this change should be removed, but it should be considered if reallocation and memory analysis is an issue here.

As a side note: Same applies to https://github.com/linux-pam/linux-pam/pull/696

cgzones: Since `pam_overwrite_n()` gets called in the iteration-expression `realloc(3)` in the condition-expression always operates on cleaned memory.

stoeckmann: It's uncertain how often getline has called realloc internally, e.g. on initial long line. It could be like 32->64->128 for a 100 character long line. It really depends on the getline implementation.

Also, it would depend on the FILE handling of the library if the internal FILE buffer is cleaned, so while this is probably the best attempt that can be done, it's probably not sufficient.

modules/pam_unix/unix_chkpwd.c
```diff
@@ -138,8 +138,11 @@ int main(int argc, char *argv[])
 	  /* if the caller specifies the username, verify that user
 	     matches it */
 	  if (user == NULL || strcmp(user, argv[1])) {
-	    /* no match -> permanently change to the real user and proceed */
-	    if (setuid(getuid()) != 0)
+	    uid_t ruid = getuid();
+	    gid_t rgid = getgid();
+
+	    /* no match -> permanently change to the real user and group, check for no-return, and proceed */
+	    if (setgid(rgid) != 0 || (rgid != 0 && setgid(0) != -1) || setuid(ruid) != 0 || (ruid != 0 && setuid(0) != -1))
```
ldv-alt, line 145: Why `setgid(0)` and `setuid(0)`?
If `rgid != 0`, this code would do `setgid(rgid)` followed by `setgid(0)`, what's the purpose of that?
Likewise with `ruid`.  This doesn't look correct.

modules/pam_unix/passverify.c
```diff
@@ -858,14 +859,14 @@ PAMH_ARG_DECL(int unix_update_passwd,
       freecon(passwd_context_raw);
     }
 #endif
-    pwfile = fopen(PW_TMPFILE, "w");
+    pwfile = fopen(PW_TMPFILE, "we");
```
ldv-alt, line 862: "e" required glibc >= 2.7, which is probably OK nowadays, but let's note this in the commit message.

ldv-alt (changes requested):

modules/pam_unix/unix_chkpwd.c
cgzones: Is a sanity check that privileges are really gone (they are expect to fail (`!= -1`), see https://wiki.sei.cmu.edu/confluence/display/c/POS37-C.+Ensure+that+privilege+relinquishment+is+successful.

ldv-alt (approved):

PR #718, opened 2024-01-13
modules: do not call pam_sm_authenticate

ldv-alt:
Calling an exported function from the module is unsafe as there is no guarantee that the function that will be actually called is the one that is provided by the module.

### DIFF
--- a/modules/pam_lastlog/pam_lastlog.c
+++ b/modules/pam_lastlog/pam_lastlog.c
@@ -665,9 +665,8 @@ cleanup:
 }
 
 /* --- authentication (locking out inactive users) functions --- */
-int
-pam_sm_authenticate(pam_handle_t *pamh, int flags,
-		    int argc, const char **argv)
+static int
+pam_auth(pam_handle_t *pamh, int flags, int argc, const char **argv)
 {
     int retval, ctrl;
     const char *user = NULL;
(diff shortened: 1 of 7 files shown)


modules/pam_lastlog/pam_lastlog.c
```diff
@@ -665,9 +665,8 @@ last_login_failed(pam_handle_t *pamh, int announce, const char *user, time_t llt
 }
 
 /* --- authentication (locking out inactive users) functions --- */
-int
-pam_sm_authenticate(pam_handle_t *pamh, int flags,
-		    int argc, const char **argv)
+static int
+pam_auth(pam_handle_t *pamh, int flags, int argc, const char **argv)
```
stoeckmann, line 669: All other functions got the name of their respective module. Shouldn't this one be called `pam_lastlog` for consistency as well?

ldv-alt: Given that `pam_lastlog.so` provides non-trivial `pam_sm_open_session` and `pam_sm_close_session`,
a function called `pam_lastlog` would be somewhat confusing.

stoeckmann: Ah okay, makes sense then.

PR #720, opened 2024-01-13
pam_succeed_if: handle long and illegal values

stoeckmann:
This PR prevents truncation of very long fields like user, rhost etc. These items can be set by pam_set_item and are therefor not purely related to the feature of arbitrarily long configuration lines.

Also empty strings should not be treated as the numerical value 0.

For easier testing and verification, I have added a new unit test file.

### DIFF
--- a/modules/pam_succeed_if/pam_succeed_if.c
+++ b/modules/pam_succeed_if/pam_succeed_if.c
@@ -70,13 +70,13 @@ evaluate_num(const pam_handle_t *pamh, const char *left,
 
 	errno = 0;
 	l = strtoll(left, &p, 0);
-	if ((p == NULL) || (*p != '\0') || errno) {
+	if ((p == NULL) || (*p != '\0') || (p == left) || errno) {
 		pam_syslog(pamh, LOG_INFO, "\"%s\" is not a number", left);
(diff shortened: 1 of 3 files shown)


modules/pam_succeed_if/pam_succeed_if.c
```diff
@@ -292,7 +292,7 @@ evaluate(pam_handle_t *pamh, int debug,
 	 const char *left, const char *qual, const char *right,
 	 struct passwd **pwd, const char *user)
 {
-	char buf[LINE_MAX] = "";
+	char buf[21] = "";
```
ldv-alt, line 295: I understand `21` is `sizeof(long) * log10(256) + 1`, but could this be written explicitly?
Given that we do not have to save a few bytes, `3` is an acceptable approximation of `log10(256)`.

I'd probably also renamed `buf` to something saying it contains a string representation of a number.

stoeckmann: Adjusted

ldv-alt (approved):
The commit message still mentions `long long conversion (64 bit)` while the code deals with `unsigned long` and doesn't depend on its size.

PR #724, opened 2024-01-15
pam_unix: do not allow comma as field separator

stoeckmann:
The opasswd file shall not use comma as a separator. Enforce colon just like pam_pwhistory does as well.

A comma can be part of a user name, although its usage is discouraged. If such a user exists, it could happen that stored passwords of another user are checked.

Potentially dangerous use of `,` separator spotted with @BenBE.

Proof of Concept:
1. Use `remember` option in /etc/pam.d/passwd:
```
password  required    pam_unix.so       shadow try_first_pass remember=10
```
2. Create `/etc/security/opasswd` file
```
install /dev/null -m 600 /etc/security/opasswd
```
3. Create user `a,b` and set its password three times (first two will be added to /etc/security/opasswd eventually)
```
useradd --badname a,b
passwd a,b           # e.g. "abc"
sudo -u a,b passwd   # e.g. "def"
sudo -u a,b passwd   # e.g. "123"
```
4. Create user `a` and set its password two times (first one passes because it's run by root)
```
useradd a
passwd a             # e.g. "abc"
sudo -u a passwd     # second password of a,b (in this example "def")
```

The user `a` cannot set the password `def` because it was already used by user `a,b`.

### DIFF
--- a/modules/pam_unix/pam_unix_passwd.c
+++ b/modules/pam_unix/pam_unix_passwd.c
@@ -351,14 +351,13 @@ static int check_old_password(const char *forwho, const char *newpass)
 		return PAM_ABORT;
 
 	for (; getline(&buf, &n, opwfile) != -1; pam_overwrite_n(buf, n)) {
-		if (!strncmp(buf, forwho, len) && (buf[len] == ':' ||
-			buf[len] == ',')) {
+		if (!strncmp(buf, forwho, len) && buf[len] == ':') {
 			char *sptr;
 			buf[strlen(buf) - 1] = '\0';
-			/* s_luser = */ strtok_r(buf, ":,", &sptr);
-			/* s_uid = */ strtok_r(NULL, ":,", &sptr);
-			/* s_npas = */ strtok_r(NULL, ":,", &sptr);
-			s_pas = strtok_r(NULL, ":,", &sptr);
+			/* s_luser = */ strtok_r(buf, ":", &sptr);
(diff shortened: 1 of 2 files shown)


stoeckmann:
The use of `remember` is already discouraged in manual page and the passwords are saved as MD5. I am not sure if there is any form of required backwards-compatibility with these comma separators.

If that's the case, we can just skip this PR.

ldv-alt (commented):
After this change, in pam_unix_passwd.c  there is
`s_pas = strtok_r(NULL, ",", &sptr);`
while in passverify.c there is
`s_pas = strtok_r(NULL, "", &sptr);`.

I don't quite follow where does this difference come from.

stoeckmann:
> I don't quite follow where does this difference come from.

Yeah, it's tricky. In passverify.c it has to be `""` because `s_pas` will be used in `fprintf` to add the already existing password hashes to opasswd. A strtok_r with `,` would insert a `\0` character, which removes all password hashes except the first.

That's why passverify.c uses `strpbrk` in the while-loop, too.

ldv-alt:
> Yeah, it's tricky. In passverify.c it has to be `""` because `s_pas` will be used in `fprintf` to add the already existing password hashes to opasswd. A strtok_r with `,` would insert a `\0` character, which removes all password hashes except the first.
> 
> That's why passverify.c uses `strpbrk` in the while-loop, too.

OK, so it's a slightly different change: `:` used as a separator where it shouldn't have been.

PR #725, opened 2024-01-15
pam_pwhistory: parse opasswd lines verbatim

stoeckmann:
Users may have a hash character in their name, which would be removed. This in turn effectively defeats the purpose of pam_pwhistory for the user.

Proof of Concept:
1. Use pam_pwhistory as suggested in its manual page in /etc/pam.d/passwd:
```
password     required       pam_pwhistory.so
password     required       pam_unix.so        use_authtok shadow
```
2. Create user `a#b` and set its password twice
```
useradd --badname a#b
passwd a#b             # e.g. "abc"
passwd a#b             # e.g. "def"
```
3. Try to use the first password again
```
passwd a#b             # first password (in this example "abc")
```

The password should not be accepted. Yet, every call adds a new line to /etc/security/opasswd instead, because the removal of comments makes the lines unparseable for pam_pwhistory.

### DIFF
--- a/modules/pam_pwhistory/opasswd.c
+++ b/modules/pam_pwhistory/opasswd.c
@@ -169,23 +169,20 @@ check_old_pass, const char *user, const char *newpass, const char *filename, int
 
   while (!feof (oldpf))
     {
-      char *cp;
       ssize_t n = getline (&buf, &buflen, oldpf);
-
-      cp = buf;
 
       if (n < 1)
         break;
 
-      cp[strcspn(cp, "\n")] = '\0';
-      if (*cp == '\0')        /* ignore empty lines */
+      buf[strcspn(buf, "\n")] = '\0';
+      if (buf[0] == '\0')        /* ignore empty lines */
         continue;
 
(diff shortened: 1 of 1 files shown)


stoeckmann:
As a side note: I have tried to add a user with blanks in front, e.g. `   abc` but `getpwnam` fails on my glibc-based system, yet `passwd --badname` accepts it and writes a proper line to /etc/passwd. We should still avoid removing blanks at beginning of opasswd lines.

stoeckmann:
Last but not least: The pam_pwhistory module also stores passwords which have been set by root. This by itself is not bad directly, but the hashes within opasswd are not properly checked for equality.

If, for example, remember is set to 10 and root sets the same password 10 times, then all previously set passwords are forgotten.

Not sure if this edge case has to be documented though.

ldv-alt (approved):
I'm not sure what do you mean by "parse lines directly" because they have been parsed directly all the time.  The patch removes processing of comments and leading whitespace. Maybe "verbatim"?

stoeckmann:
> Maybe "verbatim"?

Fine for me. Adjusted.

PR #727, opened 2024-01-16
modules: zero out crypt_r(3) data before usage

cgzones:
The manual page of crypt_r(3) recommends to zero the entire data object.

### DIFF
--- a/modules/pam_pwhistory/opasswd.c
+++ b/modules/pam_pwhistory/opasswd.c
@@ -121,21 +121,32 @@ parse_entry (char *line, opwd *data)
   return 0;
 }
 
+/* Return 1 if the passwords are equal, 0 if they are not, and -1 on error. */
 static int
 compare_password(const char *newpass, const char *oldpass)
 {
   char *outval;
   int retval;
 #ifdef HAVE_CRYPT_R
-  struct crypt_data output = { 0 };
+  struct crypt_data *cdata;
 
-  outval = crypt_r (newpass, oldpass, &output);
+  cdata = calloc(1, sizeof(*cdata));
+  if (!cdata)
+    return -1;
+
+  outval = crypt_r (newpass, oldpass, cdata);
 #else
   outval = crypt (newpass, oldpass);
 #endif
 
   retval = outval != NULL && strcmp(outval, oldpass) == 0;
+
+#ifdef HAVE_CRYPT_R
+  pam_overwrite_object(cdata);
+  free(cdata);
+#else
   pam_overwrite_string(outval);
+#endif
   return retval;
 }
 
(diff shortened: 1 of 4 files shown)


modules/pam_pwhistory/opasswd.c
```diff
@@ -129,6 +129,7 @@ compare_password(const char *newpass, const char *oldpass)
 #ifdef HAVE_CRYPT_R
   struct crypt_data output;
 
+  memset(&output, 0, sizeof(struct crypt_data));
   output.initialized = 0;
```
stoeckmann, line 133: Setting `initialized = 0` at this and the other occurrences is not needed anymore.

modules/pam_pwhistory/opasswd.c
```diff
@@ -129,6 +129,7 @@ compare_password(const char *newpass, const char *oldpass)
 #ifdef HAVE_CRYPT_R
   struct crypt_data output;
 
+  memset(&output, 0, sizeof(struct crypt_data));
```
stoeckmann, line 132: Also `struct crypt_data output = { 0 }` instead simplifies this statement.

modules/pam_userdb/pam_userdb.c
```diff
@@ -287,7 +287,7 @@ user_lookup (pam_handle_t *pamh, const char *database, const char *cryptmode,
 	    } else {
 #ifdef HAVE_CRYPT_R
 	      struct crypt_data *cdata = NULL;
-	      cdata = malloc(sizeof(*cdata));
+	      cdata = calloc(1, sizeof(*cdata));
```
stoeckmann, line 290: Not part of the change introduced here, but the `cdata` variable is not overridden after its usage. If this does not belong on this PR, it should be fixed later on. Otherwise an extra commit might be useful here.

And talking about it, I think the allocation can be skipped because `cdata` could reside in stack, which removes the error handling if allocation fails.

modules/pam_pwhistory/opasswd.c
cgzones: Applied.

modules/pam_pwhistory/opasswd.c
cgzones: Dropped the unnecessary explicit assignments of `->initialized`.

modules/pam_userdb/pam_userdb.c
cgzones: Added a commit to zero out the data.

Do we want to use the stack for `struct crypt_data`, since its size seems to be 32768 bytes?


It's currently used once on the stack here: https://github.com/linux-pam/linux-pam/blob/cf8c50f2c76fd29c018a2220d243f41801aaadb1/modules/pam_pwhistory/opasswd.c#L130
and otherwise on the heap.

modules/pam_pwhistory/opasswd.c
```diff
@@ -127,9 +127,7 @@ compare_password(const char *newpass, const char *oldpass)
   char *outval;
   int retval;
 #ifdef HAVE_CRYPT_R
-  struct crypt_data output;
-
-  output.initialized = 0;
+  struct crypt_data output = { 0 };;
```
stoeckmann, line 130: Style note: two semicolons

modules/pam_userdb/pam_userdb.c
stoeckmann: Okay, knowing about its size now I am definitely against having it on stack. Especially since PAM might be used in threads which should not be that excessive with their stacks.

At a later point in time (not part of this PR IMHO since it's all about clearing data) the stack usage in pam_pwhistory should disappear instead.

cgzones: Added commit to move the allocation to the heap.

stoeckmann:
Looks good to me.

ldv-alt (approved):

PR #728, opened 2024-01-16
Two build system tweaks

cgzones:
* configure: require libsystemd on --enable-logind

  Fail if `--enable-logind` is specified while libsystemd cannot be found, so the feature is reliably enabled on success.

* doc: do not clean bootstrapped files

  The two files custom-html.xsl and custom-man.xsl are created at configure time.  Only delete them on `make distclean` instead of `make clean` to be able to re-generate manual pages afterwards.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -4,7 +4,8 @@
 
 SUBDIRS = man specs sag adg mwg
 
-CLEANFILES = *~ custom-html.xsl custom-man.xsl
+CLEANFILES = *~
+DISTCLEANFILES = custom-html.xsl custom-man.xsl
 
 dist_html_DATA = index.html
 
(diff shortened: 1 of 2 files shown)


doc/Makefile.am
```diff
@@ -4,7 +4,8 @@
 
 SUBDIRS = man specs sag adg mwg
 
-CLEANFILES = *~ custom-html.xsl custom-man.xsl
+CLEANFILES = *~
+DISTCLEANFILES = custom-html.xsl custom-man.xsl
```
stoeckmann, line 8: Fixes b285afe0ea45f391b819b9f61316827c59ba6c56 which was not included in a release. Strongly agree and thanks a lot for spotting and describing the consequences!

ldv-alt:
> * configure: require libsystemd on --enable-logind
>       Fail if `--enable-logind` is specified while libsystemd cannot be found, so the feature is reliably enabled on success.

It's on request rather than on success.

ldv-alt (approved):

PR #734, opened 2024-01-17
include stdint.h in pam_namespace.c

### LINKED ISSUE #733: missing include in `modules/pam_namespace/pam_namespace.c`
`modules/pam_namespace/pam_namespace.c` makes use of `SIZE_MAX` but doesn't include `stdint.h`; resulting in build failures on 1.6.0.
```
pam_namespace.c: In function 'process_line':
pam_namespace.c:649:41: error: 'SIZE_MAX' undeclared (first use in this function)
  649 |         if (count > UINT_MAX || count > SIZE_MAX / sizeof(uid_t)) {
      |                                         ^~~~~~~~
pam_namespace.c:41:1: note: 'SIZE_MAX' is defined in header '<stdint.h>'; did you forget to '#include <stdint.h>'?
   40 | #include "argv_parse.h"
  +++ |+#include <stdint.h>
   41 | 
pam_namespace.c:649:41: note: each undeclared identifier is reported only once for each function it appears in
  649 |         if (count > UINT_MAX || count > SIZE_MAX / sizeof(uid_t)) {
      |                                         ^~~~~~~~
make[3]: *** [Makefile:819: pam_namespace.lo] Error 1
```

the following fixes it:
```
sed -i -e '/argv_parse\.h/a\
#include <stdint.h>\
' modules/pam_namespace/pam_namespace.c
```

jhheider:
missing include causes compilation errors.

closes #733

### DIFF
--- a/modules/pam_namespace/pam_namespace.c
+++ b/modules/pam_namespace/pam_namespace.c
@@ -34,6 +34,8 @@
 
 #define _ATFILE_SOURCE
 
+#include "config.h"
+#include <stdint.h>
 #include "pam_cc_compat.h"
 #include "pam_inline.h"
 #include "pam_namespace.h"


modules/pam_namespace/pam_namespace.c
```diff
@@ -38,6 +38,7 @@
 #include "pam_inline.h"
 #include "pam_namespace.h"
 #include "argv_parse.h"
+#include <stdint.h>
```
stoeckmann, line 41: Thank you for the PR. It's a shame that this was neither triggered on my system nor in the pipelines. It would be nice if you could move it above the other includes for consistency reasons.

Otherwise definitely fully agree, of course.

ldv-alt (approved):
I added `#include "config.h"` because it has to be included prior to any system header.
Also, I copied relevant information from #733 to the commit message.

PR #735, opened 2024-01-17
pam_unix: support shadow passwd -k

stoeckmann:
If shadow's passwd is called with option -k, then only expired passwords should be changeable.

This effectively adds support of PAM_CHANGE_EXPIRED_AUTHTOK to pam_unix as described in pam_chauthtok manual page.


stoeckmann:
Waiting for possible regressions after 1.6.0 release is definitely over now. Let's get this feature in.

Issue #736, opened 2024-01-18
pam-1.6.0 build error: too few arguments to function '_pam_parse_conf_file'


hdejin:
I'm trying to build 1.6.0 and getting this error:

pam_handlers.c:503:30: error: too few arguments to function '_pam_parse_conf_file'
  503 |                     retval = _pam_parse_conf_file(pamh, f, NULL, PAM_T_ANY, 0, 1);

ldv-alt:
Apparently, --enable-read-both-confs configure option is still used.
@BenBE: it's a regression introduced by commit d686b51a7c58a3fc288030384d1de3e4bc0ebeee.

BenBE:
In initial calls to that functions the level argument has to be duplicated; will take a closer look at a fix later today.

stoeckmann:
I have created a PR which fixes this issue.

PR #737, opened 2024-01-18
libpam: fix build with --enable-read-both-confs

stoeckmann:
If configure option --enable-read-both-confs is used, the build fails with 1.6.0 due to missing stack level depth argument passed to _pam_parse_conf_file.

Solves https://github.com/linux-pam/linux-pam/issues/736.


ldv-alt (approved):

ldv-alt:
I've also added `Resolves` and `Fixes` tags to the commit message.

BenBE (approved):
LGTM. Wouldn't have done anything else either … ;-)

Issue #738, opened 2024-01-18
v1.6.0: pam_env test always crashes in pam_open_session()


thkukuk:
Regression compared to 1.5.3.
We see on all builds that `make check` crashes in tst-pam_env-retval:
```
[  131s] make[4]: Entering directory '/home/abuild/rpmbuild/BUILD/Linux-PAM-1.6.0/modules/pam_env'
[  131s] ../../build-aux/test-driver: line 112: 15021 Aborted                 (core dumped) "$@" >> "$log_file" 2>&1
[  131s] FAIL: tst-pam_env-retval
...
[  131s] tst-pam_env-retval.c:187: Assertion failed: PAM_PERM_DENIED (0x6) == pam_open_session(pamh, 0) (0)
[  131s] FAIL tst-pam_env-retval (exit status: 134)

```

It is always in `pam_open_session()`, but not always the same call to it. strace shows a wired mix of loading modules, libraries and config files from the running system and from the build system.

ldv-alt:
There shouldn't be any pam modules or pam config files loaded from the running system.

ldv-alt:
Could it be something related to --enable-econf configuration?  We don't test this in CI yet because it needs libeconf >= 0.5.0.

stoeckmann:
It's my first assumption as well. The `pam_env` module is the first econf module covered with a test. `pam_shells` uses econf, but has no test.

thkukuk:
That was our assumption first, too, but disabling libeconf did not change the picture.

stoeckmann:
Does it happen with other tests as well? I mean, did you try to disable the test and let others run, too?

ldv-alt:
Also, tst-pam_env-retval.c was already present in v1.5.3.

thkukuk:
In v1.5.3 it worked
Didn't had yet the time to diff 1.5.3 with 1.6.0, tomorrow...

ldv-alt:
Tried to reproduce locally,
```
strace -e%file -o'|grep env' .libs/lt-tst-pam_env-retval
```
doesn't show any attempts to access system pam files.

stoeckmann:
Can reproduce it. Not sure if this is how it happened, though:

```
./configure --enable-vendordir=/tmp
make
install -D modules/pam_env/pam_env.conf /tmp/security/pam_env.conf
make check
```

Since VENDORDIR is set, the supposedly non-existing file is parsed. This leads to a successful pam_open_session which in turn makes the test fail.

[...]

PR #740, opened 2024-01-18
pam_env: fix --enable-vendordir fallback logic

ldv-alt:
* modules/pam_env/pam_env.c (_parse_config_file) [!USE_ECONF && VENDOR_DEFAULT_CONF_FILE]: Do not fallback to vendor pam_env.conf file if the config file is specified via module arguments.

Link: https://github.com/linux-pam/linux-pam/issues/738
Fixes: v1.5.3~69 ("pam_env: Use vendor specific pam_env.conf and environment as fallback")


stoeckmann:
Yes, makes sense and fixes my build with `--enable-vendordir` option.

ldv-alt:
@stoeckmann: I've changed the patch slightly so that the code is hopefully slightly easier to read, please give it another try.

stoeckmann:
Still works and looks good.

PR #741, opened 2024-01-18
pam_env: remove escaped newlines from econf lines

stoeckmann:
The libeconf routines do not remove escaped newlines the way we want to process them later on. Manually remove them from values.

### DIFF
--- a/modules/pam_env/pam_env.c
+++ b/modules/pam_env/pam_env.c
@@ -160,6 +160,28 @@ isDirectory(const char *path) {
    return S_ISDIR(statbuf.st_mode);
 }
 
+/*
+ * Remove escaped newline from string.
+ *
+ * All occurrences of "\\n" will be removed from string.
+ */
+static void
+econf_unescnl(char *val)
+{
+    char *dest, *p;
+
+    dest = p = val;
+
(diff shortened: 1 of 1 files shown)


ldv-alt:
Can we be sure that libeconf routines won't do it, so we wouldn't end up doing double unescaping?

stoeckmann:
An alternative approach would be to remove this edge case from the test. Another limitation that exists if econf is in use is a fixed maximum line length.

So, considering that we have differences anyway, we might skip these special situations from tests. Or add more ifdefs to the test.

ldv-alt:
Ideally, there shouldn't be any material differences, otherwise config files would have different meaning depending on whether libeconf is used or not.

ldv-alt (approved):
Let's hope libeconf won't start doing its own unescaping of newlines.

ldv-alt:
I've also added `Resolves` and `Fixes` tags to the commit message.

schubi2:
> > Another limitation that exists if econf is in use is a fixed maximum line length.
> 
> libeconf should not have such a limitation. If there is one somewhere, it's a bug.

It has been a bug. I have fixed it and have released a new version of libeconf (0.6.1)

PR #746, opened 2024-01-19
build: correctly set WITH_SELINUX conditional

stoeckmann:
React on actual test if SELinux is available, not just if SELinux should be tested for.

Currently the supposedly disabled binaries are still installed even if SELinux is not available.

Fixes: cb9f88ba944d ("pam_unix: build unix_update only with SELinux enabled")

Reported here: https://wiki.linuxfromscratch.org/blfs/ticket/19142

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -529,12 +529,12 @@ fi
 AC_ARG_ENABLE([selinux],
         AS_HELP_STRING([--disable-selinux],[do not use SELinux]),
         WITH_SELINUX=$enableval, WITH_SELINUX=yes)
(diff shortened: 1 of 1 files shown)


stoeckmann:
CC @cgzones, it would be nice if you can review this one.

ldv-alt (approved):
I'd say the build should fail if it was configured using --enable-selinux but SELinux was not available.  Consequently, the default would have to be `WITH_SELINUX=check` instead of `WITH_SELINUX=yes`.
Unfortunately, our configure script is not consistent in this respect.
Anyway, it's far beyond the scope of this bug.

PR #749, opened 2024-01-20
configure: fail if specified option cannot be satisfied

cgzones:
The options whether to build with support for libprelude, libaudit, libselinux, or libeconf are set to enable-if-available.  These options also have a configure flag `--enable-foo`, which currently fall back to the feature being disabled if not available.

Change these feature flags to fail if specified explicitly and the required dependencies cannot be satisfied.

Prompted by #728 and #746


configure.ac
```diff
@@ -543,6 +547,8 @@ if test -n "$LIBSELINUX" ; then
     AC_CHECK_FUNCS(setkeycreatecon)
     AC_CHECK_FUNCS(getseuser)
     LIBS=$BACKUP_LIBS
+elif test "$WITH_SELINUX" = "yno" ; then
```
ldv-alt, line 550: This doesn't look correct.

ldv-alt (changes requested):

ldv-alt (approved):

PR #751, opened 2024-01-20
Unify PAM return value usage

cgzones:
PAM defines a set of common return-values, like `PAM_SUCCESS` or `PAM_AUTH_ERR`, which are used internally and in public API.  Commonly these return-values are stored in variables named `retval`.

Unify the naming of variables such that if-and-only-if a variable is named `retval` it holds a PAM style return-value.

Also use explicitly `PAM_SUCESS` instead of the value 0 in assignments and comparisons.


stoeckmann:
While I agree in using PAM_SUCCESS instead of 0, I'm not sure where the advantage of unification of retval is. It's not like it's any special term. Contrary, it's just a short version of "return value". How should anyone guarantee that future commits won't use "retval" as return value for something else again?

ldv-alt:
You could play with creating a enum, e.g. `pam_retval_t`, and using it consistently instead of `int`, although I'm not sure we can change the public API this way, even though the enum and `int` would have the same representation.

PR #753, opened 2024-01-21
pam_faildelay: improve delay parsing

stoeckmann:
Use `long long` on 32 bit systems to also support `UINT_MAX` without falling back to strtoul, which automatically converts a negative number to a positive value.

While at it, prevent more invalid numbers.

Proof of Concept:

This line triggers a syslog warning:
```auth            optional        pam_faildelay.so delay=abc```

This one does not:
```auth            optional        pam_faildelay.so delay=1abc```

### DIFF
--- a/modules/pam_faildelay/pam_faildelay.c
+++ b/modules/pam_faildelay/pam_faildelay.c
@@ -79,12 +79,12 @@
 
 /* --- authentication management functions (only) --- */
 
-static long parse_delay(const char *val)
+static long long parse_delay(const char *val)
 {
-    long delay;
+    long long delay;
     char *endptr;
 
-    delay = strtol (val, &endptr, 10);
+    delay = strtoll (val, &endptr, 10);
     if (delay == 0 && val == endptr)
       return -1;
     return delay;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/pam_faildelay/pam_faildelay.c [hunk 1]
```diff
@@ -79,13 +79,13 @@
...
 {
-    long delay;
+    long long delay;
     char *endptr;
 
-    delay = strtol (val, &endptr, 10);
-    if (delay == 0 && val == endptr)
+    delay = strtoll (val, &endptr, 10);
+    if (delay < 1 || val == endptr || *endptr != '\0')
       return -1;
```
ldv-alt, line 89: This looks like an unintended change of behavior: "0" which used to be a valid delay is now turned into an error.

modules/pam_faildelay/pam_faildelay.c
```diff
@@ -94,14 +94,14 @@ int pam_sm_authenticate(pam_handle_t *pamh, int flags UNUSED,
...
-    long delay = -1;
+    long long delay = -1;
 
     /* step through arguments */
     for (i = 0; i < argc; i++) {
 	const char *val = pam_str_skip_prefix (argv[i], "delay=");
 	if (val != NULL) {
 	  delay = parse_delay (val);
-	  if (delay < 0 || (unsigned long) delay > UINT_MAX)
+	  if (delay < 0 || (unsigned long long) delay > UINT_MAX)
```
ldv-alt, line 104: The `delay < 0` condition is now redundant, but I suggest keeping it here anyway and let the compiler optimize this.

ldv-alt (changes requested):

modules/pam_faildelay/pam_faildelay.c
stoeckmann: Ok, I'll keep it. But thanks for pointing out!

modules/pam_faildelay/pam_faildelay.c
stoeckmann: Oh, true. Fixed.

ldv-alt (approved):

PR #754, opened 2024-01-21
modules: add pamc headers to the search path only when needed

stoeckmann:
The pam client library libpamc is only needed if libpam_misc is in use. But libpam_misc is only used by an SELinux helper binary.

Remove the libpamc includes from search path in all other cases.

No functional change, but spotted while experimenting with ideas how to properly handle `pam_assemble_line` and the debug features which are currently stored in header files.

### DIFF
--- a/modules/pam_access/Makefile.am
+++ b/modules/pam_access/Makefile.am
@@ -21,8 +21,7 @@ else
 secureconfdir = $(SCONFIGDIR)
 endif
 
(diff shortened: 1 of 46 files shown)


ldv-alt (approved):

ldv-alt:
I suggest replacing `include pamc headers` with `add pamc headers to the search path` to avoid confusion with the headers being included from .h and .c files.

stoeckmann:
Adjusted

PR #755, opened 2024-01-22
libpam_internal: supply debug functionality

stoeckmann:
Move function bodies from headers into dedicated object files stored in libpam_internal. This library won't be installed.

Keep the debug function body in header, even though disabled when building Linux-PAM, to stay API compatible with previous versions.

### DIFF
--- a/libpam/include/security/_pam_macros.h
+++ b/libpam/include/security/_pam_macros.h
@@ -94,11 +94,25 @@ do {                                              \
 #define _PAM_LOGFILE "/var/run/pam-debug.log"
 #endif
 
+#ifdef PAM_NO_HEADER_FUNCTIONS
+UNUSED
+extern void _pam_output_debug_info(const char *file, const char *fn
+				   , const int line);
+UNUSED
+PAM_FORMAT((printf, 1, 2))
+extern void _pam_output_debug(const char *format, ...);
+#else
+#ifdef PAM_DEBUG_C
+#define PAM_DEBUG_SCOPE
+#else
+#define PAM_DEBUG_SCOPE static
+#endif
+
 #ifdef UNUSED
 UNUSED
 #endif
-static void _pam_output_debug_info(const char *file, const char *fn
-				   , const int line)
+PAM_DEBUG_SCOPE void _pam_output_debug_info(const char *file, const char *fn
+					    , const int line)
 {
     FILE *logfile;
     int must_close = 1, fd;
@@ -127,7 +141,7 @@ static void _pam_output_debug_info(const char *file, const char *fn
 UNUSED
 #endif
 PAM_FORMAT((printf, 1, 2))
-static void _pam_output_debug(const char *format, ...)
+PAM_DEBUG_SCOPE void _pam_output_debug(const char *format, ...)
 {
     va_list args;
     FILE *logfile;
@@ -157,6 +171,8 @@ static void _pam_output_debug(const char *format, ...)
 
     va_end(args);
 }
+#undef PAM_DEBUG_SCOPE
+#endif
 
 #define D(x) do { \
     _pam_output_debug_info(__FILE__, __FUNCTION__, __LINE__); \
--- a/libpam_internal/pam_debug.c
+++ b/libpam_internal/pam_debug.c
@@ -1,0 +1,21 @@
+/*
+ * This provides the necessary functions to do debugging in PAM.
+ * Cristian Gafton <gafton@redhat.com>
+ */
+
+#include "config.h"
+
+#ifdef PAM_DEBUG
+
+#include "security/_pam_macros.h"
+
+#undef PAM_MACROS_H
+#undef PAM_NO_HEADER_FUNCTIONS
+#define PAM_DEBUG_C 1
+#include "security/_pam_macros.h"
+
+#else
+
+extern int ISO_C_forbids_an_empty_translation_unit;
+
+#endif /* PAM_DEBUG */
--- a/libpam_misc/Makefile.am
+++ b/libpam_misc/Makefile.am
@@ -15,7 +15,8 @@ libpam_misc_la_LDFLAGS = -no-undefined -version-info 82:1:82
 if HAVE_VERSIONING
   libpam_misc_la_LDFLAGS += -Wl,--version-script=$(srcdir)/libpam_misc.map
 endif
-libpam_misc_la_LIBADD = $(top_builddir)/libpam/libpam.la
+libpam_misc_la_LIBADD = $(top_builddir)/libpam_internal/libpam_internal.la \
+			$(top_builddir)/libpam/libpam.la
 
 lib_LTLIBRARIES = libpam_misc.la
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,8 @@
 
 AUTOMAKE_OPTIONS = 1.9 gnu dist-xz no-dist-gzip check-news
 
-SUBDIRS = libpam tests libpamc libpam_misc modules po conf xtests
+SUBDIRS = libpam_internal libpam tests libpamc libpam_misc modules po conf \
+	  xtests
 
 if HAVE_DOC
 SUBDIRS += doc
--- a/configure.ac
+++ b/configure.ac
@@ -249,6 +249,7 @@ if test x"$enable_debug" = x"yes" ; then
    AC_DEFINE([PAM_DEBUG],,
 		[lots of stuff gets written to /var/run/pam-debug.log])
 fi
+AC_DEFINE([PAM_NO_HEADER_FUNCTIONS],1,[disable function bodies in headers])
 
 AC_ARG_ENABLE(docbook_rng,
 	AS_HELP_STRING([--enable-docbook-rng=FILE],[RNG file for checking XML files @<:@default=http://docbook.org/xml/5.0/rng/docbookxi.rng@:>@]),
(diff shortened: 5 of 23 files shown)


stoeckmann:
@ldv-alt what do you think? Is this the approach we should use to also move `pam_assemble_line` and possibly the string buffering at some time as well (if it's actually needed in multiple modules)?

ldv-alt:
> @ldv-alt what do you think? Is this the approach we should use to also move `pam_assemble_line` and possibly the string buffering at some time as well (if it's actually needed in multiple modules)?

IMHO it makes perfect sense.

libpam/include/security/_pam_macros.h
```diff
@@ -94,6 +94,18 @@ do {                                              \
 #define _PAM_LOGFILE "/var/run/pam-debug.log"
 #endif
 
+#ifdef PAM_NO_HEADER_FUNCTIONS
+#ifdef UNUSED
+UNUSED
+#endif
```
ldv-alt, line 100: Given that `UNUSED` is always defined when Linux-PAM is built, let's use it unconditionally when `PAM_NO_HEADER_FUNCTIONS` is defined.

libpam_internal/pam_debug.c
```diff
@@ -0,0 +1,110 @@
...
+ * _PAM_LOGFILE must exist and be writable to the programs you debug.
+ */
+
+#ifndef _PAM_LOGFILE
+#define _PAM_LOGFILE "/var/run/pam-debug.log"
+#endif
+
+#ifdef UNUSED
+UNUSED
+#endif
```
ldv-alt, line 34: Let's use `UNUSED` unconditionally in all files besides installed header files.

ldv-alt (changes requested):

libpam/include/security/_pam_macros.h
stoeckmann: Adjusted

libpam_internal/pam_debug.c
stoeckmann: And adjusted as well

stoeckmann: Oh, didn't notice that this comment was directed at pam_debug.c. Now it's adjusted.

stoeckmann:
I have successfully compiled this branch with `--enable-selinux`. Is there something else I should try before removing the Draft status? Not sure if there is any other binary hidden somewhere which is only compiled if a specific configure option is set.

libpam_misc/Makefile.am
```diff
@@ -15,7 +15,8 @@ libpam_misc_la_LDFLAGS = -no-undefined -version-info 82:1:82
 if HAVE_VERSIONING
   libpam_misc_la_LDFLAGS += -Wl,--version-script=$(srcdir)/libpam_misc.map
 endif
-libpam_misc_la_LIBADD = $(top_builddir)/libpam/libpam.la
+libpam_misc_la_LIBADD = $(top_builddir)/libpam_internal/libpam_internal.la \
+			$(top_builddir)/libpam/libpam.la
```
ldv-alt, line 19: Another question is whether libpam_internal.la should be listed here and in other places before or after libpam.la.
The rule of thumb is that the library that's used by other libraries should be listed after the libraries which use it.

So if libpam_internal uses libpam, then libpam_internal should be listed before libpam, and vice versa.

stoeckmann: In general I agree. In this case though, the nature of libpam_internal leads to the effect that it's fully included in libpam, so there is no real inter-dependency going on.

But I'm fine with reordering these lines to have libpam_internal in front of libpam if you want to.

ldv-alt:
> I have successfully compiled this branch with `--enable-selinux`. Is there something else I should try before removing the Draft status? Not sure if there is any other binary hidden somewhere which is only compiled if a specific configure option is set.

If you think the change is ready, then it's time to remove the Draft status.

libpam_misc/Makefile.am
stoeckmann: Since it won't hurt and it's correct from a logical point of view (even if it's not technically required), I've adjusted the ordering.

libpam_internal/pam_debug.c
```diff
@@ -0,0 +1,102 @@
...
+
+#ifndef _PAM_LOGFILE
+#define _PAM_LOGFILE "/var/run/pam-debug.log"
+#endif
+
+UNUSED
+void _pam_output_debug_info(const char *file, const char *fn, const int line);
+UNUSED
+PAM_FORMAT((printf, 1, 2))
+void _pam_output_debug(const char *format, ...);
```
ldv-alt, line 36: Can't we just `#include "security/_pam_macros.h"` and obtain these declarations from that header?

libpam_internal/pam_debug.c
```diff
@@ -0,0 +1,102 @@
...
+
+UNUSED
```
ldv-alt, line 38: Function attributes likes `UNUSED` are needed for function prototypes only.

libpam_internal/pam_debug.c
```diff
@@ -0,0 +1,102 @@
...
+    }
+    fprintf(logfile,"[%s:%s(%d)] ",file, fn, line);
+    fflush(logfile);
+    if (must_close)
+        fclose(logfile);
+}
+
+UNUSED
+PAM_FORMAT((printf, 1, 2))
+void _pam_output_debug(const char *format, ...)
```
ldv-alt, line 67: Function attributes likes `UNUSED` and `PAM_FORMAT` are needed for function prototypes only.

ldv-alt (commented):
This way we end up with two copies of `_pam_output_debug_info` and `_pam_output_debug` functions definitions: one in security/_pam_macros.h, and another one in libpam_internal/pam_debug.c; is there any way to avoid this duplication?

stoeckmann:
> This way we end up with two copies of `_pam_output_debug_info` and `_pam_output_debug` functions definitions: one in security/_pam_macros.h, and another one in libpam_internal/pam_debug.c; is there any way to avoid this duplication?

Not without breaking API if there is some software out there using _pam_macros.h directly without using anything else like libpam. That's why I added the definition to disable them for Linux-PAM.

libpam_internal/pam_debug.c
stoeckmann: True, that's easier.

libpam_internal/pam_debug.c
stoeckmann: Removed

stoeckmann:
> This way we end up with two copies of `_pam_output_debug_info` and `_pam_output_debug` functions definitions: one in security/_pam_macros.h, and another one in libpam_internal/pam_debug.c; is there any way to avoid this duplication?

What we COULD do though is including the header and disabling the definition explicitly right before doing so. This way we use the header function bodies only in this one object file...

ldv-alt:
> > This way we end up with two copies of `_pam_output_debug_info` and `_pam_output_debug` functions definitions: one in security/_pam_macros.h, and another one in libpam_internal/pam_debug.c; is there any way to avoid this duplication?
> 
> What we COULD do though is including the header and disabling the definition explicitly right before doing so. This way we use the header function bodies only in this one object file...

We would have to tweak the function prototypes as well so they won't be declared static.

stoeckmann:
> We would have to tweak the function prototypes as well so they won't be declared static.

Perhaps making the `static` keyword "DEFINE"able so this one file can actually remove the `static` keyword?

ldv-alt:
> > We would have to tweak the function prototypes as well so they won't be declared static.
> 
> Perhaps making the `static` keyword "DEFINE"able so this one file can actually remove the `static` keyword?

I'd rather used some new keyword instead of `static`, and defined it appropriately.

stoeckmann:
> I'd rather used some new keyword instead of `static`, and defined it appropriately.

I have added a new commit as a proposal. Maybe this way?

stoeckmann:
Looking at this code mess, it's probably easier to import `_pam_macros.h` in `pam_debug.c`, call the static functions from new functions without an underline which are not static, and then introduce a new `pam_debug.h` file which others use.

libpam/include/security/_pam_macros.h
```diff
@@ -94,11 +94,31 @@ do {                                              \
 #define _PAM_LOGFILE "/var/run/pam-debug.log"
 #endif
 
+#ifdef PAM_NO_HEADER_FUNCTIONS
+#ifdef PAM_DEBUG_C
+#define PAM_DEBUG_TYPE void
+#else
+#define PAM_DEBUG_TYPE extern void
+#endif
```
ldv-alt, line 102: I think we can just define `PAM_DEBUG_TYPE` unconditionally here because `void` and `extern void` mean the same in declarations.

libpam/include/security/_pam_macros.h
```diff
@@ -94,11 +94,31 @@ do {                                              \
...
+UNUSED
+PAM_FORMAT((printf, 1, 2))
+PAM_DEBUG_TYPE _pam_output_debug(const char *format, ...);
+#undef PAM_DEBUG_TYPE
+#else
+#ifdef PAM_DEBUG_C
+#define PAM_DEBUG_TYPE void
+#else
+#define PAM_DEBUG_TYPE static void
+#endif
```
ldv-alt, line 115: I suggest not to include `void` into PAM_DEBUG_TYPE.
Let's rename it to `PAM_DEBUG_SCOPE` or something like this because it's all about presence or absence of `static` keyword.

ldv-alt (commented):
Could you please squash these two commits so it would be easier to review?

libpam/include/security/_pam_macros.h
stoeckmann: In that case we can drop the whole PAM_DEBUG_C aspect here.

libpam/include/security/_pam_macros.h
stoeckmann: Yes, SCOPE sounds better here. Adjusted.

ldv-alt: I still suggest not to include `void` into `PAM_DEBUG_SCOPE`, let it be just `static` or nothing.

libpam_internal/pam_debug.c
```diff
@@ -0,0 +1,22 @@
...
+
+#ifdef PAM_DEBUG
+
+#define PAM_DEBUG_C 1
+
+#include "security/_pam_macros.h"
+
+#undef PAM_MACROS_H
+#undef PAM_NO_HEADER_FUNCTIONS
+#include "security/_pam_macros.h"
```
ldv-alt, line 16: Let's move `#define PAM_DEBUG_C 1` after `#undef PAM_NO_HEADER_FUNCTIONS` for better readability.

stoeckmann: Adjusted

libpam/include/security/_pam_macros.h
stoeckmann: And adjusted as well

ldv-alt (approved):

PR #756, opened 2024-01-23
doc: fix hrefs

Nielotz:
Hi, I have no idea how should i build and open docs, but if I simply **make** it, the main [index.html](https://github.com/linux-pam/linux-pam/blob/master/doc/index.html) hrefs to wrong location.

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -12,9 +12,9 @@
       currently not complete.
     <p>
     <ul>
-      <li> <a href="Linux-PAM_SAG.html">The System Administrators' Guide</a>
-      <li> <a href="Linux-PAM_MWG.html">The Module Writers' Guide</a>
-      <li> <a href="Linux-PAM_ADG.html">The Application Developers' Guide</a>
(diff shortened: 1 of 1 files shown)


Nielotz:
However It may be lack of libs on Manjaro

stoeckmann:
The documentation is built when all requirements are met. These should be:
- docbook-xml (or called docbook5-xml)
- docbook-xsl
- libxml2
- libxslt
- w3m or links as text browsers

Exact information can be retrieved here: https://github.com/linux-pam/linux-pam/blob/master/configure.ac#L689

If you compile the documentation this way, the index.html and all these other files end up in the same directory, e.g. `/usr/share/doc/Linux-PAM` if `./configure` is called without further arguments.

So, the end result is actually correct.

Nielotz:
Thank you for explanation how it works. 
I am using Linux occasionally (however currently interaction with linux-pam is a part of my thesis), and I haven't heard where docs are until now. Maybe it would be worth considering, to put some information into README file?

Issue #747, opened 2024-01-19
pam 1.6.0-3 breaks openvpn login


tomasmez:
Hello, i posted the following bug at the Archlinux bug tracking system, but it appears to be an upstream bug:
https://gitlab.archlinux.org/archlinux/packaging/packages/pam/-/issues/3#note_157561

Pasting the description here for consistency:

> sudo: reports wrong password
> 
> openvpn: the error in the openvpn-connect log is:
> 
> ⏎[Jan 19, 2024, 10:31:24] SSL Handshake: peer certificate: CN=emmet, 2048 bit RSA, cipher: TLS_AES_256_GCM_SHA384 TLSv1.3 Kx=any Au=any Enc=AESGCM(256) Mac=AEAD
> ⏎[Jan 19, 2024, 10:31:24] Session is ACTIVE ⏎[Jan 19, 2024, 10:31:24] EVENT: GET_CONFIG
> ⏎[Jan 19, 2024, 10:31:24] Sending PUSH_REQUEST to server...
> ⏎[Jan 19, 2024, 10:31:24] AUTH_FAILED
> ⏎[Jan 19, 2024, 10:31:24] EVENT: AUTH_FAILED
> ⏎[Jan 19, 2024, 10:31:24] EVENT: DISCONNECTED ⏎[Jan 19, 2024, 10:36:37]
> OpenVPN core 3.git::d3f8b18b win x86_64 64-bit built on Feb 7 2023 16:08:10
> 
> Downgrading fixed both issues

I did a git-bisect between v1.5.3 and v1.6.0 and found that b3020da7da384d769f27a8713257fbe1001878be  is the first offending commit.

As a side note: i have another install in which sudo was not affected (openvpn is not installed there).

Tomas

tomasmez:
did more testing. Openvpn is being started via a systemd service.

when exectuted as root. the problem is not present.

i suspect systemd restrictions in place affect how openvpn access pam.

could this be my problem?

stoeckmann:
Looking at the downstream bug report and later comments, `sudo` seems to be actually unaffected by these changes, but merely has an issue because `pam_faillock.so` properly works when it gets triggered by repeated failed authentication attempts by OpenVPN. Is this correct?

[...]

PR #757, opened 2024-01-24
pam_unix: try to set uid to 0 for unix_chkpwd

stoeckmann:
The geteuid check does not cover all cases. If a program runs with elevated capabilities like CAP_SETUID then we can still check credentials of other users.

Keep logging for future analysis though.

Resolves: https://github.com/linux-pam/linux-pam/issues/747
Fixes: b3020da7da38 ("pam_unix/passverify: always run the helper to obtain shadow password file entries")

### DIFF
--- a/modules/pam_unix/pam_unix_acct.c
+++ b/modules/pam_unix/pam_unix_acct.c
@@ -110,14 +110,15 @@ int _unix_run_verify_binary(pam_handle_t *pamh, unsigned long long ctrl,
       _exit(PAM_AUTHINFO_UNAVAIL);
     }
 
-    if (geteuid() == 0) {
-      /* must set the real uid to 0 so the helper will not error
-         out if pam is called from setuid binary (su, sudo...) */
-      if (setuid(0) == -1) {
-          pam_syslog(pamh, LOG_ERR, "setuid failed: %m");
-          printf("-1\n");
-          fflush(stdout);
-          _exit(PAM_AUTHINFO_UNAVAIL);
+    /* must set the real uid to 0 so the helper will not error
+       out if pam is called from setuid binary (su, sudo...) */
+    if (setuid(0) == -1) {
+      uid_t euid = geteuid();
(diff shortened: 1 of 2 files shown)


stoeckmann:
Just in case we want to support capabilities.

modules/pam_unix/pam_unix_acct.c
```diff
@@ -110,15 +110,10 @@ int _unix_run_verify_binary(pam_handle_t *pamh, unsigned long long ctrl,
...
-          pam_syslog(pamh, LOG_ERR, "setuid failed: %m");
-          printf("-1\n");
-          fflush(stdout);
-          _exit(PAM_AUTHINFO_UNAVAIL);
-      }
+    /* must set the real uid to 0 so the helper will not error
+       out if pam is called from setuid binary (su, sudo...) */
+    if (setuid(0) == -1) {
+        pam_syslog(pamh, LOG_DEBUG, "setuid failed: %m");
     }
```
ldv-alt, line 117: Shouldn't it still exit with an error if `geteuid() == 0`?

stoeckmann: I'd say that makes sense.

ldv-alt: OK, the next thing I'd suggest is to use `LOG_DEBUG` only for `geteuid() != 0` case. Let's keep `LOG_ERR` in case of exiting because it's indeed an error, and there is a long standing practice to log it this way.

stoeckmann: Yeah, let's keep old behavior for this case. Although I'm curious how to trigger it.

ldv-alt: > Yeah, let's keep old behavior for this case. Although I'm curious how to trigger it.

A seccomp filter that disallows it?

stoeckmann: Interesting idea! :)

ldv-alt (approved):

PR #759, opened 2024-01-31
libpam_internal: introduce pam_line

stoeckmann:
The pam_assemble_line function is renamed to pam_line_assemble and moved into libpam_internal so it can be shared across libpam and the pam_env module.

Applied renaming to all other relevant functions and data structures so it is easier to locate them in files.


stoeckmann:
Should have no functional change. If function changes in any way, it's a bug.

ldv-alt (approved):

PR #760, opened 2024-02-02
pam_limits: trim spaces in limit values

dimich-dmb:


modules/pam_limits/pam_limits.c
```diff
@@ -844,7 +844,7 @@ split(char *line, char **domain, char **ltype, char **item, char **value)
     *domain = strtok_r(line, " \t", &saveptr);
     *ltype = strtok_r(NULL, " \t", &saveptr);
     *item = strtok_r(NULL, " \t", &saveptr);
-    *value = strtok_r(NULL, "", &saveptr);
+    *value = strtok_r(NULL, " \t", &saveptr);
```
stoeckmann, line 847: See https://github.com/linux-pam/linux-pam/pull/762.

This change would allow values like `unlimited for everyone` because it would return `unlimited`, ignoring everything behind the next whitespace.

It's better to fail on such invalid values as well.

stoeckmann:
Hi @dimich-dmb,

I like the additional validation of number parsing. Can you add that one in another PR or open this PR again?

Issue #761, opened 2024-02-02
pam_limits: literal values aren't recognized in limits.conf


dimich-dmb:
Values "-", "unlimited", "infinity" aren't recognized as `-1`.
`/etc/security/limits.conf`:
```
*                hard    core            unlimited
*                soft    core            0
```
journal:
```
pam_limits(sudo:session): wrong limit value '           unlimited' for limit type 'hard'
```
linux-pam 1.6.0

This pull request seems to fix the issue: https://github.com/linux-pam/linux-pam/pull/760

stoeckmann:
Confirmed. This is a regression introduced with 1.6.0.

I have created PR https://github.com/linux-pam/linux-pam/pull/762 which verifies correctness of values by still considering everything after the third field to be a value, i.e. considering `unlimited for everyone` as a single value instead of truncating it to `unlimited`.

dimich-dmb:
> considering `unlimited for everyone` as a single value instead of truncating it to `unlimited`.

Note another (tiny) issue: e.g. value `37somegarbage` will be converted to `37` by `strtoull()` with no error/warning.
I had to not squash two commits into single PR, my bad. Probably need to create another issue.

stoeckmann:
I've created a lot of PRs without issues in the past, so just a PR with the commit you've already prepared would work totally fine. But it's up to you. If you want to add an issue as well, it won't hurt either.

PR #762, opened 2024-02-02
pam_limits: fix sscanf removal regression

stoeckmann:
- Remove whitespaces around value in limits.conf
- Keep copy of input line for error messages

Fixes https://github.com/linux-pam/linux-pam/issues/761.

### DIFF
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -882,6 +882,7 @@ parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
     FILE *fil;
     char *buf = NULL;
     size_t n = 0;
+    unsigned long long lineno = 0;
 
     /* check for the conf_file */
     if (ctrl & PAM_DEBUG_ARG)
@@ -904,6 +905,8 @@ parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
         int rngtype;
         size_t j;
         uid_t min_uid = (uid_t)-1, max_uid = (uid_t)-1;
+
+        lineno++;
 
         line = buf;
         /* skip the leading white space */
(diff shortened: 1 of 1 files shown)


modules/pam_limits/pam_limits.c
```diff
@@ -832,6 +832,23 @@ set_if_null(char **dest, char *def)
     return 1;
 }
 
+static char *
+trim(char *s)
+{
+    char *p;
+
+    while (*s == ' ' || *s == '\t')
```
ldv-alt, line 840: Since `s` is a value returned by `strtok_r`, it can be `NULL`.

modules/pam_limits/pam_limits.c
```diff
@@ -886,16 +903,20 @@ parse_config_file(pam_handle_t *pamh, const char *uname, uid_t uid, gid_t gid,
...
+        tptr = strchr(line,'\n');
+        if (tptr)
+            *tptr = '\0';
+
+        _pam_drop(orig);
+        orig = strdup(line);
+
         /* skip the leading white space */
         while (*line && isspace((unsigned char)*line))
             line++;
```
ldv-alt, line 916: Is it really useful to include the leading white space in `orig`?

ldv-alt (changes requested):

modules/pam_limits/pam_limits.c
stoeckmann: Thanks!

modules/pam_limits/pam_limits.c
stoeckmann: I'd keep the original line as it is so an administrator knows exactly which line is affected and that PAM did not misread it in any way. Alternatively, it might be sufficient to print the filename and line number in the log.

ldv-alt: I agree, the filename and line number seem to be a better choice.

stoeckmann: Okay, I've adjusted the message.

ldv-alt (approved):

PR #764, opened 2024-02-04
pam_limits: reject malformed numeric values

dimich-dmb:
`endptr` pointing to the end of string ensures no extra characters follow digits in a number. For example, value "1024k" will be rejected.

Check for minus sign at the beginning of string ensures no negative value supplied. For example, "-2" will be rejected instead of being converted to unsigned representation.
Special values "-1" and "-" are handled earlier in another clause.

No checks for ULLONG_MAX result (ULONG_MAX for 32-bit file offsets) and ERANGE error performed. Valid numeric values greater than ULLONG_MAX, for example, "99999999999999999999", are converted to ULLONG_MAX and will be interpeted as RLIM_INFINITY by subsequent setrlimit().

`value_orig` removed as unused. `lim_value` isn't being modified anywhere.

### DIFF
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -550,7 +550,6 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
     int int_value = 0;
     rlim_t rlimit_value = 0;
     char *endptr;
-    const char *value_orig = lim_value;
 
     if (ctrl & PAM_DEBUG_ARG)
 	 pam_syslog(pamh, LOG_DEBUG, "%s: processing %s %s %s for %s",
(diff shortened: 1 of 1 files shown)


dimich-dmb:
This PR makes sence only after https://github.com/linux-pam/linux-pam/pull/762 merged.

modules/pam_limits/pam_limits.c
```diff
@@ -550,7 +550,6 @@ process_limit (const pam_handle_t *pamh, int source, const char *lim_type,
     int int_value = 0;
     rlim_t rlimit_value = 0;
     char *endptr;
-    const char *value_orig = lim_value;
```
ldv-alt, line 553: Could you move the cleanup part of this change into a separate commit, please?

ldv-alt (changes requested):

modules/pam_limits/pam_limits.c
dimich-dmb: I don't see the point to split 3-lines commit into two: neither first to remove unnecessary variable but leave buggy line, nor first to fix buggy line with unnecessary variable and second to remove it.

ldv-alt: The point is that unrelated changes that address different issues should be committed separately,

PR #773, opened 2024-02-29
pam_start: Include <libintl.h> for bindtextdomain

marv:
Fixes the following build error (RISCV64 platform using musl-1.2.4 libc with LLVM toolchain):
```
pam_start.c:39:2: error: call to undeclared function 'bindtextdomain'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   39 |         bindtextdomain(PACKAGE, LOCALEDIR);
      |         ^
1 error generated.
```

### DIFF
--- a/libpam/pam_start.c
+++ b/libpam/pam_start.c
@@ -15,6 +15,10 @@
 #include <string.h>
 #include <syslog.h>
 
+#ifdef HAVE_BINDTEXTDOMAIN
+#include <libintl.h>
+#endif
+
 static int _pam_start_internal (
     const char *service_name,
     const char *user,


ldv-alt:
Do you configure the build using --disable-nls?  If not, then libintl.h is included by config.h already.

What you probably need is the following change instead:
```
--- a/libpam/pam_start.c
+++ b/libpam/pam_start.c
@@ -25,7 +25,7 @@ static int _pam_start_internal (
     D(("called pam_start: [%s] [%s] [%p] [%p]"
        ,service_name, user, pam_conversation, pamh));
 
-#ifdef HAVE_BINDTEXTDOMAIN
+#if defined ENABLE_NLS && defined HAVE_BINDTEXTDOMAIN
     /* Bind text domain to pull in PAM translations for a case where
        linux-pam is installed to non-default prefix.
 
```

PR #780, opened 2024-03-22
pam_get_data: manual page claims that PAM_NO_MODULE_DATA is returned in case of NULL data value

3v1n0:
The manual for pam_get_data(3) states:
   Module data not found or there is an entry, but it has the value `NULL`.

However, if the data is set and then unset, no error is returned.

So when checking for the value, actually ensure that it's not `NULL`.

---

The root issue for this is that `pam_set_data` does not really ever unsets the data, so ~~I'll suggest another alternative in a new PR~~ (done in #781), but this is the quicker implementation.

As per this, add tests since we had no get tests.

Also fix null pointer access when getting the data without a pointer receiver, I think it can be useful in some cases and I don't think we should error in such case, but if that's not wanted, we could just return `PAM_SYSTEM_ERROR`

### DIFF
--- a/doc/man/pam_get_data.3.xml
+++ b/doc/man/pam_get_data.3.xml
@@ -79,8 +79,7 @@
         <term>PAM_NO_MODULE_DATA</term>
         <listitem>
            <para>
-              Module data not found or there is an entry, but it has
-              the value NULL.
+              No module specific data is present.
           </para>
         </listitem>
       </varlistentry>
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -12,7 +12,7 @@ CLEANFILES = *~
 TESTS = tst-pam_start tst-pam_end tst-pam_fail_delay tst-pam_open_session \
 	tst-pam_close_session tst-pam_acct_mgmt tst-pam_authenticate \
 	tst-pam_chauthtok tst-pam_setcred tst-pam_get_item tst-pam_set_item \
-	tst-pam_getenvlist tst-pam_get_user tst-pam_set_data \
+	tst-pam_getenvlist tst-pam_get_user tst-pam_get_data tst-pam_set_data \
 	tst-pam_mkargv tst-pam_start_confdir
 
 EXTRA_DIST = confdir
--- a/tests/tst-pam_get_data.c
+++ b/tests/tst-pam_get_data.c
@@ -1,0 +1,180 @@
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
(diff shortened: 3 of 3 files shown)


ldv-alt:
Given that the current behavior of returning `PAM_NO_MODULE_DATA` only when the module data was not found can be tracked back to the very first git commit ea488580, and the manual page was introduced much later by commit f7827ccfd,
let's consider the current behavior as correct and fix the manual page instead.

Also, the current behavior is more versatile as it allows to differentiate these two conditions (no data entry vs data entry with NULL value).

3v1n0:
> Also, the current behavior is more versatile as it allows to differentiate these two conditions (no data entry vs data entry with NULL value).

Not fully think this is true, since setting `NULL` data everywhere should mean that there's no data, and a module could use any invalid pointer value as `NULL` item instead, if one wants to explicitly mark this difference. While the fact of not returning an error when returning a `NULL` pointer seems to me dangerous as it opens more gates to a potential null-pointer deference.

However, if you really think having the current behavior is better, I can adjust the docs and tests.

ldv-alt:
We can wait for others to comment. 
@thkukuk, do you by any chance remember why the manual page you added back in 2006 doesn't quite match the implementation?

thkukuk:
> We can wait for others to comment. @thkukuk, do you by any chance remember why the manual page you added back in 2006 doesn't quite match the implementation?

No,  I don't remember. But the implementation, headers and pam.3 say "no module specific data is present", nothing about data == NULL.
So I would say that pam_get_data.3 is wrong. Most likely I didn't adjust the template correct for this value when creating the new manual pages.
Since we document it everywhere else as the implementation behaves, I would fix the manual page. 

I know from many code where it is a big difference if a value is NULL or not set. I don't know of pam_get_data() usage, but I would not an incompatibility here after so many decades by changing the behavior.

Additional, I checked openpam, they also don't check if the data is NULL. 

So I suggest to fix the manual page (use the wording from the header file or pam.3), but don't make the code incompatible to other PAM implementations and break compatibility with the last decades.

ldv-alt (approved):

doc/man/pam_get_data.3.xml
3v1n0, line 1: I'd stress in the description that data can be `NULL`.

3v1n0: And maybe in the set-data doc too.

ldv-alt:
> @ldv-alt maybe also [2efd736](https://github.com/linux-pam/linux-pam/commit/2efd736b06fc0b8586b3ce2764a5f69566300dec) can be preserved or want a new PR for that?

Given that `const void **data` argument has never been allowed to be NULL, allowing it now would be useless: modules wouldn't be able to pass NULL anyway because they are not guaranteed to be using the latest pam.

Issue #758, opened 2024-01-29
pam_unix/passverify.c does not allow empty passwords with non-empty hashes


severach:
I have a user with a password length of 0 crypt and set in /etc/shadow. The new code b3020da7da384d769f27a8713257fbe1001878be in `modules/pam_unix/passverify.c` does not allow this user to login.

The hash was generated with this code.

```
#!/usr/bin/perl

use strict;
use warnings;
use integer;
no overloading;

# 2021-06-13 PAM disabled 3DES MD5 and SHA256 passwords.
# The famous U6aMy0wojraho no longer works.
# Most password utilities will not generate a hash for a blank password and blank salt.
# This generates a high security SHA512 blank password for passwordless logons.
# -CJS

print crypt("", '$6$$'); # U6=3DES with well known salt, $1$$=MD5, $5$$=SHA256, $6$$=SHA512
```

This patch fixes the problem though it still isn't the same as the the code in 1.53.

```
diff -ru5 a/modules/pam_unix/passverify.c b/modules/pam_unix/passverify.c
--- a/modules/pam_unix/passverify.c 2024-01-17 05:29:36.000000000 -0500
+++ b/modules/pam_unix/passverify.c 2024-01-29 13:46:34.277941821 -0500
@@ -236,17 +236,17 @@
 #else
      /* we must run helper for NIS+ passwords */
      return PAM_UNIX_RUN_HELPER;
 #endif
    } else if (is_pwd_shadowed(*pwd)) {
-#ifdef HELPER_COMPILE
      /*
       * shadow password file entry for this user,
       * if shadowing is enabled
       */
      *spwdent = getspnam(name);
      if (*spwdent == NULL || (*spwdent)->sp_pwdp == NULL)
+#ifdef HELPER_COMPILE
        return PAM_AUTHINFO_UNAVAIL;
 #else
      /*
       * The helper has to be invoked to deal with
       * the shadow password file entry.
```

ldv-alt:
I don't see how it is related to #686.

ldv-alt:
I suppose the problem is elsewhere.

severach:
I don't know what the helper is but I suspect the problem is that the helper does not approve blank passwords. This code does approve blank passwords through a comedy of errors.

ldv-alt:
The helper behavior with regards to blank passwords is controlled by `nullok` option.  The same applied to `pam_unix` as a whole.

severach:
[...]

PR #784, opened 2024-03-28
pam_unix: allow empty passwords with non-empty hashes

trofi:
Before the change pam_unix has different behaviours for a user with empty password for these two `/etc/shadow` entries:

    nulloktest:$6$Yy4ty2jJ$bsVQWo8qlXC6UHq1/qTC3UR60ZJKmKApJ3Wj7DreAy8FxlVKtlDnplFQ7jMLVlDqordE7e4t49GvTb.aI59TP0:1::::::
    nulloktest::1::::::

The entry with a hash was rejected and the entry without was accepted.

The rejection happened because 9e74e90147c "pam_unix: avoid determining if user exists" introduced the following rejection check (slightly simplified):

        ...
        } else if (p[0] == '\0' && nullok) {
                if (hash[0] != '\0') {
                        retval = PAM_AUTH_ERR;
                }

We should not reject the user with a hash assuming it's non-empty. The change does that by pushing empty password check into `verify_pwd_hash()`.

`NixOS` generates such hashed entries for empty passwords as if they were non-empty using the following perl code:

    sub hashPassword {
        my ($password) = @_;
        my $salt = "";
        my @chars = ('.', '/', 0..9, 'A'..'Z', 'a'..'z');
        $salt .= $chars[rand 64] for (1..8);
        return crypt($password, '$6$' . $salt . '$');
    }

Resolves: https://github.com/linux-pam/linux-pam/issues/758
Fixes: 9e74e90147c "pam_unix: avoid determining if user exists"

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -76,9 +76,13 @@ PAMH_ARG_DECL(int verify_pwd_hash,
 
 	strip_hpux_aging(hash);
 	hash_len = strlen(hash);
-	if (!hash_len) {
+
+	if (p && p[0] == '\0' && !nullok) {
+		/* The passed password is empty */
+		retval = PAM_AUTH_ERR;
+	} else if (!hash_len) {
 		/* the stored password is NULL */
-		if (nullok) { /* this means we've succeeded */
+		if (p && p[0] == '\0' && nullok) { /* this means we've succeeded */
 			D(("user has empty password - access granted"));
 			retval = PAM_SUCCESS;
 		} else {
@@ -1109,12 +1113,6 @@ helper_verify_password(const char *name, const char *p, int nullok)
 	if (pwd == NULL || hash == NULL) {
 		helper_log_err(LOG_NOTICE, "check pass; user unknown");
 		retval = PAM_USER_UNKNOWN;
-	} else if (p[0] == '\0' && nullok) {
-		if (hash[0] == '\0') {
-			retval = PAM_SUCCESS;
-		} else {
-			retval = PAM_AUTH_ERR;
-		}
 	} else {
 		retval = verify_pwd_hash(p, hash, nullok);
 	}


modules/pam_unix/passverify.c
```diff
@@ -76,9 +76,15 @@ PAMH_ARG_DECL(int verify_pwd_hash,
 
 	strip_hpux_aging(hash);
 	hash_len = strlen(hash);
+
+	/* The passed password is empty */
+	if (p && p[0] == '\0' && !nullok) {
+		retval = PAM_AUTH_ERR;
+	}
```
ldv-alt, line 83: Hmm, this assignment doesn't seem to have any effect as `retval` is getting assigned later in this function anyway.

trofi: Great catch! Forgot to test the `nullok == 0` case and missed it. Updated it to reside in `if {} else {}` branches.

Tested the cases with:

```c
               // empty request and entry, correct password:
               assert(verify_pwd_hash("", "", 1) == PAM_SUCCESS);
               assert(verify_pwd_hash("", "", 0) != PAM_SUCCESS);
               assert(verify_pwd_hash("", "$6$Yy4ty2jJ$bsVQWo8qlXC6UHq1/qTC3UR60ZJKmKApJ3Wj7DreAy8FxlVKtlDnplFQ7jMLVlDqordE7e4t49GvTb.aI59TP0", 1) == PAM_SUCCESS);
               // this case:
               assert(verify_pwd_hash("", "$6$Yy4ty2jJ$bsVQWo8qlXC6UHq1/qTC3UR60ZJKmKApJ3Wj7DreAy8FxlVKtlDnplFQ7jMLVlDqordE7e4t49GvTb.aI59TP0", 0) != PAM_SUCCESS);

               // empty request and entry, incorrect password:
               assert(verify_pwd_hash("1", "", 1) != PAM_SUCCESS);
               assert(verify_pwd_hash("1", "", 0) != PAM_SUCCESS);
               assert(verify_pwd_hash("1", "$6$Yy4ty2jJ$bsVQWo8qlXC6UHq1/qTC3UR60ZJKmKApJ3Wj7DreAy8FxlVKtlDnplFQ7jMLVlDqordE7e4t49GvTb.aI59TP0", 1) != PAM_SUCCESS);
               assert(verify_pwd_hash("1", "$6$Yy4ty2jJ$bsVQWo8qlXC6UHq1/qTC3UR60ZJKmKApJ3Wj7DreAy8FxlVKtlDnplFQ7jMLVlDqordE7e4t49GvTb.aI59TP0", 0) != PAM_SUCCESS);

               // non-empty, correct password:
               assert(verify_pwd_hash("test", "$y$j9T$BCdF/dSFuBgDI2aBHUCE3/$DsEklvSUTOp/sG2vtt0PHD1tKzHDIYClPO53Exd1vu1", 1) == PAM_SUCCESS);
               assert(verify_pwd_hash("test", "$y$j9T$BCdF/dSFuBgDI2aBHUCE3/$DsEklvSUTOp/sG2vtt0PHD1tKzHDIYClPO53Exd1vu1", 0) == PAM_SUCCESS);
               // non-empty, incorrect password:
               assert(verify_pwd_hash("test2", "$y$j9T$BCdF/dSFuBgDI2aBHUCE3/$DsEklvSUTOp/sG2vtt0PHD1tKzHDIYClPO53Exd1vu1", 1) != PAM_SUCCESS);
               assert(verify_pwd_hash("test2", "$y$j9T$BCdF/dSFuBgDI2aBHUCE3/$DsEklvSUTOp/sG2vtt0PHD1tKzHDIYClPO53Exd1vu1", 0) != PAM_SUCCESS);
```

The case 4 was broken, works now.

ldv-alt (approved):

ldv-alt:
> @trofi @ldv-alt I noticed that this change causes a delay for login(1) right after entering the user name.

If this happens, it's a regression from 9e74e90147c530801e3ea3428d64371722c90e01 that was made to address information leaks.

trofi:
> @trofi @ldv-alt I noticed that this change causes a delay for login(1) right after entering the user name. So with this change, we now essentially leak information that a specific user exists on the machine: a nonexistent user will continue to the password prompt immediately, whereas any user that has a password hash set will cause a delay while we check against the empty password.

I was not able to reproduce it locally. Do you use any special settings in `/etc/pam.d/login`?

Mine are:

```
# $ cat /etc/pam.d/login |& unnix
cat /etc/pam.d/login | unnix
# Account management.
account required pam_unix.so # unix (order 10900)

# Authentication management.
auth sufficient pam_unix.so likeauth nullok try_first_pass # unix (order 11500)
auth required pam_deny.so # deny (order 12300)

# Password management.
password sufficient pam_unix.so nullok yescrypt # unix (order 10200)

# Session management.
session required pam_env.so conffile=/etc/pam/environment readenv=0 # env (order 10100)
session required pam_unix.so # unix (order 10200)
session required pam_loginuid.so # loginuid (order 10300)
session required /<<NIX>>/linux-pam-1.6.1/lib/security/pam_lastlog.so silent # lastlog (order 10700)
session optional /<<NIX>>/systemd-255.4/lib/security/pam_systemd.so # systemd (order 12000)
```

How I tested:

```
$ sudo su -
# login

nz login: doesnotexist
Password:
(instant "Password:", delayed result on invalid password, that user does not exist)

nz login: root
Password:
(instant "Password:", delayed result on invalid password, that user exists, allwoed password)
```

I see the same results on both `pam-1.6.1` and `pam-1.6.0` with 8d0c575336ad301cd14e16ad2fdec6fe621764b8 applied on top.

pks-t:
On Tue, Apr 30, 2024 at 03:07:36PM -0700, Sergei Trofimovich wrote:
> > @trofi @ldv-alt I noticed that this change causes a delay for
> > login(1) right after entering the user name. So with this change, we
> > now essentially leak information that a specific user exists on the
> > machine: a nonexistent user will continue to the password prompt
> > immediately, whereas any user that has a password hash set will
> > cause a delay while we check against the empty password.
> 
> I was not able to reproduce it locally. Do you use any special settings in `/etc/pam.d/login`?

No, I have the exact same configuration as you do on nixos-unstable.
I pinned down the regression into the range between

    66adc1e47f87 (Merge pull request #304601 from andresilva/firefox-125.0.1, 2024-04-16)

and

    5c24cf2f0a12 (Merge pull request #303600 from r-ryantm/auto-update/python311Packages.mailchecker, 2024-04-19)

This range contains e3201828048c (Merge pull request #298994 from
trofi/linux-pam-revert-helper, 2024-03-29), which backports this pull
request here.

I didn't pin it down exactly to that commit because it causes almost a
world rebuild of my system when I go further into that range. In any
case, the commit range does not contain any other changes that would be
related to authentication, as far as I can see.

> How I tested:
> 
> ```
> $ sudo su -
> # login
> 
> nz login: doesnotexist
> Password:
> (instant "Password:", delayed result on invalid password, that user does not exist)
> 
> nz login: root
> Password:
> (instant "Password:", delayed result on invalid password, that user exists, allwoed password)
> ```

Yeah, this is how I can reproduce the issue. What may be of interest is
that I use yescrypt with a weight of 11 (mkpasswd -m yescrypt -R 11).
This does take a few seconds to decrypt and thus makes the delay after
entering the username quite noticeable.

trofi:
Aha, that makes sense. Reproduced locally. Another (slightly more user visible test) is to set an empty `""` hashed password on a user. I think `login` is expected to ask for a password prompt here, but it does not and logs the user in after a `Login:` prompt. Will double-check against `pam-1.5.3`. Looking.

dchmelik:
Neither should be rejected; they're not on FreeBSD UNIX (a genetic UNIX) and doing so is non-UNIX-like behaviour forcing users & system administrators (sysadmins) to use things the way they might dislike, including computers in secure areas or without important data, and with chroots & virtual machines on which password is simply overkill.  Should I open a new issue about this or would you consider classic UNIX (originally in laboratories people could use without password, and even some users did) standards and update the issue?

PR #787, opened 2024-04-11
Support elogind as logind alternative

Polynomial-C:
Can be enabled with --enable-logind=elogind

logind from systemd is still the default

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -556,17 +556,19 @@ elif test "$WITH_SELINUX" = "yes" ; then
 fi
 
 LOGIND_CFLAGS=
-SYSTEMD_LIBS=
+LOGIND_LIBS=
 AC_ARG_ENABLE([logind],
   AS_HELP_STRING([--disable-logind], [Disable logind support]),
   [WITH_LOGIND=$enableval], [WITH_LOGIND=check])
-if test "$WITH_LOGIND" = "check"; then
-  PKG_CHECK_MODULES([SYSTEMD], [libsystemd >= 254], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $SYSTEMD_CFLAGS"], [:])
-elif test "$WITH_LOGIND" = "yes"; then
-  PKG_CHECK_MODULES([SYSTEMD], [libsystemd >= 254], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $SYSTEMD_CFLAGS"], [])
-fi
+AS_IF([test "$WITH_LOGIND" != "no"], [
+  AS_CASE(["$enableval"], [elogind], [
+    PKG_CHECK_MODULES([LOGIND], [libelogind >= 255], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $LOGIND_CFLAGS"])
+  ], [
+    PKG_CHECK_MODULES([LOGIND], [libsystemd >= 254], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $LOGIND_CFLAGS"], [AS_IF([test "$WITH_LOGIND" != "check"], [AC_MSG_ERROR([Cannot find libsystemd])])])
+  ])
+])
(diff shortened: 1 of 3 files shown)


Polynomial-C:
You can find the elogind project here: https://github.com/elogind/elogind

It's a drop-in replacement for systemd-logind as it's based on the systemd sources.

thkukuk:
Basically the patch is ok, but I expect that Linux-PAM will use more systemd functions in the future, even if I don't have anything concrete in mind at the moment.

That's why I would prefer SYSTEMD_LIBS and EGLOGIND_LIBS to be two independent variables from the beginning, so that we don't have to separate them later again.

@ldv-alt what do you think?

configure.ac
```diff
@@ -556,17 +556,19 @@ elif test "$WITH_SELINUX" = "yes" ; then
...
 AC_ARG_ENABLE([logind],
   AS_HELP_STRING([--disable-logind], [Disable logind support]),
   [WITH_LOGIND=$enableval], [WITH_LOGIND=check])
-if test "$WITH_LOGIND" = "check"; then
-  PKG_CHECK_MODULES([SYSTEMD], [libsystemd >= 254], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $SYSTEMD_CFLAGS"], [:])
-elif test "$WITH_LOGIND" = "yes"; then
-  PKG_CHECK_MODULES([SYSTEMD], [libsystemd >= 254], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $SYSTEMD_CFLAGS"], [])
-fi
+AS_IF([test "$WITH_LOGIND" != "no"], [
+  AS_CASE(["$enableval"], [elogind], [
```
ldv-alt, line 564: Wouldn't it be better to use `AS_CASE(["$WITH_LOGIND"], ...)` instead of `AS_IF(...)` and `AS_CASE(["$enableval"], ...)`?

configure.ac
```diff
@@ -556,17 +556,19 @@ elif test "$WITH_SELINUX" = "yes" ; then
...
+    PKG_CHECK_MODULES([LOGIND], [libelogind >= 255], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $LOGIND_CFLAGS"])
+  ], [
+    PKG_CHECK_MODULES([LOGIND], [libsystemd >= 254], [LOGIND_CFLAGS="-DUSE_LOGIND=1 $LOGIND_CFLAGS"], [AS_IF([test "$WITH_LOGIND" != "check"], [AC_MSG_ERROR([Cannot find libsystemd])])])
```
ldv-alt, line 567: When using `AS_CASE`, it would be more readable to handle each case separately rather than  merging `yes` and `check` cases.

ldv-alt (changes requested):

ldv-alt:
> Basically the patch is ok, but I expect that Linux-PAM will use more systemd functions in the future, even if I don't have anything concrete in mind at the moment.
> 
> That's why I would prefer SYSTEMD_LIBS and EGLOGIND_LIBS to be two independent variables from the beginning, so that we don't have to separate them later again.
> 
> @ldv-alt what do you think?

We could have SYSTEMD_* and ELOGIND_* variables initialized in configure.ac, and LOGIND_* variables based on them that are AC_SUBST'ed and would currently be used in Makefile.am files.  This way we could make SYSTEMD_* variables AC_SUBST'ed when needed.

thkukuk:
> However, why would PAM use (any) systemd (functions), as systemd-logind uses PAM, and has several PAM modules. One of them, pam_systemd, is distributed as pam_elogind by us.

If PAM would not use systemd functions, there would be no need for you to add elogind support.
One year ago PAM didn't used any libsystemd function and nobody could imagine that we ever would. Now we are using libsystemd functions to replace `utmp` with systemd-logind. 
Nobody knows what will come tomorrow, and which functions more we need/want to use.

ldv-alt:
Superseded by #859.

PR #789, opened 2024-04-21
Set default fail delay to 0.5 seconds

### LINKED ISSUE #778: Fix delay UX
PAM's `pam_unix` has a default fixed delay of approximately 2 seconds. As far as I can tell most software and distros stick with this and use it for everything - ssh, sudo, login, etc.

From a usability point of view that sucks. It's a significant amount of time to wait when you make a typo and it add dubious security benefits. There are no security benefits at all for sudo (you can just use `unix_chkpwd`), and I'm dubious about `sshd` - I expect you can get around it just by creating a load of parallel connections (though I haven't tried it).

I would like to fix this annoying paper cut. I can think of the following solutions and wanted your feedback on what you would accept: 

1. Just change the default delay to 0.5s. This is short enough to be not annoying to users making typos but isn't really going to change any security properties (nobody is going to care about waiting 4x longer to brute force a login). This also has the advantage of being really really easy to do.
2. Add a parameter to `faildelay` to have increasing delays, and store the current delay count as part of the session (if that is possible; I couldn't quite figure it out and it seems like a bad idea to use global variables). This isn't much work either.
3. Add a parameter to `faildelay` to have increasing delays, and persist the fail information to disk in the same way that `faillock` does. This is a ton of work. I'll probably give up if this is the only acceptable solution.
4. Add a parameter to `faillock` to have increasing delays.
5. Add a way to *decrease* the delay. The issue with API at the moment is that modules in the PAM stack can only increase the delay. The way Debian has the configs set up, there's a default config with the 2s delay, which means you can't e.g. force it back to 0 for `sudo`. It's not ideal to have the default delay as 0 because that isn't fail-safe. Every config *except* `sudo` then has to remember to set the delay.

My preference is 1. What do you think?

Timmmm:
The default fail delay is 2 seconds (+/- 50% randomly). This is done to prevent brute force attacks, but unfortunately because isn't applied exponentially, it also greatly degrades the UX for legitimate users. Every typo is immediately met with an annoying 2 second delay.

A full fix would be to persist the login failure times and apply an exponential delay (this is what Windows does), however this is quite complicated due to PAM's architecture (see the `faillock` plugin for an example of this).

This simple fix greatly improves the UX at very minimal security cost. Any brute force attack that can tolerate a 0.5 second delay will also work with a 2 second delay. In other words a factor of 4 is nothing from an attacker's perspective, but makes a big difference from a legitimate user's perspective.

Fixes #778

### DIFF
--- a/modules/pam_faillock/pam_faillock.c
+++ b/modules/pam_faillock/pam_faillock.c
@@ -473,7 +473,7 @@ pam_sm_authenticate(pam_handle_t *pamh, int flags,
 		goto err;
 
 	if (!(opts.flags & FAILLOCK_FLAG_NO_DELAY)) {
-		pam_fail_delay(pamh, 2000000);	/* 2 sec delay on failure */
+		pam_fail_delay(pamh, 500000);	/* 0.5 sec delay on failure */
 	}
 
 	if ((rv=get_pam_user(pamh, &opts)) != PAM_SUCCESS) {
--- a/modules/pam_unix/support.c
+++ b/modules/pam_unix/support.c
@@ -721,7 +721,7 @@ int _unix_verify_password(pam_handle_t * pamh, const char *name
 
 	if (off(UNIX_NODELAY, ctrl)) {
 		D(("setting delay"));
-		(void) pam_fail_delay(pamh, 2000000);	/* 2 sec delay for on failure */
+		(void) pam_fail_delay(pamh, 500000);	/* 0.5 sec delay for on failure */
 	}
 
 	/* locate the entry for this user */


Timmmm:
Sorry VSCode didn't want to save Dag-Erling Smørgrav's name unmodified so I just fixed it. That line is 22 years old so I guess it kind of predates UTF-8 & I couldn't figure out the intended original encoding.

BenBE:
@Timmmm You can use `git add -p`/`git reset -p` to skip parts of the changes (before commiting). Also `git gui` can be of great help to drop that unintended change (afterwards) from the commit.

Timmmm:
Thanks, I had no idea Git came with a GUI! Anyway I fixed it and made a separate PR to fix the encoding: #790

Timmmm:
I removed the erroneous change to `CHANGELOG`.

Timmmm:
Any chance someone could take a look at this?

Timmmm:
Ok, will do when it gets reviewed.

BenBE:
LGTM.

Only thing I noticed (but that's in the original source too) is that the return value of `pam_fail_delay` is ignored in both locations. And while for one location it is made explicit, that ignoring is performed, the other location silently ignores the potential error case. I'm not deep enough into the source to tell if this causes some issues, but at least that's one point of inconsistency. Fixing this is best done in a new PR.

Zero3:
@Timmmm / @BenBE I see this PR has been laying around in "LGTM" status for a year by now. I am curious as to whether there are any plans for merging this? :)

BenBE:
> @Timmmm / @BenBE I see this PR has been laying around in "LGTM" status for a year by now. I am curious as to whether there are any plans for merging this? :)

I sure hope so, that this gets merged at some point. But I'm not a maintainer here, just someone who happens to have taken a look at the PR itself.

Timmmm:
@Zero3 @BenBE I think it's unlikely unfortunately. The maintainers prefer the 2 second delay because:

> [We find the short pause is just enough to take a breath and refocus after a mistyped password, preventing a second fumbled attempt.](https://github.com/linux-pam/linux-pam/issues/778#issuecomment-3146412689)

They also said

> I would need more than anecdotal feedback. Could you please provide some compelling evidence to support the claim that a 2-second delay is widely considered a poor user experience?

And I [attempted to determine exactly what kind of evidence they would accept](https://github.com/linux-pam/linux-pam/issues/778#issuecomment-3147980871) (I don't want to put a lot of effort in only for my evidence to be dismissed on a technicality), but they stopped replying.

I wouldn't rule out this happening one day. The issue now has 21 upvotes and zero downvotes which has got to be fairly decent evidence! But it's not a democracy - if the maintainers disagree I think that's that.

I may just cut the gordian knot and set `NOPASSWD` for `sudo`. Password input for `sudo` is mostly security theatre anyway.

PR #790, opened 2024-04-24
Fix CHANGELOG encoding

Timmmm:
Dag-Erling Smørgrav's name was encoded using some encoding other than UTF-8 (that line predates the UTF-8 consensus). This changes it to use UTF-8, so that it renders properly and so that editors that guess the file is UTF-8 do not get confused about it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -296,7 +296,7 @@ bug report - outstanding bugs are listed here:
 * A Tru64 fix (given other stuff has already resolved this, it
   actually just a comment actually) from 'Eddie'. (Bug 418450 -
   agmorgan)
-* pam_handlers: BSD fix from Dag-Erling Smørgrav and Anton Berezin
+* pam_handlers: BSD fix from Dag-Erling SmÃ¸rgrav and Anton Berezin
   (Bug 486063 - agmorgan)
 * added the dynamic/* directory to the distribution. If you go in
   there after building the rest of the tree, you'll make a pam.so


ldv-alt:
Thanks, but why are you touching this file in the first place? It's obsolete for almost 20 years now.
I don't mind fixing it's encoding, but please format the commit message so that lines of text won't exceed 72 symbols.

Timmmm:
> Thanks, but why are you touching this file in the first place?

I was editing something else in this file and the non-UTF-8 character caused some issues. See https://github.com/linux-pam/linux-pam/pull/789

spaette:
> editors that guess the file

This is a VSCode issue.

It renders correctly opening in emacs and vim in my terminal.

> fixing its encoding

IMHO, that is unnecessary.

```
$ echo $LANG
da_DK.UTF-8
$ sed -n '299,300p' /tmp/linux-pam/CHANGELOG
* pam_handlers: BSD fix from Dag-Erling Sm�rgrav and Anton Berezin
  (Bug 486063 - agmorgan)
$ luit -encoding 'ISO 8859-1' sed -n '299,300p' /tmp/linux-pam/CHANGELOG
* pam_handlers: BSD fix from Dag-Erling Smørgrav and Anton Berezin
  (Bug 486063 - agmorgan)
$ 
```

Timmmm:
Not exactly. Text files have no in-band encoding information, so everything has to guess the encoding. Emacs and Vim are ancient so they guess Latin1 by default. VSCode is more recent so it guesses UTF-8, as do most modern tools. For example GitHub (check the diff here).

UTF-8 is far more common than Latin1 these days so it makes sense to use that. Also all UTF-8 is valid Latin1 but the reverse is not true.

I'm not going to lose sleep over this though. I was just trying to help.

spaette:
Perhaps, @ldv-alt a less verbose commit message would be appropriate.

_for example_

convert CHANGELOG character encoding

> Not exactly.

_cf_: vscode issue (Closed)

`#32798`

> no in-band encoding information

`https://harmful.cat-v.org/software/`

Harmful things that are so superfluous and useless that require no alternative:

- Unicode BoM in UTF-8.

> guess the encoding

_to paraphrase you_ (in quotes)

KDE's text editors "guess the encoding" as ISO 8859-15.

Timmmm:
> Harmful things that are so superfluous and useless that require no alternative: Unicode BoM in UTF-8.

True, but I don't see the relevance. A BoM only distinguishes between UTF-16 LE and BE. That's it.

> KDE's text editors "guess the encoding" as ISO 8859-15.

Good for them!

Are you in a position to merge this? If not it seems like a lot of needless debate for a tiny fix. Perhaps we should wait to see if the project maintainers are as opposed to UTF-8 as you!

spaette:
@Timmmm

I recall on Windows an editor inserting a BOM in a file upon saving. 

I don't have write authority.

Is it really a "Fix" and for what?

I addressed relevance. 

I didn't assert opposition.

ldv-alt:
> > Thanks, but why are you touching this file in the first place?
> 
> I was editing something else in this file and the non-UTF-8 character caused some issues. See #789

Sorry, I still don't understand why you decided to edit `CHANGELOG` file in the first place.
As I mentioned earlier, this file is obsolete since 2005.  In fact, it starts with the following comment:
`This file is no longer used for tracking changes for Linux-PAM`.

I don't mind converting `CHANGELOG` file to UTF-8, but it looks like this historic file should rather be moved out of the top level directory, if not removed.

spaette:
My sympathies at this point of time is to close this pull.

> moved out of the top level directory, if not removed

Doesn't automake require presence of "standard" files including the ChangeLog?

ldv-alt:
The `ChangeLog` file is generated by "make dist".
Two other files, `CHANGELOG` and `ChangeLog-CVS`, are historic files that seem to be confusing rather than useful.

Timmmm:
Ah sorry I missed that. I agree the presence of 4 changelog files (one of which doesn't exist initially) is quite confusing! Probably best to delete them or move them into a subdirectory. At the very least `CHANGLOG` should be renamed to `CHANGELOG.old` or similar.

I would probably just delete them if it were me, seeing as they're available in Git history still. You could mention them at the bottom of `NEWS`. And then maybe rename `NEWS` to `CHANGELOG` since that is the more popular name.

What do you think? I could make a PR to do all that if you like?

ldv-alt (approved):

spaette:
```
$ cd /tmp && tar xf Linux-PAM-1.6.1.tar.xz
$ cd Linux-PAM-1.6.1 && ls [A-Z]*
ABOUT-NLS  CHANGELOG  ChangeLog      Copyright  Make.xml.rules.in  Makefile.in  README
AUTHORS    COPYING    ChangeLog-CVS  INSTALL    Makefile.am        NEWS
$ 
```

PR #795, opened 2024-05-13
libpam: make /etc/passwd checks 8 bit clean

goeranu:
libpam/pam_modutil_chck_user.c: explicitly convert the character read from /etc/passwd to type "char" to get the same signedness as the one compared with. Otherwise the functionality will depend on the (implementation defined) signedness of "char" on the platform built for.

According to the stricter standards, non-ASCII user names are not valid. In reality user names use many different styles, nicely illustrated [here](https://systemd.io/USER_NAMES/). I can't see any harm it making the code work correctly also on platforms where _char_ defaults to _signed_. Only one line needed to be changed.

Tested on Fedora 40.


ldv-alt (commented):
There is a small typo in the commit message: `pam_modutil_chck_user`

goeranu:
> There is a small typo in the commit message: `pam_modutil_chck_user`

Thanks for that! I fixed it now. While I usually stay away from forced pushes, I felt it would be appropriate in this case.

ldv-alt (approved):

PR #803, opened 2024-05-30
pam_set_item: disallow setting service to NULL

Blarse:
This also prevents a possible segfault when pam_set_item tries to convert an empty service_name to lower case.

### DIFF
--- a/libpam/pam_item.c
+++ b/libpam/pam_item.c
@@ -38,6 +38,12 @@ int pam_set_item (pam_handle_t *pamh, int item_type, const void *item)
     switch (item_type) {
 
     case PAM_SERVICE:
+	if (item == NULL) {
+	    pam_syslog(pamh, LOG_ERR,
+		       "pam_set_item: attempt to set service to NULL");
+	    retval = PAM_BAD_ITEM;
+	    break;
+	}
 	/* Setting handlers_loaded to 0 will cause the handlers
 	 * to be reloaded on the next call to a service module.
 	 */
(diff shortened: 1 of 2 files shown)


libpam/pam_item.c
```diff
@@ -38,6 +38,12 @@ int pam_set_item (pam_handle_t *pamh, int item_type, const void *item)
     switch (item_type) {
 
     case PAM_SERVICE:
+	if (item == NULL) {
+	    pam_syslog(pamh, LOG_ERR,
+		       "pam_set_item: attempt to set service to NULL");
+	    retval = PAM_BAD_ITEM;
+	    break;
+	}
 	/* Setting handlers_loaded to 0 will cause the handlers
```
ldv-alt, line 47: Given that without this change pam_set_item(pamh, PAM_SERVICE, NULL) just segfaults, and pam_start doesn't allow NULL service name, setting service to NULL is essentially not allowed anyway, so I think it's safe to apply this change.

With regards to the return value, PAM_BAD_ITEM is probably OK: it's the closest error code among those that are listed in the documentation for pam_set_item.  Apparently, pam_set_item(pamh, PAM_CONV, NULL) returns PAM_PERM_DENIED, but this is not documented.

ldv-alt (changes requested):
Looks good, thanks.
Please add your comment from the PR to the commit message itself.

Blarse:
> Please add your comment from the PR to the commit message itself.

I think this has already been done, if I understood you correctly.

ldv-alt (approved):

PR #805, opened 2024-06-05
pam_shells: Initialize econf_file* variable to NULL

schubi2:
The next version of libeconf (0.7.0) depends on an initialized econf_file* variable because it will be used for settings options. See:
https://github.com/openSUSE/libeconf/blob/master/NEWS


ldv-alt:
Does it mean that old users linked with libeconf would just crash after libeconf update?

schubi2:
We have discussed this again and we will not publish this release of libeconf. We will find another solution for it .....Sorry for the noise.

PR #806, opened 2024-06-05
modules/pam_succeed_if: Add support for evaluating pam env variables

3v1n0:
It may be useful in services files to be able to skip modules or take different paths depending on variables that may have been set by a previous module in the stack.

So make it possible for pam_succeed_if to get an env variable and do comparisons depending on its value.

Since variables can also be just set or unset, introduce a new syntax to perform such check, such as:

```
  env:VARIABLE is unset
  env:VARIABLE is set
```


modules/pam_succeed_if/pam_succeed_if.8.xml
```diff
@@ -217,6 +219,24 @@
...
+        <listitem>
+          <para>variable has not been set to any value.</para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>env:variable is value</term>
+        <listitem>
+          <para>variable equals value.</para>
+        </listitem>
+      </varlistentry>
```
ldv-alt, line 239: Why do you need this syntax, wouldn't `env:variable = value` be enough?

modules/pam_succeed_if/pam_succeed_if.c
```diff
@@ -350,6 +350,18 @@ evaluate(pam_handle_t *pamh, int debug,
 			tty == NULL)
 			tty = "";
 		left = (const char *)tty;
+	} else if (strncasecmp(left, "env:", 4) == 0 && strlen (left + 3) > 1) {
+		const char *env = left + 4;
```
ldv-alt, line 354: Please use `pam_str_skip_icase_prefix` instead.

modules/pam_succeed_if/pam_succeed_if.c
```diff
@@ -350,6 +350,18 @@ evaluate(pam_handle_t *pamh, int debug,
...
+		left = pam_getenv(pamh, env);
+		if (strcasecmp(qual, "is") == 0) {
+			if (strcasecmp(right, "set") == 0) {
+				return left != NULL ? PAM_SUCCESS : PAM_AUTH_ERR;
+			}
+			if (strcasecmp(right, "unset") == 0) {
+				return left == NULL ? PAM_SUCCESS : PAM_AUTH_ERR;
+			}
+			qual = "=";
```
ldv-alt, line 363: Please make sure this won't segfault in case of `left == NULL`.
Also, please make sure operations make sense when the variable is not set.

ldv-alt (changes requested):

modules/pam_succeed_if/pam_succeed_if.8.xml
3v1n0: As written in the commit message, for an env variable we may also just want to check if it's only set or unset (!= empty), and since this is something special to env variables only I didn't want to clash with current equality check not change its syntax for a special case.

To allow doing 
env:VARIABLE is unset
env:VARIABLE is set

However this was the thing for which I wasn't fully sure, about which one it was the best way to handle

ldv-alt: Sorry, I still don't see how `env:VARIABLE is arbitrary_value` is any better than `env:VARIABLE = arbitrary_value`, given that some values have a special meaning.

3v1n0: well, it was to have a specific mapping that applies only to env, but I'm fine to change it.

So the idea is to support `= set`? But then how to differentiate it from when the value need to be `"set"`

ldv-alt: I think it would be fine to have `env:VARIABLE is set|unset` and `env:VARIABLE op value` where `op` is any binary operation like `=`.

ldv-alt: The question is how to define `env:VARIABLE op value` for various `op` when `env:VARIABLE` is not set.

3v1n0: > The question is how to define `env:VARIABLE op value` for various `op` when `env:VARIABLE` is not set.

Well I'd say that's the case in which unset variable is basically considered an empty value, and if someone really wants to check if that's set to an empty value has to require both conditions: `env:VARIABLE is set env:VARIABLE = ''`.

PR #808, opened 2024-06-06
modules/pam_unix: fix snprintf argument type

amper128:
modules/pam_unix/passverify.c:503:9: warning: %u in format string (no. 1) requires 'unsigned int' but the argument type is 'signed int'. [invalidPrintfArgType_uint]
  sp += snprintf(sp, sizeof(salt) - (16 + 1 + (sp - salt)), "rounds=%u$", rounds);

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -500,7 +500,7 @@ PAMH_ARG_DECL(char * create_password_hash,
 #else
 	sp = stpcpy(salt, algoid);
 	if (on(UNIX_ALGO_ROUNDS, ctrl)) {
-		sp += snprintf(sp, sizeof(salt) - (16 + 1 + (sp - salt)), "rounds=%u$", rounds);
+		sp += snprintf(sp, sizeof(salt) - (16 + 1 + (sp - salt)), "rounds=%i$", rounds);
 	}
 	crypt_make_salt(sp, 16);
 #endif /* CRYPT_GENSALT_IMPLEMENTS_AUTO_ENTROPY */


modules/pam_unix/passverify.c
```diff
@@ -500,7 +500,7 @@ PAMH_ARG_DECL(char * create_password_hash,
 #else
 	sp = stpcpy(salt, algoid);
 	if (on(UNIX_ALGO_ROUNDS, ctrl)) {
-		sp += snprintf(sp, sizeof(salt) - (16 + 1 + (sp - salt)), "rounds=%u$", rounds);
+		sp += snprintf(sp, sizeof(salt) - (16 + 1 + (sp - salt)), "rounds=%i$", rounds);
```
ldv-alt, line 503: Given that `rounds < 0` doesn't make any sense, I suggest changing the type of `rounds` to an unsigned type instead.

ldv-alt (changes requested):

modules/pam_unix/passverify.c
amper128: Changing `rounds` to an unsigned type will require changing the argument type in `_set_ctrl()` also called in `pam_sm_acct_mgmt()`, `pam_sm_authenticate(),` `pam_sm_setcred()`, `pam_sm_open_session()` and `pam_sm_close_session()` where `int argc` is passed.

ldv-alt: Anyway, printing `rounds` as a signed integer doesn't make sense.

PR #824, opened 2024-09-03
Implement building Linux-PAM using meson

ldv-alt:
On my non-representative hardware, the full build using autotools
(`./autogen.sh && CFLAGS=-O2 ./configure && make -j$(nproc) && make -j$(nproc) install`)
takes about 45 seconds.

On the same hardware, the full build using meson
(`meson setup -Doptimization=2 dir && meson compile -C dir && meson install -C dir`)
takes just about 7.5 seconds.


ldv-alt:
@keszybz, could you have a look at this, please?

thkukuk:
I really like this switch. But please, if we  do the switch, remove the old configure script and Makefiles, maintaining two build systems is a nightmare.

ldv-alt:
> I really like this switch. But please, if we do the switch, remove the old configure script and Makefiles, maintaining two build systems is a nightmare.

My initial plan was to make a release that still supports both build systems, but contains a deprecation warning, and remove autotools support right after the release.  This would allow downstreams ample time to prepare for migration.  Do you think this approach is overcautious and we should rather do the switch right away?

thkukuk:
> This would allow downstreams ample time to prepare for migration. Do you think this approach is overcautious and we should rather do the switch right away?

I don't really see a need to prepare migration, in the end all distributions should meanwhile have packages using meson, so it's just replacing configure & make with the corresponding meson files.
But I'm fine if we keep configure for one more release and remove it only afterwards.
May fear is that we end like util-linux: configure and meson with the big problem that they are never in sync and both have their own problems and bugs.

ikerexxe:
For what it's worth, as a Fedora maintainer I'm fine with just providing meson and changing the build system for a release. It's a big change that I think will pay off in the medium/long term.

ldv-alt:
Well, if nobody sees any benefits in keeping autotools support for the next release, I'm fine with dropping it before the release.

conf/pam_conv1/meson.build
```diff
@@ -0,0 +1,22 @@
...
+  output: 'pam_conv_l.c',
+  depends: pam_conv_y,
+  command: [prog_flex, '-o', '@OUTPUT@', '@INPUT@'],
+)
+
+executable(
+  'pam_conv1',
+  sources: [pam_conv_l, pam_conv_y],
+  include_directories: [libpam_inc],
+  c_args: ['-Wno-unused-function', '-Wno-sign-compare'],
```
keszybz, line 20: A question of preference… but in general it's nicer to have such things one-per-line, because — inevitably — more args will be added, and then the diff and blame are more useful.

doc/guide-html-meson.build
```diff
@@ -0,0 +1,32 @@
...
+  install_dir: docdir,
+  install_tag: 'doc',
+)
+
+meson.add_install_script(
+  install_html,
+  meson.current_build_dir(),
+  docdir,
+  html,
+  install_tag: 'doc'
```
keszybz, line 31: `,` missing.

doc/guide-meson.build
```diff
@@ -0,0 +1,93 @@
...
+)
+
+
+html = custom_target(
+  input: xml,
+  output: name + '.html',
+  command: [
+    prog_xsltproc,
+    '-o',
+    '@OUTPUT@',
```
keszybz, line 34: Single line here for those two.

doc/guide-meson.build
```diff
@@ -0,0 +1,93 @@
...
+)
+
+custom_target(
+  input: html,
+  output: name + '.txt',
+  command: [
+    redir_exe,
+    '@INPUT@',
+    '@OUTPUT@',
+    browser
```
keszybz, line 53: `,`

doc/guide-meson.build
```diff
@@ -0,0 +1,93 @@
...
+)
+
+
+fop = custom_target(
+  input: xml,
+  output: name + '.fop',
+  command: [
+    prog_xsltproc,
+    '-o',
+    '@OUTPUT@',
```
keszybz, line 67: Same here.

doc/guide-meson.build
```diff
@@ -0,0 +1,93 @@
...
+  command: [
+    prog_fop,
+    '@INPUT@',
+    '@OUTPUT@',
+  ],
+  install: true,
+  install_dir: docdir,
+  install_tag: 'doc',
+)
+
```
keszybz, line 91: Unnecessary empty line here.

doc/man/meson.build
```diff
@@ -0,0 +1,70 @@
...
+               xml],
+              check: true)
+
+  custom_target(man[0],
+    input: xml,
+    output: man,
+    depends: custom_man_xsl,
+    command: [prog_xsltproc,
+              '-o',
+              '@OUTPUT0@',
```
keszybz, line 57: Here too.

thkukuk:
I'm trying to build our pam packages currently with meson and this PR. We build two packages: a very small one and a full features one.

I see the following error:
```
[   10s] Run-time dependency libsystemd found: NO (tried pkgconfig)
[   10s] 
[   10s] meson.build:307:13: ERROR: Dependency "libsystemd" not found, tried pkgconfig

```
I have to explicit disable logind. But that is defined as "auto", so shouldn't meson skip libsystemd dependency if it does not find it?
If I disable logind with "-Dlogind=disabled", it works.

Similar for pam_userdb:
`meson.build:332:15: ERROR: C shared or static library 'db' not found
`
pam_userdb is "auto", but we get an error if the required libdb is not found. Disabling pam_userdb also solves this. 
Same for "nis" and libnsl and "docs" and the dependencies.

It's irritating that you need to look at the meson.build script to find out which option you  need to disable if you don't want a dependency.

keszybz (approved):
This is a lot of lines to review ;) I gave it a quick scan, and it looks all very nice. There *probably* are some minor things that'll need to be adjusted, but I think it's best to just merge this and use it.

I made some small suggestions. The only bigger issue I wonder about is whether it makes sense to symlink a shared `meson.build` file into multiple subdirectories. IIUC, this was done to allow the outputs to be produced in those subdirectories. If that is not necessary, I'd consider just using a single file. But maybe that's not feasible…

keszybz:
BTW, I get this:
```
[34/362] Generating doc/specs/parse_y.[ch] with a custom command
../doc/specs/parse_y.y: warning: 2 shift/reduce conflicts [-Wconflicts-sr]
../doc/specs/parse_y.y: warning: 8 reduce/reduce conflicts [-Wconflicts-rr]
../doc/specs/parse_y.y: note: rerun with option '-Wcounterexamples' to generate conflict counterexamples
../doc/specs/parse_y.y:97.3-123.1: warning: rule useless in parser due to conflicts [-Wother]
   97 | | doc stuff RIGHT stuff RIGHT stuff NEWLINE {
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Maybe this is related to bison version? (I have `bison-3.8.2-7.fc40.x86_64` here.)

thkukuk:
There are missing options which we used with configure:
- `--htmldir`
- `--pdfdir`

ldv-alt:
> I'm trying to build our pam packages currently with meson and this PR. We build two packages: a very small one and a full features one.
> 
> I see the following error:
> 
> ```
> [   10s] Run-time dependency libsystemd found: NO (tried pkgconfig)
> [   10s] 
> [   10s] meson.build:307:13: ERROR: Dependency "libsystemd" not found, tried pkgconfig
> ```
> 
> I have to explicit disable logind. But that is defined as "auto", so shouldn't meson skip libsystemd dependency if it does not find it? If I disable logind with "-Dlogind=disabled", it works.

I didn't expect this behavior.  I definitely tested this, and I get the following instead without any errors:
```
Run-time dependency libsystemd found: NO (tried pkgconfig and cmake)
```

I expect that
```
option('logind', type: 'feature', value: 'auto',
       description: 'logind support in pam_issue and pam_timestamp')
...
libsystemd = dependency('libsystemd', version: '>= 254', required: get_option('logind'))
```
would make libsystemd dependency optional.  I wonder why it doesn't work this way on your side.

This is meson 1.5.1.

@keszybz, any ideas why this check would behave so differently?

ldv-alt:
> There are missing options which we used with configure:
> 
>     * `--htmldir`
> 
>     * `--pdfdir`

Indeed, these are provided by autoconf.  Added.

ldv-alt:
> > I'm trying to build our pam packages currently with meson and this PR. We build two packages: a very small one and a full features one.
> > I see the following error:
> > ```
> > [   10s] Run-time dependency libsystemd found: NO (tried pkgconfig)
> > [   10s] 
> > [   10s] meson.build:307:13: ERROR: Dependency "libsystemd" not found, tried pkgconfig
> > ```
> > I have to explicit disable logind. But that is defined as "auto", so shouldn't meson skip libsystemd dependency if it does not find it? If I disable logind with "-Dlogind=disabled", it works.
> 
> I didn't expect this behavior. I definitely tested this, and I get the following instead without any errors:
> ```
> Run-time dependency libsystemd found: NO (tried pkgconfig and cmake)
> ```

This (no libsystemd, no error) also happens in github CI.

thkukuk:
The cc.find_library() calls are missing "required=false" as argument. From meson documentation:
`If set true, Meson will abort with an error if the library could not be found. Otherwise, Meson will continue and the found method of the returned object will return false.`
`default = true`
No idea why it works for you, it should not if cc.find_library() get's called according to the documentation. Adding "required=false" fixes it for me.

ldv-alt:
> The cc.find_library() calls are missing "required=false" as argument.

Indeed, but this affects only the pam_userdb case.  The use of crypt library doesn't seem to be optional because it's used unconditionally by pam_pwhistory, and other libraries are not looked up using cc.find_library().

thkukuk:
Of course not all find_library() calls require "required: false" as argument. Some have it already (like 'util'), other don't need it, but the three ones in the userdb case are missing it.

thkukuk:
> This (no libsystemd, no error) also happens in github CI.

According to the meson documentation the libsystemd check is correct and should work. But I see the error on all systems with meson 1.3.1 and 1.5.1. Need to try to debug that later...

thkukuk:
Between, I updated our PAM package with this PR and adjusted the spec file, this was really simple.
So for me, the current state is good enough to get merged.

ldv-alt:
The only remaining issue I'm aware of is what do we do with the `make releasedocs` concept.

The idea of distributing pre-built documentation became flawed when we started to use profile-docbook.xsl to generate parametrized manual pages. 
Now, distributing anything pre-built wouldn't play well with the meson model of placing everything it builds into a separate build directory tree.

All that goes slightly beyond the scope of this PR, but should be decided upon before the final removal of autotools build support.

ldv-alt:
> Well, if nobody sees any benefits in keeping autotools support for the next release, I'm fine with dropping it before the release.

OK, I've submitted a PR that drops autotools support for the next release: #839.

PR #825, opened 2024-09-04
pam_issue: only count class user

thkukuk:
systemd added new types of classes (e.g. manager*). So we cannot use the count of all sessions anymore, but have to check which class this is.
This is backward compatible, systemd v209 or newer is required.

### DIFF
--- a/modules/pam_issue/pam_issue.c
+++ b/modules/pam_issue/pam_issue.c
@@ -165,13 +165,31 @@ read_issue_quoted(pam_handle_t *pamh, FILE *fp, char **prompt)
 		{
 		    unsigned int users = 0;
 #ifdef USE_LOGIND
-		    int sessions = sd_get_sessions(NULL);
+		    char **sessions_list;
+		    int sessions = sd_get_sessions(&sessions_list);
 
 		    if (sessions < 0) {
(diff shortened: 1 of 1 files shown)


modules/pam_issue/pam_issue.c
```diff
@@ -165,13 +165,31 @@ read_issue_quoted(pam_handle_t *pamh, FILE *fp, char **prompt)
...
+			  continue;
+
+			if (strncmp(class, "user", 4) == 0) // user, user-early, user-incomplete
+			  users++;
+			free(class);
+		      }
+
+		      for (i = 0; i < sessions; i++)
+			free(sessions_list[i]);
+		      free(sessions_list[i]);
```
ldv-alt, line 192: The second `free(sessions_list[i])` looks wrong.

ldv-alt (changes requested):
Also, it would be great if the PR description was copied into the commit message.

modules/pam_issue/pam_issue.c
thkukuk: Correct, fixed, thanks.

ldv-alt (approved):

Issue #822, opened 2024-08-27
pam_limits: possibly incomplete fix for #537


StephDC:
Ref: #537 

Apparently the issue shown above is not completely fixed by #541 , as pam_limits still gets the number of sessions a user have from utmp.

https://github.com/linux-pam/linux-pam/blob/4d1d4f699551f395be065e60513eaf85f01b84e6/modules/pam_limits/pam_limits.c#L39

However on a system that uses `systemd-logind`, this creates a nasty discrepancy between the number of sessions seen from pam_limits and pam_systemd. One can use `ssh -N` to login to the system, allowing pam_systemd to create a new session invisible to pam_limits, thus bypassing the maxlogins limit set to such user. In extreme cases, this could result in more sessions than systemd-logind handles (SessionsMax=8192) being requested and those excess sessions remained in sshd.service, or if pam_systemd was set to requisite, everyone including root from that point would be denied login.

thkukuk:
If you compile pam_limits with systemd support, utmp support is not compiled in.
So if your pam_limits module still uses utmp, it is not compiled with systemd support.
So most important question: is your pam_limits module compiled with systemd support or not?
Next problem: I would never use pam_limits or anything else to limited logged in users, as there is no definition, what a logged in user is. Even pam_systemd does not count every logged in user as logged in user. So yes, you will always have a discrepancy between what different tools reports. And without standard, which describes exactly when a user as to be counted as logged in and when not, this is not fixable. And thus a standard does not exist and will, in my opinion, never come.

StephDC:
Would you mind to point to me from within pam_limits/pam_limits.c where it omits the utmp.h?

A quick ldd check told me that while the other two files, pam_issue.so and pam_timestamp.so was linked against libsystemd.so.0, pam_limit.so is not.

I also did not find the `#ifdef USE_LOGIND` part around the `#include <utmp.h>`. Only a simple `#include <utmp.h>` whenever the system has `#ifdef __linux__`.

I am on the master branch commit https://github.com/linux-pam/linux-pam/commit/4d1d4f699551f395be065e60513eaf85f01b84e6 just in case I missed something.

thkukuk:
Sorry, I should have looked more carefull: Yes, the PR did not add systemd support to pam_limits, and nobody did ever claim that that PR would add it.

StephDC:
Thus, the issue #537 is not completely fixed, and pam_limits.c as mentioned in such issue remains to be fixed.

[...]

PR #826, opened 2024-09-06
pam_limits: use systemd-logind instead of utmp (#822)

thkukuk:
The utmp database is unreliable for counting logged in users, since there is no standard which defines who should create an entry at which time for which reason. And it has a Y2038 problem with glibc/x86-64. Query systemd-logind for the number of user sessions instead.

Using systemd-logind requires that "maxlogins" is at minimum "2", since pam_systemd for the first login calls "systemd --user", which calls again the PAM stack. But this time there is already an user session, so with maxlogins=1 systemd would get a denied by PAM and the session management is in an undefined state.

### DIFF
--- a/modules/pam_limits/pam_limits.c
+++ b/modules/pam_limits/pam_limits.c
@@ -36,7 +36,12 @@
 #include <sys/resource.h>
 #include <limits.h>
 #include <glob.h>
+#ifdef USE_LOGIND
+#include <systemd/sd-login.h>
+#else
 #include <utmp.h>
+#endif
+
 #ifndef UT_USER  /* some systems have ut_name instead of ut_user */
 #define UT_USER ut_user
 #endif
@@ -240,7 +245,6 @@ static int
 check_logins (pam_handle_t *pamh, const char *name, int limit, int ctrl,
               struct pam_limit_s *pl)
 {
-    struct utmp *ut;
     int count;
 
     if (ctrl & PAM_DEBUG_ARG) {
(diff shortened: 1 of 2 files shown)


thkukuk:
Question: what would be the correct return value, if the limit is too low or we have a systemd error?
LOGIN_ERR is according to the description wrong.
LIMIT_ERR is ignored, which also means, this limit is ignored.

modules/pam_limits/pam_limits.c
```diff
@@ -273,6 +275,77 @@ check_logins (pam_handle_t *pamh, const char *name, int limit, int ctrl,
...
+	char *user = NULL;
+	char *class = NULL;
+
+	if (sd_session_get_class(sessions_list[i], &class) < 0 || class == NULL)
+	  continue;
+
+	if (strncmp(class, "user", 4) != 0)  { /* user, user-early, user-incomplete */
+	  free (class);
+	  continue;
+	}
```
ldv-alt, line 311: Looks like `class` is leaked here in case of `strncmp(class, "user", 4) == 0`.

ldv-alt:
> Question: what would be the correct return value, if the limit is too low or we have a systemd error? LOGIN_ERR is according to the description wrong. LIMIT_ERR is ignored, which also means, this limit is ignored.

Yes, pam_limits's choice of handling errors is unfortunate, but it's probably too late to change.

ldv-alt (approved):

PR #827, opened 2024-09-06
pam_userdb: don't overwrite free'd memory

keeffturner:
crypt_r can return a pointer into a provided crypt_data struct.  Callers should not modify the string returned by crypt_r after freeing the corresponding crypt_data struct.


ldv-alt (approved):
@keeffturner, I've reworked this PR a bit, please have a look.

keeffturner (commented):
ldv's changes look good to keeffturner

PR #828, opened 2024-09-10
pam_userdb: don't return void result as int

thkukuk:
gdbm_close() returns void, so we cannot use that as return argument, if the calling function returns int.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -372,6 +372,12 @@ else
       if have
         use_db = 'gdbm'
         cdata.set('HAVE_GDBM_H', 1)
+        if cc.compiles('''#include <gdbm.h>
+                       int db_close(void *dbm) {return gdbm_close(dbm);}''',
+                       args: '-D_GNU_SOURCE',
+                       name: 'gdbm_close returns int')
+          cdata.set('GDBM_CLOSE_RETURNS_INT', 1)
+        endif
       endif
     endif
   endif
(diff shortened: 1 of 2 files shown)


thkukuk:
I found this compiler error by accident when testing meson.

ikerexxe:
That's not what the [man pages](https://manpages.debian.org/testing/libgdbm-dev/gdbm_close.3.en.html) state. Is there any updated man pages? Or maybe you are using an old GDBM version.

thkukuk:
It's what the header file and the compiler states on my systems.

ldv-alt (approved):

ikerexxe:
I checked just in case and my system header says it returns an `int`. Digging into the history of this package I discovered the following change:
```
Version 1.17 - 2018-07-30

* int gdbm_close and gdbm_sync

Both functions now return 0 on success. On error, they return -1
and set gdbm_errno and errno to the appropriate error codes.
```

I'm using GDBM v1.23, so it makes sense that I see an `int` as the return statement. Can you check if you are using an updated version of this package on your system?

thkukuk:
No idea, I did just send the package to a bunch of test systems.
If Linux-PAM requires now a specific libgdbm version, we need introduce a dependency for that.
In anyways, it's in both cases a bug of Linux-PAM, as it compiled before.

ldv-alt:
I've added a gdbm_close return type check.
@thkukuk @ikerexxe, please have a look.

ikerexxe (approved):
LGTM! Thanks for taking the feedback into account

PR #839, opened 2024-10-14
build: drop autotools support

ldv-alt:
Following on from the discussion at #824, let's finish the migration from autotools to meson before the next release.


.gitignore
```diff
@@ -1,35 +0,0 @@
-*.3
```
ikerexxe, line 1: You completely removed `.gitignore` file, is that right? Shouldn't you add some directory like `build`?

ldv-alt: I'm not sure we should add a build directory to `.gitignore` file, because meson doesn't mandate any particular name for the build directory, and it doesn't have to be placed inside the source tree.

ikerexxe: As a project we can agree to use a standard folder and avoid cluttering the repository with unwanted files. What I want is to avoid a long list of modified files in git that have been created by the build system that I clearly don't want to upload to the repository. Whether someone decides to use another folder inside or outside the source tree is something they will have to deal with on their own.

cgzones: For me meson automatically generates a `build/.gitignore` file with the following content:
```
# This file is autogenerated by Meson. If you change or delete it, it won't be recreated.
*
```

So is a project wide ignore rule necessary?

ldv-alt: Apparently, no project wide .gitignore is necessary.  I've just re-added to CI the same `git status` check we had for autotools-based builds, and it passes.

ldv-alt:
Any more comments?

ikerexxe (approved):
I took another look and it looks good to me

ldv-alt:
OK, let's move on.

PR #840, opened 2024-10-16
pam_get_authtok.c: add a check so that pamh is not NULL

andr3y-k0v4l3v:
This also prevents a possible segfault when pam_get_authtok_internal and pam_get_authtok_verify accesses the pamh variable with a NULL value.


tests/tst-pam_get_authtok.c
```diff
@@ -0,0 +1,62 @@
...
+      return 1;
+    }
+
+  /* 2: Call pam_get_authtok with NULL as pam handle */
+  retval = pam_get_authtok (NULL, 0, &authtok, prompt);
+  if (retval != PAM_SYSTEM_ERR)
+    {
+      fprintf (stderr, "pam_get_authtok (NULL, ...) not returned PAM_SYSTEM_ERR\n");
+      return 1;
+    }
```
ldv-alt, line 59: Please use `ASSERT_EQ` from test_assert.h.

See e.g. tests/tst-pam_get_data.c

ldv-alt (changes requested):
doc/man/pam_get_authtok.3.xml contains `RETURN VALUES` section that lists valid return values of the functions being changed in this PR, please update this documentation.

ldv-alt (approved):

PR #841, opened 2024-10-18
pam_access_fixes

ikerexxe:
* modules/pam_access/pam_access.c: match the local address regardless of
  the IP version in use.
* modules/pam_access/access.conf.5.xml: `LOCAL` keyword behaviour
  explanation was focused on the development internals. Let's clarify it
  by rephrasing it to something a sysadmin can understand.

In some circumstances the `localhost` may be translated to IPv4 or IPv6,
but the configuration file only indicated the address for one of the two
versions. Since the originating value is set in `PAM_RHOST` and PAM has
no control over it, let's match the local addresses regardless of the IP
version in use.

Resolves: https://issues.redhat.com/browse/RHEL-23018
Resolves: https://issues.redhat.com/browse/RHEL-39943

### DIFF
--- a/modules/pam_access/pam_access.c
+++ b/modules/pam_access/pam_access.c
@@ -306,6 +306,23 @@ isipaddr (const char *string, int *addr_type,
   return is_ip;
 }
 
+/* is_local_addr - checks if the IP address is local */
+static int
+is_local_addr (const char *string, int addr_type)
+{
+  if (addr_type == AF_INET) {
+    if (strcmp(string, "127.0.0.1") == 0) {
+      return YES;
+    }
+  } else if (addr_type == AF_INET6) {
+    if (strcmp(string, "::1") == 0) {
+      return YES;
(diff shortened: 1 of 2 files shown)


modules/pam_access/pam_access.c
```diff
@@ -306,6 +306,23 @@ isipaddr (const char *string, int *addr_type,
...
+      return YES;
+    }
+  } else if (addr_type == AF_INET6) {
+    if (strcasecmp(string, "::1") == 0) {
+      return YES;
+    }
+  }
+
+  return NO;
+}
```
ldv-alt, line 324: Hmm, why strcasecmp?

ldv-alt: Given that ip addresses don't contain anything case sensitive, `strcmp` should be enough.

ldv-alt (approved):

PR #843, opened 2024-10-22
Avoid accessing null pointer

bin-ly:
If not satisfied  “ handler_type == PAM_HT_MODULE || handler_type == PAM_HT_SILENT_MODULE) &&
mod_path != NULL)”  this condition, mod is always NULL, then there may be unpredictable errors in   “ _pam_dlsym (mod->dl_handle, sym)”


ldv-alt:
This analysis is wrong, there is no potential NULL pointer dereference.

PR #852, opened 2024-11-05
doc: make PDF documentation optional

yselkowitz:
While building any documentation requires xsltproc, only the PDF documentation requires fop, which is dependency-heavy, and not always available (e.g. it is not included in RHEL).  This allows the HTML documentation to be built while skipping the PDFs.

/cc @ikerexxe

### DIFF
--- a/doc/guide-meson.build
+++ b/doc/guide-meson.build
@@ -56,35 +56,37 @@ custom_target(
 )
 
 
-fop = custom_target(
-  input: xml,
-  output: name + '.fop',
-  command: [
-    prog_xsltproc,
-    '-o', '@OUTPUT@',
-    '--nonet',
-    '--xinclude',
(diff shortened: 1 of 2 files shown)


doc/guide-meson.build
```diff
@@ -56,6 +56,7 @@ custom_target(
 )
 
 
+if prog_fop.found()
 fop = custom_target(
```
ldv-alt, line 60: Is this hunk really needed?  If yes, could you indent it, please?

ldv-alt (changes requested):

ikerexxe:
Impressive! This makes my life easier when it comes to packaging.

doc/guide-meson.build
yselkowitz: Indented.

ldv-alt (approved):

PR #855, opened 2024-11-15
pam_modutil_search_key: prevent possible null pointer dereference

Blarse:
Prevent a possible null pointer dereference in strcasecmp when key is NULL.


ldv-alt (dismissed):
Note that `key` argument is marked with `PAM_NONNULL` attribute,
so passing NULL would be a programming error.

Blarse:
> Note that `key` argument is marked with `PAM_NONNULL` attribute, so passing NULL would be a programming error.

In this case I agree with you, it is a programming error on the program side. Closing this PR.

PR #856, opened 2024-11-15
build: add static_library targets for libpam, libpam_misc

MingcongBai:
Previously with Autotools, it was possible to build libpam and libpam_misc as static libraries. This was useful for Busybox to link against and Debian is also likely to need them (as part of libpam-dev).

### DIFF
--- a/libpam/meson.build
+++ b/libpam/meson.build
@@ -58,6 +58,21 @@ libpam = shared_library(
   install: true,
 )
 
+libpam = static_library(
+  'pam',
+  sources: libpam_src,
+  include_directories: [libpam_inc],
+  c_args: [
+    '-DDEFAULT_MODULE_PATH="@0@/"'.format(securedir),
+    '-DLIBPAM_COMPILE',
+  ],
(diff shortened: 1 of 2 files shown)


thkukuk:
> Previously with Autotools, it was possible to build libpam and libpam_misc as static libraries. This was useful for Busybox to link against and Debian is also likely to need them (as part of libpam-dev).

Support for static linking was dropped long ago (8.5 years), that this was still possible for the libraries was an oversight.
What you want to do is not really possible: the PAM modules will still be loaded via dlopen(), so your binary is **not** static.
And even worse: since PAM modules are linked against libpam.so, your binary will suddenly have two libpam: the linked static one, and the one pulled in via the PAM module. And if this two libpam versions are incompatible...

So I advice against building libpam static. Or we should at minimum do what glibc is doing: if you link against a function static which is not static, print a fat compiler warning, that what you do will not work as you expect.

ldv-alt (changes requested):
By any means please do not enable building of these static libraries by default.
If you have to build them, please make this optional.

PR #859, opened 2024-11-24
meson: support elogind as a logind provider

Dudemanguy:
The systemd functionality used here is also supported by elogind. Search for libelogind as well and use the appropriate headers.

Not sure what approach is wanted here. I opted to not change the semantics of the `logind` option and just search for both libraries.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -318,9 +318,16 @@ else
   stringparam_profileconditions += ';no_openssl_hmac'
 endif
 
-
-libsystemd = dependency('libsystemd', version: '>= 254', required: get_option('logind'))
-cdata.set('USE_LOGIND', libsystemd.found() ? 1 : false)
+use_logind = get_option('logind')
+use_elogind = get_option('elogind').require(not use_logind.enabled()).disable_auto_if(use_logind.enabled())
+use_logind = use_logind.disable_auto_if(use_elogind.enabled())
+logind_dep = dependency('libsystemd', version: '>= 254', required: use_logind)
+cdata.set('USE_LIBSYSTEMD', logind_dep.found() ? 1 : false)
+if not logind_dep.found()
+  logind_dep = dependency('libelogind', version: '>= 254', required: use_elogind)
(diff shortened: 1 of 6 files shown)


ldv-alt:
Does your patch build with -Dlogind=disabled?

ldv-alt:
I was thinking about the following logic:
```
logind_dep = null_dep
use_elogind = get_option('elogind')
use_logind = get_option('logind').disable_auto_if(use_elogind.enabled())
if use_logind.allowed()
  logind_dep = dependency('libsystemd', version: '>= 254', required: use_logind)
  cdata.set('USE_LIBSYSTEMD', logind_dep.found() ? 1 : false)
endif
if not logind_dep.found() and use_elogind.allowed()
  logind_dep = dependency('libelogind', version: '>= 254', required: use_elogind)
  cdata.set('USE_LIBELOGIND', logind_dep.found() ? 1 : false)
endif
cdata.set('USE_LOGIND', logind_dep.found() ? 1 : false)
```

However, it would allow `-Dlogind=enabled -Delogind=enabled`, which is probably counter-intuitive.

ldv-alt:
Maybe the following logic would work better (completely untested):
```
logind_dep = null_dep
use_logind = get_option('logind')
use_elogind = get_option('elogind').require(not use_logind.enabled()).disable_auto_if(use_logind.enabled())
use_logind = use_logind.disable_auto_if(use_elogind.enabled())
libsystemd = dependency('libsystemd', version: '>= 254', required: use_logind)
libelogind = dependency('libelogind', version: '>= 254', required: use_elogind)
if libsystemd.found()
  logind_dep = libsystemd
  cdata.set('USE_LIBSYSTEMD', 1)
elif libelogind.found()
  logind_dep = libelogind
  cdata.set('USE_LIBELOGIND', 1)
endif
cdata.set('USE_LOGIND', logind_dep.found() ? 1 : false)
```

Dudemanguy:
That works well. I updated the patch with that logic instead which also simplifies the dependencies in the modules.

ldv-alt (approved):

PR #861, opened 2024-11-28
meson.build: correct check for existence of two preprocessor defines

kanavin:
sizeof is meant for *types*, and the test program produced by it has incorrect syntax

__NR_keyctl something;

and will always fail to compile.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -198,12 +198,12 @@ foreach ident: check_functions
   endif
 endforeach
 
-enable_pam_keyinit = cc.sizeof('__NR_keyctl', prefix: '#include <sys/syscall.h>') > 0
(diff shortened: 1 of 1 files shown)


ldv-alt (approved):

ldv-alt (changes requested):
This change results to the following warnings:

```
meson.build:201: WARNING: Project targets '>= 0.62.0' but uses feature introduced in '1.3.0': compiler.has_define.
meson.build:206: WARNING: Project targets '>= 0.62.0' but uses feature introduced in '1.3.0': compiler.has_define.
```

I don't think we can raise the required meson version to 1.3.0.

ldv-alt (approved):

PR #862, opened 2024-12-03
pam_unix: fix diagnostic message in debug mode

JetXujing:
When configured using -Dpam-debug=true, _pam_unix_approve_pass prints
a diagnostic message with addresses of password strings.  Apparently,
since the times predating the git history of the project in this
diagnostic message the addresses of the old and new passwords were
mixed up.


ldv-alt (changes requested):
Thanks, this seems to be quite an old typo that predates the git history of the project.

Please amend the commit message so it would be clear that the diagnostic message being fixed is the message printed only when PAM is configured using -Dpam-debug=true.

JetXujing:
Thank you for your review. The submission information has been updated.

ldv-alt (approved):

PR #869, opened 2025-01-01
pam_listfile: Few changes on the checks done on the configuration file

cquike:
I have been using pam_listfile for a while and I was curious about some details of the implementation. While looking at the source code I found a couple of issues in the checks done on the configuration file that I thought could be improved:  

- Currently the file checks are done with lstat() and afterwards the file is open with open(). This has the potential issue of someone being able to replace the configuration file after the checks have been done. I made a check in which I point to a configuration file in /tmp and I added an artificial additional sleep() between the checks and the open() calls. In between a malicious user can change the configuration file, specially if the user has write permission in the directory.

- Check that the configuration file is owned by the user which runs the PAM application. For the typical usage in a Linux login this would mean root.

- Consistently return error code based on onerr parameter. According to the documentation the onerr parameter specifies what return code should be set if there is a problem with the configuration file.

This is my first PR to linux-pam, so apologies if it does not follow some standard practices of the project.

### DIFF
--- a/modules/pam_listfile/pam_listfile.c
+++ b/modules/pam_listfile/pam_listfile.c
@@ -12,6 +12,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/types.h>
+#include <fcntl.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #include <syslog.h>
@@ -60,6 +61,7 @@ pam_listfile(pam_handle_t *pamh, int argc, const char **argv)
     char *aline=NULL;
     const char *apply_val;
     struct stat fileinfo;
+    int fd;
     FILE *inf;
     int apply_type;
     size_t n=0;
@@ -263,7 +265,6 @@ pam_listfile(pam_handle_t *pamh, int argc, const char **argv)
 		break;
 	    default:
 		pam_syslog(pamh,LOG_ERR,
-
 			 "Internal weirdness, unknown extended item %d",
 			 extitem);
 		return onerr;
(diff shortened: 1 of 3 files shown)


modules/pam_listfile/pam_listfile.c
```diff
@@ -307,6 +316,7 @@ pam_listfile(pam_handle_t *pamh, int argc, const char **argv)
     assert(PAM_SUCCESS == 0);
     assert(PAM_AUTH_ERR != 0);
 #endif
+    inf = fdopen(fd,"r");
```
cgzones, line 319: `fdopen(3)` can theoretically fail

cquike:
Indeed, thanks for catching that. I have now changed the code to take into account that fdopen can fail.

modules/pam_listfile/pam_listfile.c
cquike: Indeed, thanks for catching that. I have now changed the code to take into account that fdopen can fail.

modules/pam_listfile/pam_listfile.c
```diff
@@ -275,9 +276,24 @@ pam_listfile(pam_handle_t *pamh, int argc, const char **argv)
...
+	return onerr;
+    }
+
+    if (fileinfo.st_uid != getuid()) {
+	pam_syslog(pamh,LOG_ERR,
+		 "%s is not owned by the user running the application using PAM",
+		 ifname);
+        close(fd);
 	return onerr;
     }
```
ldv-alt, line 298: Is there any particular reason why do you want to unconditionally prohibit opening files belonging a different user, e.g. root?

cquike: The idea is that only a user with the proper privileged is the owner of the file. This could avoid situations in which the configuration file points to a file that does not exist in the first place but is created by a non-privileged user afterwards.
However I agree that root should be allowed, since by definition it has ll the possible privileges. I have now updated the code to reflect that.

cquike: Hi,
did you have the chance to review the last change? I have now rebased the branch on top of master.
Thanks!

cquike:
I have now rebased this on top of master. Any additional comments on this PR will be appreciated.
Thanks!

ldv-alt (changes requested):
Please do not squash unrelated changes into a single commit.

cquike:
> Please do not squash unrelated changes into a single commit.
You mean that the three different items that have been improved should have three different commits?
Thanks!

PR #871, opened 2025-01-16
pam_access: fix stack overflow with huge config

stoeckmann:
If /etc/security/access.conf contains a line with a huge amount of exceptions, the recursive nature of list_match could lead to a stack overflow.

Implement the logic without falling back to recursion.

Proof of Concept:

1. Use `pam_access.so` for e.g. `login`
```
# cat > /etc/pam.d/login << EOF
auth required pam_access.so
auth required pam_unix.so
account required pam_unix.so
password required pam_unix.so
session required pam_unix.so
EOF
```

2. Create `/etc/security/access.conf` with lots of exceptions
```
# python -c 'print("+:ALL" + 131072 * " EXCEPT ALL" + ":ALL")' > /etc/security/access.conf
```

3. Run `login` (from shadow-utils for an easy login tool which can be run from terminal emulator)
```
# login
host login: root
Segmentation fault
```

Explanation of implementation:

Take a line like this one:
```
-:ALL EXCEPT ALL:ALL
```

It implies that for everyone except everyone access should be denied. So in fact, this line matches nobody.

If you take this one instead:
```
-:ALL EXCEPT ALL EXCEPT ALL:ALL
```

It implies that everyone matches, because "except everyone" is further excepted by everyone.

To abstract these lines, see it like this:
```
-:+++ EXCEPT --- EXCEPT +++:+++
```

At first, we are including (+++). If we encounter an EXCEPT, we become excluding (---) and then including again and so on. We continue this until we encounter one such block (+++ or ---) which we did not match. All further exceptions are not meant for us.


stoeckmann:
Created a similar PR in shadow to stay in sync (differing code styles and strtok vs strsep, but same logic).

stoeckmann:
Since question arised in shadow PR if I can explain the refactoring in more detail, I'll do it here for single point of discussion:

- Considering the recursive calls, the old code splitted this single line into smaller chunks. Each chunk was split by "EXCEPT". Pretty much like a loop around `strsep(... "EXCEPT")`.
- With each recursion, the code checked if the looked up string is within such a chunk and returns YES or NO.
- To properly support the inversion of meaning of "EXCEPT" blocks, the result of recursive calls is negated.

This in turn means that, if the code keeps recursing into itself, it becomes negations of negations of negations ...

The new iterative approach "remembers" the meaning of the current chunk (inclusion). It also remembers if a match happened within the current chunk (matched). Since we don't need any further information about which exact chunk we are in, we can simply flip (including) whenever we jump to the next chunk. And this flip happens whenever "EXCEPT" is encountered.

ikerexxe:
> ```
> -:ALL EXCEPT ALL:ALL
> ```
> 
> It implies that for everyone except everyone access should be denied. So in fact, this line matches nobody.
> 
> If you take this one instead:
> 
> ```
> -:ALL EXCEPT ALL EXCEPT ALL:ALL
> ```

Are any of the previous lines examples of a `huge amount of exceptions`? Or did you have something else in mind?

> Created a similar PR in shadow to stay in sync (differing code styles and strtok vs strsep, but same logic).

Not to lose sight of it, https://github.com/shadow-maint/shadow/pull/1187 is the PR mentioned by Tobias.

alejandro-colomar:
> Are any of the previous lines examples of a `huge amount of exceptions`? Or did you have something else in mind?

This one is:

```
# python -c 'print("+:ALL" + 131072 * " EXCEPT ALL" + ":ALL")' > /etc/security/access.conf
```

(pasted from <https://github.com/linux-pam/linux-pam/pull/871#issue-2793700383>).

See:

```sh
$ python3 -c 'print("+:ALL" + 131072 * " EXCEPT ALL" + ":ALL")' | wc
      1  262145 1441802
```

PR #872, opened 2025-01-21
meson: Find programs using the native file only, not the cross file

chewi:
These should all be native programs found on the build machine. We should not try to execute them via QEMU. Users would therefore expect to configure them using the native file, not the cross file.

See https://mesonbuild.com/Reference-manual_functions.html#find_program.


chewi:
CC @thesamesam.

ldv-alt (approved):

PR #883, opened 2025-02-24
pam_namespace: fix logic in return value handling

obp-anssi:
The case in which protect_dir() returns an error and the flag POLYDIR_CREATE (flag "create" in namespace.conf) is not set was not handled. Therefore, the program continued without a polydir and returned later on failed mount(2) or stat(2) calls.

### DIFF
--- a/modules/pam_namespace/pam_namespace.c
+++ b/modules/pam_namespace/pam_namespace.c
@@ -1654,16 +1654,14 @@ static int ns_setup(struct polydir_s *polyptr,
 
     retval = protect_dir(polyptr->dir, 0, 0, idata);
 
-    if (retval < 0 && errno != ENOENT) {
-	pam_syslog(idata->pamh, LOG_ERR, "Polydir %s access error: %m",
-		polyptr->dir);
-	return PAM_SESSION_ERR;
-    }
-
     if (retval < 0) {
-	if ((polyptr->flags & POLYDIR_CREATE) &&
-		create_polydir(polyptr, idata) != PAM_SUCCESS)
-		return PAM_SESSION_ERR;
+        if (errno != ENOENT || !(polyptr->flags & POLYDIR_CREATE)) {
+            pam_syslog(idata->pamh, LOG_ERR, "Polydir %s access error: %m",
+                    polyptr->dir);
+            return PAM_SESSION_ERR;
+        }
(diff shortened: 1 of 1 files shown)


modules/pam_namespace/pam_namespace.c
```diff
@@ -1654,7 +1654,7 @@ static int ns_setup(struct polydir_s *polyptr,
 
     retval = protect_dir(polyptr->dir, 0, 0, idata);
 
-    if (retval < 0 && errno != ENOENT) {
+    if (retval < 0 && (errno != ENOENT || !(polyptr->flags & POLYDIR_CREATE))) {
 	pam_syslog(idata->pamh, LOG_ERR, "Polydir %s access error: %m",
 		polyptr->dir);
 	return PAM_SESSION_ERR;
```
ldv-alt, line 1660: The resulting code becomes somewhat confusing.  What do you think about the following change instead:
```diff
--- a/modules/pam_namespace/pam_namespace.c
+++ b/modules/pam_namespace/pam_namespace.c
@@ -1654,16 +1654,14 @@ static int ns_setup(struct polydir_s *polyptr,
 
     retval = protect_dir(polyptr->dir, 0, 0, idata);
 
-    if (retval < 0 && errno != ENOENT) {
-	pam_syslog(idata->pamh, LOG_ERR, "Polydir %s access error: %m",
-		polyptr->dir);
-	return PAM_SESSION_ERR;
-    }
-
     if (retval < 0) {
-	if ((polyptr->flags & POLYDIR_CREATE) &&
-		create_polydir(polyptr, idata) != PAM_SUCCESS)
-		return PAM_SESSION_ERR;
+	if (errno != ENOENT || !(polyptr->flags & POLYDIR_CREATE)) {
+	    pam_syslog(idata->pamh, LOG_ERR, "Polydir %s access error: %m",
+		       polyptr->dir);
+	    return PAM_SESSION_ERR;
+	}
+	if (create_polydir(polyptr, idata) != PAM_SUCCESS)
+	    return PAM_SESSION_ERR;
     } else {
 	close(retval);
     }
```

obp-anssi: I agree it's cleaner, and the logic stays the same. I'm changing my commit

ldv-alt (approved):

PR #886, opened 2025-02-25
pam_inline.h: introduce pam_asprintf(), pam_snprintf(), and pam_sprintf() helpers

ldv-alt:
pam_asprintf() is essentially asprintf() with the following semantic difference: it returns the string itself instead of its length.

pam_snprintf() is essentially snprintf() with the following semantic difference: it returns -1 in case of truncation.

pam_sprintf() is essentially snprintf() but with a check that the buffer is an array, and with an automatically calculated buffer size.

Use of these helpers would make error checking simpler.

### DIFF
--- a/libpam/include/pam_cc_compat.h
+++ b/libpam/include/pam_cc_compat.h
@@ -21,6 +21,12 @@
 # define PAM_ATTRIBUTE_ALIGNED(arg)	/* empty */
 #endif
 
+#if PAM_GNUC_PREREQ(3, 0)
+# define PAM_ATTRIBUTE_MALLOC		__attribute__((__malloc__))
+#else
+# define PAM_ATTRIBUTE_MALLOC		/* empty */
+#endif
+
 #if PAM_GNUC_PREREQ(4, 6)
 # define DIAG_PUSH_IGNORE_CAST_QUAL					\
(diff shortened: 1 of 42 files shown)


thkukuk:
Rewriting the code to use asprintf instead of calculating length + malloc + sprintf is a good idea.
But I'm not sure if this functions make error checking really simpler, I don't see that from the patches. But on a first glance, looks good to me.

ldv-alt:
> Rewriting the code to use asprintf instead of calculating length + malloc + sprintf is a good idea.

It's generally a good idea unless we're operating on the data that has to be securely erased because `asprintf` doesn't provide this guarantee.  In those few cases in this series where this rewriting is done, it's safe.

> But I'm not sure if this functions make error checking really simpler, I don't see that from the patches. But on a first glance, looks good to me.

With regards to the replacement of `asprintf` with `pam_asprintf`, it's definitely simpler in a few cases where otherwise we had to zero the pointer, and about the same (it's a matter of taste indeed) in other cases.

With regards to the replacement of `snprintf` with `pam_sprintf`, we don't have return value checks in most cases, but it's simpler in those cases where we have, and without this `sizeof(buf)` argument it's simpler anyway.

PR #903, opened 2025-06-08
man: fix SEE ALSO typo

dlejay:


ldv-alt:
Is there any rationale for changing the man section number?

dlejay:
> > Is there any rationale for changing the man section number?
> 
> I have seen distros who changed that, but that are downstream patches, not upstream.

I understand now... I found a discrepancy in the numbering of the manual on Debian 12.
I had no idea Debian was renumbering PAM from section 8 to section 7.

Issue #902, opened 2025-06-03
RFE: unix_chkpwd without setuid


poettering:
I am quite interested in making basic Linux userspace work without setuid binaries, so that we can set NoNewPrivs in PID 1 early on and remove an entire class of security issues. This works relatively well these days, but there are of course a couple of suid binaries in the basic OS that do matter, that are harder to get rid of.

One of them is unix_chkpwd. I was wondering if you'd be open to accept the following modifications in order to make it safe for operation in a setuid-less world:

1. in libpam, when forking out unix_chkpwd, first try to connect() to some AF_UNIX stream socket in /run/ somewhere. if that works, great! instead of talking to unix_chkpwd as a child via pipes, it should just talk to the socket then. If the socket doesn't work, just do what was usually done.

2. unix_chkpwd should be tweaked to allow execution as a socket activated service (i.e. in systemd Accept=yes style), so that it can be bound to the aforementioned socket path. This would require minor modifications on the protocol, i.e. the argv[2] parameter should probably move into the protocol in this case, as first line. and argv[1] should just be removed in this case. the existing getuid() based validation should be reworked to optionally use SO_PEERCRED on stdin instead. The test result should also be moved into the protocol. These modifications would be really minimal I'd claim: the essence of the tool would remain the same, there are just two distinct ways to talk to it, either via pipes as setuid child or via socket as IPC. But to a large degree both ways would share the same code.

Net result: if the IPC stuff exists it would be used and work. But there's also nice behaviour in environments where the IPC stuff doesn't exist: the logic would automatically degrade to using the classic way to do things.

Does that sound reasonable and acceptable to you? Opinions?

poettering:
Oh, love it. Varlink even! excellent!

PR #912, opened 2025-06-24
pam_unix: add support for pwaccessd (#902)

thkukuk:
[pwaccess](https://github.com/thkukuk/pwaccess) is a systemd socket activated service which provides passwd and shadow entries to root or the user of that entry to avoid having binaries owned by group shadow and/or setuid/setgid to do the authentication.
Currently it can replace `unix_chkpwd` and first PoC patches for `shadow` exists.

Reasons:
- avoid UID/GID shift with image based updates, so everything in /usr should be owned by root:root
- security, don't allow setuid/setgid binaries. Requested by:
  - systemd developers (https://github.com/linux-pam/linux-pam/issues/902)
  - our security team

The code is written in a way that `unix_ckpwd` is called as fallback if the service does not work.

`pwaccess` itself is currently under review of our security team.

### DIFF
--- a/modules/pam_unix/support.c
+++ b/modules/pam_unix/support.c
@@ -22,6 +22,9 @@
 #ifdef HAVE_NIS
 #include <rpcsvc/ypclnt.h>
 #endif
+#ifdef USE_PWACCESS
+#include <pwaccess.h>
+#endif
 
 #include <security/_pam_macros.h>
 #include <security/pam_modules.h>
@@ -515,6 +518,93 @@ int _unix_comesfromsource(pam_handle_t *pamh,
 #include <sys/types.h>
 #include <sys/wait.h>
 
+#ifdef USE_PWACCESS
+int _unix_pwaccess_check_expired(pam_handle_t *pamh, const char *user,
+				 long *daysleft)
+{
+    int  r, retval = PAM_SYSTEM_ERR;
+    bool pwchangeable = false;
+    long dl = -1;
+    char *error = NULL;
+
+    D(("called."));
+
+    r = pwaccess_check_expired(user, &dl, &pwchangeable, &error);
+    if (r < 0 || error) {
+        if (PWACCESS_IS_NOT_RUNNING(r)) {
+	    free(error);
+	    return PAM_SYSTEM_ERR;
+	}
+	if (r == -ENOENT) {
+	    free(error);
+	    return PAM_USER_UNKNOWN;
+	}
+	pam_syslog(pamh, LOG_ERR, "pwaccess failed: %s", error ? error : strerror(-r));
+	free(error);
+	return PAM_AUTHINFO_UNAVAIL;
+    }
+    *daysleft = dl;
+
+    switch(r) {
+        case PWA_EXPIRED_NO:
+	    if (!pwchangeable) {
+	        D(("password change too recent"));
+                return PAM_AUTHTOK_ERR;
+	    }
+	    retval = PAM_SUCCESS;
+	    break;
+        case PWA_EXPIRED_YES:
+            retval = PAM_ACCT_EXPIRED;
+            break;
+        case PWA_EXPIRED_CHANGE_PW:
+            retval = PAM_NEW_AUTHTOK_REQD;
+            break;
+        case PWA_EXPIRED_DISABLED:
+            retval = PAM_AUTHTOK_EXPIRED;
+            break;
+        default:
+	    retval = PAM_SYSTEM_ERR;
+    }
+
+    return retval;
+}
+
+static int _unix_pwaccess_verify_password(pam_handle_t *pamh, const char *passwd,
+			      unsigned long long ctrl, const char *user)
+{
+    int  r;
+    bool authenticated = false;
+    bool nullok = false;
+    char *error = NULL;
+
+    D(("called."));
+
+    if (off(UNIX__NONULL, ctrl))
+        nullok = true;
+
+
+    r = pwaccess_verify_password(user, passwd, nullok, &authenticated, &error);
+    if (r < 0 || error) {
+        if (PWACCESS_IS_NOT_RUNNING(r)) {
+	    free(error);
+	    return PAM_SYSTEM_ERR;
+	}
+	if (r == -ENOENT) {
+	    free(error);
+	    return PAM_USER_UNKNOWN;
+	}
+	pam_syslog(pamh, LOG_ERR, "pwaccess failed: %s", error ? error : strerror(-r));
+	free(error);
+	return PAM_AUTHINFO_UNAVAIL;
+    }
+
+    if (authenticated)
+	return PAM_SUCCESS;
+
+    return PAM_AUTH_ERR;
+}
+#endif
+
 static int _unix_run_helper_binary(pam_handle_t *pamh, const char *passwd,
 				   unsigned long long ctrl, const char *user)
 {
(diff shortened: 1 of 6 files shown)


modules/pam_unix/support.c
```diff
@@ -515,6 +518,89 @@ int _unix_comesfromsource(pam_handle_t *pamh,
...
+int _unix_verify_pwaccess(pam_handle_t *pamh, const char *user,
+			  int *daysleft)
+{
+    int  r, retval = PAM_SYSTEM_ERR;
+    bool pwchangeable = false;
+    long dl = -1;
+
+    D(("called."));
+
+    r = pwaccess_check_expired(user, &dl, &pwchangeable, NULL);
```
ldv-alt, line 531: Looks like `pwaccess_check_expired` can return a textual error. Is there any reason why this feature is not used here?

modules/pam_unix/support.c
```diff
@@ -515,6 +518,89 @@ int _unix_comesfromsource(pam_handle_t *pamh,
...
+    bool authenticated = false;
+    bool nullok = false;
+
+    D(("called."));
+
+    if (off(UNIX__NONULL, ctrl))
+        nullok = true;
+
+
+    r = pwaccess_verify_password(user, passwd, nullok, &authenticated, NULL);
```
ldv-alt, line 585: Looks like `pwaccess_verify_password` can return a textual error. Is there any reason why this feature is not used here?

thkukuk:
> This changes what pam_unix does. Shouldn't something like this be a different independent pam module? Then the Distro or user can choose at runtime if they actually want this. They could then also configure if they want to fall back to pam_unix or not. The way it is now, people will not get that choice.

pam_unix today executes a setuid helper application and communicates with it. This will be enhanced with communication with an already running service. I don't see how this changes what pam_unix does today.

A different, independent PAM module has one big advantage: we could clean up the code massive.
On the other side, a lot of functionality and thus code will stay identical, so you have a problem keeping that in sync. Another problem is the fallback if either pwaccessd is not running or you are inside a container: you have to setup some complicated rules in the pam.d service file to ignore errors of pwaccessd and use pam_unix or skip pam_unix if pwaccessd worked.
I don't think that this would be better.

About choice:
admin can disable the pwaccessd service if he does not want that. Or in the future he can disable setuid if he does not want that.
The big advantage of the current solution is, that an admin can still login as root if pwaccess get borked.

Daniel-Abrecht:
That would indeed address my concerns. I don't see any major drawbacks 
to this approach. It does seam a bit more complex though.
Also, I don't know if being able to use that binary without 
setuid/setgid could have any other side effects; aside from changing the 
user, setuid also makes the libc ignore certain environment variables on 
program start, and some other things. I don't know if that no longer 
happening could cause any issues or not.

ldv-alt:
> Also, I don't know if being able to use that binary without setuid/setgid could have any other side effects; aside from changing the user, setuid also makes the libc ignore certain environment variables on program start, and some other things. 

The helper could be made (artificially) setgid specifically to address this.

thkukuk:
> As for how this changes what pam_unix does: It changes the source of truth in regards to account information. It will always try to access the pwaccess service (or anything providing it) first. For now, that probably doesn't make much of a difference, since it really does just use the getpwnam / getspnam functions and so on, but that doesn't have to stay that way. For instance, there is a bit of overlap with systemd-userdb (which also uses varlink): https://github.com/thkukuk/pwaccess/blob/main/src/varlink-org.openSUSE.pwaccess.c#L25-L30 vs. https://systemd.io/USER_GROUP_API/#method-calls It would be conceivable that systemd-userdb eventually takes over the functionality currently provided by pwaccess, at which point nsswitch / shadow would no longer be the final source of truth. 

The source of truth is always coming from the same source, independent how many "caching" services like systemd-userdb you put in between. It's like as if you use nscd.

The last I heard that there are big concerns with systemd-userdb. And it cannot verify the password or if an account or password is expired.
But yes, could be very well that in some years libpwaccess will no longer query pwaccessd but userdb. But more likely is, that there will be a nss_userdb module, as changing all applications from using getpwent/getspent to something else will never happen.

> A completely different thing to consider is security. I believe this was part of the consideration for switching away from setuid in the first place. But if it falls back to the setuid binary anyway, there won't be any benefit, will there? In addition, the attack surface expands, as now bugs in the pwaccess service can now also introduce security issues. That said, this is just my opinion. I'm not a project maintainer, I don't really have a say in these matters anyway.

Comment from our security team: every ipc is more secure than any set*id binary.
If at least we at SUSE would enable pwaccess, then we will remove the setuid/setgid bits, so that unix_chkpwd will only be used inside containers.
So the security problem of setuid/setgid binaries don't exist anymore.
And you can secure every systemd service, in the case of pwaccessd you could give the service only access to /etc/passwd and /etc/shadow and nothing else.

thkukuk:
> Wouldn't it be better if libpwaccess functions were called from the helper instead of pam_unix.so? The advantage is that the helper could continue doing its job if it is privileged or just has an access to the shadow entry, and fall back to libpwaccess if it doesn't have the direct access. This way all extra code will be run in a separate process. Do you see any drawbacks in this approach?

With libpwaccess all extra code runs in a separate process, too, which can be locked down much more and is more secure than any setuid/setgid binary.
I don't understand the benefit of moving the varlink call into the helper binary?

ldv-alt:
> With libpwaccess all extra code runs in a separate process, too, which can be locked down much more and is more secure than any setuid/setgid binary.

When a binary is installed setgid, it doesn't necessarily mean that the binary is privileged.
setgid could be used as a hardening technique to have the executable run with the dumpable attribute cleared from the start.

> I don't understand the benefit of moving the varlink call into the helper binary?

Usually the less code runs in the main process the better.  This approach makes the cost of potential errors in that code cheaper, especially such errors as memory leaks and side effects of loading a library with dependencies.

thkukuk:
> > With libpwaccess all extra code runs in a separate process, too, which can be locked down much more and is more secure than any setuid/setgid binary.
> 
> When a binary is installed setgid, it doesn't necessarily mean that the binary is privileged. setgid could be used as a hardening technique to have the executable run with the dumpable attribute cleared from the start.

I discussed this at the openSUSE Conference last week, nobody sees how this can harden a simple IPC (varlink) call, the overall reaction was, adding an additional binary makes it only unnecessarily more complex and unsafe.

> > I don't understand the benefit of moving the varlink call into the helper binary?
> 
> Usually the less code runs in the main process the better. This approach makes the cost of potential errors in that code cheaper, especially such errors as memory leaks and side effects of loading a library with dependencies.

This only moves the problem from one binary to another one, and since we have additional fork() and execve() in between, only more complex and less robust.

modules/pam_unix/support.c
```diff
@@ -515,6 +518,103 @@ int _unix_comesfromsource(pam_handle_t *pamh,
...
+	    return PAM_SYSTEM_ERR;
+	}
+	if (r == -ENOENT) {
+	    free(error);
+	    return PAM_USER_UNKNOWN;
+	}
+	if (error)
+	    pam_syslog(pamh, LOG_ERR, "pwaccess failed: %s", error);
+	else
+	    pam_syslog(pamh, LOG_ERR, "pwaccess failed: %s", strerror(-r));
```
ldv-alt, line 545: `pam_syslog(pamh, LOG_ERR, "pwaccess failed: %s", error ? error : strerror(-r));`

modules/pam_unix/support.c
```diff
@@ -515,6 +518,103 @@ int _unix_comesfromsource(pam_handle_t *pamh,
...
```
ldv-alt, line 606: `pam_syslog(pamh, LOG_ERR, "pwaccess failed: %s", error ? error : strerror(-r));`

thkukuk:
I adjusted the patch for the long daysleft.
Our security team did an audit of pawaccess and they have no concerns.

ldv-alt:
@thkukuk, the commit message is rather terse, could you please update it to include the information from this PR?

thkukuk:
I created #929, which only replaces the `getspnam()` call, so that pam_unix behaves identical to today.
Using this varlink call will be better something for a complete new "pam_unix" module.

PR #915, opened 2025-07-01
pam_env: better syslog message for ignoring backslash at end of string

xBZZZZ:
bad line in `/etc/security/pam_env.conf`:
```
MYVAR DEFAULT="hello\"
```
syslog message before this commit:
```
pam_env(lightdm:setcred): Unrecognized escaped character: <
```
syslog message after this commit:
```
pam_env(lightdm:setcred): Ignoring backslash at end of string
```

### DIFF
--- a/modules/pam_env/pam_env.c
+++ b/modules/pam_env/pam_env.c
@@ -627,6 +627,11 @@ _expand_arg(pam_handle_t *pamh, char **value)
     if ('\\' == *orig) {
       ++orig;
       if ('$' != *orig && '@' != *orig && '\\' != *orig) {
+	if (!*orig) {
+	  D(("Ignoring backslash at end of string"));
+	  pam_syslog(pamh, LOG_ERR, "Ignoring backslash at end of string");
+	  break;
+	}
 	D(("Unrecognized escaped character: <%c> - ignoring", *orig));
 	pam_syslog(pamh, LOG_ERR,
 		   "Unrecognized escaped character: <%c> - ignoring",


modules/pam_env/pam_env.c
```diff
@@ -627,6 +627,11 @@ _expand_arg(pam_handle_t *pamh, char **value)
...
       ++orig;
       if ('$' != *orig && '@' != *orig && '\\' != *orig) {
+	if (!*orig) {
+	  D(("Ignoring backslash at end of string"));
+	  pam_syslog(pamh, LOG_ERR, "Ignoring backslash at end of string");
+	  break;
+	}
 	D(("Unrecognized escaped character: <%c> - ignoring", *orig));
 	pam_syslog(pamh, LOG_ERR,
 		   "Unrecognized escaped character: <%c> - ignoring",
```
ldv-alt, line 637: Let's keep it simple:
```
if (*orig) {
...
} else {
...
}
```
This way there would be no need for an extra `break` statement.

ldv-alt (changes requested):

modules/pam_env/pam_env.c
xBZZZZ: I don't know how is extra `else` and check `*orig` not zero again (`while` loop condition) better than `break` but ok.

ldv-alt: Thanks.
Avoiding that extra `break` keeps the change local and easy to read.
Avoiding the check in `while (*orig)` on error path is an optimization hardly worth doing.

PR #919, opened 2025-07-10
pam_unix: turn daysleft into long

stoeckmann:
The daysleft value could be larger than INT_MAX depending on system configuration (content of shadow file).

We already cover this case with a subtract function, but it's much easier to simply turn daysleft into a long. This simplifies the code and is even a no-op on 32 bit platforms (if subtract is optimized away).


thkukuk (approved):
Looks good to me.

ldv-alt (approved):

PR #929, opened 2025-08-15
pam_unix: add support for pwaccessd (#902)

thkukuk:
[pwaccess](https://github.com/thkukuk/pwaccess) is a systemd socket activated service which provides passwd and shadow entries to root or the user of that entry to avoid having binaries owned by group shadow and/or setuid/setgid to do the authentication.

Reasons:
- avoid UID/GID shift with image based updates, so everything in /usr should be owned by root:root
- security, don't allow setuid/setgid binaries. Requested by:
  - systemd developers (https://github.com/linux-pam/linux-pam/issues/902)
  - our security team

Since with #912 pam_unix could behave different than today, I decided to only replace the `getspnam()` call in unix_chkpwd, so that unix_chkpwd does not need the setuid bit if pwaccessd is running. If not, unix_chkpwd will fallback to `getspnam()`. 

`pwaccess` itself got reviewed by our security team.

### DIFF
--- a/modules/pam_unix/passverify.c
+++ b/modules/pam_unix/passverify.c
@@ -23,6 +23,9 @@
 #ifdef HAVE_CRYPT_H
 #include <crypt.h>
 #endif
+#ifdef USE_PWACCESS
+#include <pwaccess.h>
+#endif
 
 #include "pam_cc_compat.h"
 #include "pam_inline.h"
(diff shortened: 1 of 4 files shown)


modules/pam_unix/passverify.c
```diff
@@ -210,6 +213,26 @@ PAMH_ARG_DECL(int get_account_info,
...
+		char *error = NULL;
+
+		r = pwaccess_get_user_record(-1, name, NULL, spwdent, &complete, &error);
+		if (r < 0) {
+			if (!PWACCESS_IS_NOT_RUNNING(r)) {
+				if (error) {
+					pam_syslog(pamh, LOG_ERR, "%s", error);
+					free(error);
+				} else {
+					pam_syslog(pamh, LOG_ERR, "%s", strerror (-r));
```
ldv-alt, line 228: Extra whitespace between `strerror` and `(-r)`.

PR #939, opened 2025-09-09
Support the target and override_home options for pam_mkhomedir

### LINKED ISSUE #938: Option for pam_mkhomedir to create directories in an alternate location
In an enterprise setting it is common to use network home directories and directory services that result in common passwd maps across hosts.  For some subset of hosts and/or users, there may be a requirement to remove as many dependencies as possible including networked homes (a poor choice for an administrator trying to recover a critical piece of infrastructure).  In those cases pam_mkhomedir can be used to provide a local home directory for some users - but if the users's password entry points to a location managed by the automounter it is unable to do so.

It would therefore be useful for pam_mkhomedir to be able to create local home directories under an alternate non-autofs managed location and to potentially set the HOME environment variable to that local location.

jake-scott:
Fixes #938

The parent option forces pam_mkhomedir to create home directories under an alternative parent directory, and the override_home option causes the HOME environment variable to point to the managed home directory.

### DIFF
--- a/modules/pam_mkhomedir/mkhomedir_helper.c
+++ b/modules/pam_mkhomedir/mkhomedir_helper.c
@@ -388,7 +388,7 @@ create_homedir_helper(const struct passwd *_pwd, mode_t home_mode,
 }
 
 static int
-make_parent_dirs(char *dir, int make)
+make_parent_dirs(const char *dir, int make)
 {
   int rc = PAM_SUCCESS;
   char *cp = strrchr(dir, '/');
@@ -423,9 +423,10 @@ main(int argc, char *argv[])
    char *eptr;
    unsigned long home_mode = 0;
    const char *vendordir = NULL;
+   const char *target_dir = NULL;
 
    if (argc < 2) {
-	fprintf(stderr, "Usage: %s <username> [<umask> [<skeldir> [<home_mode>]]]\n", argv[0]);
+	fprintf(stderr, "Usage: %s <username> [<umask> [<skeldir> [<home_mode> [<target_dir>] [<vendordir>]]]]\n", argv[0]);
 	return PAM_SESSION_ERR;
    }
 
@@ -457,24 +458,30 @@ main(int argc, char *argv[])
        }
    }
 
-   if (argc >= 6)
-      vendordir = argv[5];
-
    if (home_mode == 0)
       home_mode = 0777 & ~u_mask;
 
-   if (pwd->pw_dir[0] != '/') {
+    if (argc >= 6) {
+      target_dir = argv[5];
+   } else {
+      target_dir = pwd->pw_dir;
+    }
+
+   if (target_dir[0] != '/') {
       pam_syslog(NULL, LOG_ERR, "Relative user home directory %s", pwd->pw_dir);
       return PAM_SESSION_ERR;
    }
 
+   if (argc >= 7)
+   vendordir = argv[6];
+
    /* Stat the home directory, if something exists then we assume it is
       correct and return a success */
-   if (stat(pwd->pw_dir, &st) == 0)
+   if (stat(target_dir, &st) == 0)
 	return PAM_SUCCESS;
 
-   if (make_parent_dirs(pwd->pw_dir, 0) != PAM_SUCCESS)
+   if (make_parent_dirs(target_dir, 0) != PAM_SUCCESS)
 	return PAM_PERM_DENIED;
 
-   return create_homedir_helper(pwd, home_mode, skeldir, pwd->pw_dir, vendordir);
+   return create_homedir_helper(pwd, home_mode, skeldir, target_dir, vendordir);
 }
--- a/modules/pam_mkhomedir/mkhomedir_helper.c
+++ b/modules/pam_mkhomedir/mkhomedir_helper.c
@@ -473,7 +473,7 @@ main(int argc, char *argv[])
    }
 
    if (argc >= 7)
-   vendordir = argv[6];
+      vendordir = argv[6];
 
    /* Stat the home directory, if something exists then we assume it is
       correct and return a success */
--- a/modules/pam_mkhomedir/pam_mkhomedir.c
+++ b/modules/pam_mkhomedir/pam_mkhomedir.c
@@ -54,9 +54,9 @@
 #include "pam_i18n.h"
 
 /* argument parsing */
-#define MKHOMEDIR_DEBUG         020	/* be verbose about things */
-#define MKHOMEDIR_QUIET         040	/* keep quiet about things */
-#define MKHOMEDIR_OVERRIDE_HOME 060 /* override the home directory */
+#define MKHOMEDIR_DEBUG         020	 /* be verbose about things */
+#define MKHOMEDIR_QUIET         040	 /* keep quiet about things */
+#define MKHOMEDIR_OVERRIDE_HOME 0100 /* override the home directory */
 
 #define LOGIN_DEFS           "/etc/login.defs"
 #define UMASK_DEFAULT        "0022"
@@ -166,7 +166,7 @@ create_homedir (pam_handle_t *pamh, options_t *opt,
    child = fork();
    if (child == 0) {
 	static char *envp[] = { NULL };
-	const char *args[] = { NULL, NULL, NULL, NULL, NULL, NULL, NULL };
+	const char *args[] = { NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL };
 
 	if (pam_modutil_sanitize_helper_fds(pamh, PAM_MODUTIL_PIPE_FD,
 					    PAM_MODUTIL_PIPE_FD,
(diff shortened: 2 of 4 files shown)


modules/pam_mkhomedir/pam_mkhomedir.c
```diff
@@ -54,8 +54,9 @@
 #include "pam_i18n.h"
 
 /* argument parsing */
-#define MKHOMEDIR_DEBUG      020	/* be verbose about things */
-#define MKHOMEDIR_QUIET      040	/* keep quiet about things */
+#define MKHOMEDIR_DEBUG         020	/* be verbose about things */
+#define MKHOMEDIR_QUIET         040	/* keep quiet about things */
+#define MKHOMEDIR_OVERRIDE_HOME 060 /* override the home directory */
```
ldv-alt, line 59: MKHOMEDIR_OVERRIDE_HOME == MKHOMEDIR_DEBUG | MKHOMEDIR_QUIET?

modules/pam_mkhomedir/pam_mkhomedir.c
```diff
@@ -172,8 +179,9 @@ create_homedir (pam_handle_t *pamh, options_t *opt,
 	args[2] = opt->umask ? opt->umask : UMASK_DEFAULT;
 	args[3] = opt->skeldir ? opt->skeldir : SKELDIR;
 	args[4] = login_homemode;
+	args[5] = dir;
 #ifdef VENDORDIR
-	args[5] = opt->skeldir ? NULL : VENDOR_SKELDIR;
+	args[6] = opt->skeldir ? NULL : VENDOR_SKELDIR;
```
ldv-alt, line 184: `args` has to be larger to do this.

modules/pam_mkhomedir/pam_mkhomedir.c
```diff
@@ -251,17 +261,70 @@ pam_sm_open_session (pam_handle_t *pamh, int flags, int argc,
...
+
+         /* Allocate memory for parent + "/" + username + null terminator */
+         homedir = malloc(parent_len + 1 + user_len + 1);
+         if (homedir == NULL) {
+            pam_syslog(pamh, LOG_ERR, "Memory allocation failed.");
+            return PAM_BUF_ERR;
+         }
+
+         /* Construct the path: parent/username */
+         snprintf(homedir, parent_len + 1 + user_len + 1, "%s/%s", opt.parent, (const char *)user);
```
ldv-alt, line 278: Please use `pam_asprintf` instead.

modules/pam_mkhomedir/pam_mkhomedir.c
```diff
@@ -251,17 +261,70 @@ pam_sm_open_session (pam_handle_t *pamh, int flags, int argc,
...
+   if (retval == PAM_SUCCESS && opt.ctrl & MKHOMEDIR_OVERRIDE_HOME) {
+      char *home_env;
+      size_t home_len = strlen("HOME=") + strlen(target_dir) + 1;
+
+      home_env = malloc(home_len);
+      if (home_env == NULL) {
+         pam_syslog(pamh, LOG_ERR, "Memory allocation failed for HOME environment variable");
+         retval = PAM_BUF_ERR;
+      } else {
+         snprintf(home_env, home_len, "HOME=%s", target_dir);
```
ldv-alt, line 312: Please use `pam_asprintf` instead.

modules/pam_mkhomedir/pam_mkhomedir.c
```diff
@@ -251,17 +261,70 @@ pam_sm_open_session (pam_handle_t *pamh, int flags, int argc,
...
+            pam_syslog(pamh, LOG_DEBUG, "Set HOME environment variable to %s", target_dir);
+         }
+         free(home_env);
       }
-      return PAM_SUCCESS;
    }
 
-   return create_homedir(pamh, &opt, user, pwd->pw_dir);
+   if (homedir != NULL) {
+      free(homedir);
```
ldv-alt, line 324: No need to check for NULL before calling `free`.

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -457,24 +458,30 @@ main(int argc, char *argv[])
...
+      target_dir = pwd->pw_dir;
+    }
+
+   if (target_dir[0] != '/') {
       pam_syslog(NULL, LOG_ERR, "Relative user home directory %s", pwd->pw_dir);
       return PAM_SESSION_ERR;
    }
 
+   if (argc >= 7)
+   vendordir = argv[6];
```
ldv-alt, line 476: Here something is wrong with the indentation.

ldv-alt (changes requested):

modules/pam_mkhomedir/pam_mkhomedir.c
jake-scott: D'oh - no - thanks!

modules/pam_mkhomedir/pam_mkhomedir.c
jake-scott: done

celbanramirez-del:
Hola

thedavidwhiteside:
upvote.  Would be a useful feature.

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -388,7 +388,7 @@ create_homedir_helper(const struct passwd *_pwd, mode_t home_mode,
 }
 
 static int
-make_parent_dirs(char *dir, int make)
+make_parent_dirs(const char *dir, int make)
```
ldv-alt, line 391: This changes `char *dir` parameter to `const char *dir`, but the function body still mutates the string via
   `*cp = '\0'` and `*cp = '/'`, introducing UB through const-qualified pointer modification.

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -457,24 +458,30 @@ main(int argc, char *argv[])
...
 
-   if (pwd->pw_dir[0] != '/') {
+    if (argc >= 6) {
+      target_dir = argv[5];
+   } else {
+      target_dir = pwd->pw_dir;
+    }
+
+   if (target_dir[0] != '/') {
       pam_syslog(NULL, LOG_ERR, "Relative user home directory %s", pwd->pw_dir);
```
ldv-alt, line 471: Error message logs `pwd->pw_dir` instead of `target_dir` when reporting a relative path, showing the wrong directory when target_dir comes from argv[5].

modules/pam_mkhomedir/mkhomedir_helper.c
```diff
@@ -423,9 +423,10 @@ main(int argc, char *argv[])
    char *eptr;
    unsigned long home_mode = 0;
    const char *vendordir = NULL;
+   const char *target_dir = NULL;
 
    if (argc < 2) {
-	fprintf(stderr, "Usage: %s <username> [<umask> [<skeldir> [<home_mode>]]]\n", argv[0]);
+	fprintf(stderr, "Usage: %s <username> [<umask> [<skeldir> [<home_mode> [<target_dir>] [<vendordir>]]]]\n", argv[0]);
```
ldv-alt, line 429: Please change the usage string to show proper nesting: `[<target_dir> [<vendordir>]]`

modules/pam_mkhomedir/pam_mkhomedir.c
```diff
@@ -251,17 +261,58 @@ pam_sm_open_session (pam_handle_t *pamh, int flags, int argc,
...
-              pwd->pw_dir);
+              target_dir);
+      }
+      retval = PAM_SUCCESS;
+   } else {
+      retval = create_homedir(pamh, &opt, user, target_dir);
+   }
+
+    /* Set HOME environment variable if override_home option is enabled */
+   if (retval == PAM_SUCCESS && opt.ctrl & MKHOMEDIR_OVERRIDE_HOME) {
```
ldv-alt, line 296: Let's add explicit parentheses for clarity:
`retval == PAM_SUCCESS && (opt.ctrl & MKHOMEDIR_OVERRIDE_HOME)`

modules/pam_mkhomedir/pam_mkhomedir.c
```diff
@@ -97,6 +100,10 @@ _pam_parse (const pam_handle_t *pamh, int flags, int argc, const char **argv,
 	 opt->umask = str;
       } else if ((str = pam_str_skip_prefix(*argv, "skel=")) != NULL) {
 	 opt->skeldir = str;
+      } else if ((str = pam_str_skip_prefix(*argv, "parent=")) != NULL) {
```
ldv-alt, line 103: New code for `parent=` and `override_home` option parsing uses spaces instead of tabs, violating the file's existing tab-based indentation convention.  Some of new lines use 4-space indentation while surrounding code uses tabs.  Some of new lines use inconsistent mixed indentation depths within the same if/else block.

ldv-alt (changes requested):
If I understand correctly, the `override_home` option sets HOME to `pwd->pw_dir` when used without `parent=`, which is redundant since that's already the user's home directory, so the feature is only useful when combined with `parent=`, but this constraint isn't documented or enforced.

PR #942, opened 2025-09-26
pam_nologin: don't display message to root

### LINKED ISSUE #941: `pam_nologin` shows "come back later" message for root
Attempting to ssh to a system as root when `/run/nologin` is present means that the login is successful but you get the "come back later message":

`"System is booting up. Unprivileged users are not permitted to log in yet. Please come back later. For technical details, see pam_nologin(8)."`

I'd expect that because root is permitted to login, they also wouldn't see the message.  I see this regularly pop up on the output of [`test.thing`](https://codeberg.org/lis/test.thing) and it's distracting and confusing.

allisonkarlitskaya:
Showing the "come back later" message to root is confusing.  Let's stop doing that.

Showing root the message (at the INFO level rather than ERR) seems to be an intended feature that has been around since at least the "Initial commit" in this repo in 2000, in this form:

```
       if (user_pwd && user_pwd->pw_uid == 0) {
         message.msg_style = PAM_TEXT_INFO;
```

But it's not clear to me if this was ever a good idea or why we would want it.  It certainly is at odds with the message that systemd writes to `/run/nologin` telling the user to "come back later", since root is indeed permitted to login successfully.

Fixes #941

### DIFF
--- a/modules/pam_nologin/pam_nologin.c
+++ b/modules/pam_nologin/pam_nologin.c
@@ -7,7 +7,6 @@
 #include "config.h"
 
 #include <limits.h>
-#include <stdio.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


ldv-alt:
Sure, feel free to close it.

PR #957, opened 2026-01-07
pam_unix: remove obsolete HAVE_YP_* to fix NIS

thkukuk:
If HAVE_NIS is defined we know that all required NIS functions are available, don't check for additional HAVE_YP_* defines, they got removed with the switch to meson.

(See #956 and https://bugzilla.redhat.com/show_bug.cgi?id=2363005)


imsedgar:
I'm also fine with this solution. It is even better. It also solves the problem.

Thanks for this update!

PR #981, opened 2026-05-19
pam_access: Fix matching of fully qualified usernames

### LINKED ISSUE #979: pam_access.so doesn't work with fully_qualified format username@domain in access.conf file & denies access.

asakure:
When **use_fully_qualified_names** is enabled in SSSD (**True**), usernames contain @ symbols (e.g., user@example.com). In such situations, `pam_access` was incorrectly parsing these as user@host patterns, causing access to fail when **user@domain** is defined in `/etc/security/access.conf` file.

This patch adds an exact string match check before attempting to parse @ as a user@host separator, fixing access control for fully qualified usernames (user@domain).

  - fully_qualified group-based matching (**group@domain**) already worked, username matching now works too.
  - Backward compatible with existing **user@host** pattern functionality

Verified with RHEL 8.10.

Fixes: [979](https://github.com/linux-pam/linux-pam/issues/979)

### DIFF
--- a/modules/pam_access/pam_access.c
+++ b/modules/pam_access/pam_access.c
@@ -675,6 +675,18 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
      * name of the user's primary group.
      */
 
+    /*
+     * Exact match for fully qualified username (user@domain) to prevent
+     * fully qualified usernames from being incorrectly parsed as user@host
+     * patterns.
+     */
+    if (strchr(string, '@') != NULL && strcasecmp(tok, string) == 0) {
+        if (item->debug)
+           pam_syslog (pamh, LOG_DEBUG,
+                       "user_match: exact match for fully qualified username '%s'", string);
+        return YES;
+    }
+
     /* Try to split on a pattern (@*[^@]+)(@+.*) */
     for (at = tok; *at == '@'; ++at);
 


modules/pam_access/pam_access.c
```diff
@@ -676,6 +676,10 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
      */
 
     /* Try to split on a pattern (@*[^@]+)(@+.*) */
+	/* Exact match for fully qualified username (user@domain) */
+    if (strchr(string, '@') != NULL && strcmp(tok, string) == 0)
```
ikerexxe: Consider replacing `strcmp()` by `strcasecmp()`. Domain name capitalization can vary within Active Directory environments; ensuring consistency helps prevent potential issues.

modules/pam_access/pam_access.c
```diff
@@ -676,6 +676,10 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
      */
 
     /* Try to split on a pattern (@*[^@]+)(@+.*) */
+	/* Exact match for fully qualified username (user@domain) */
```
ikerexxe: Indentation issue.

In addition, you could be a little bit more informative:
```
Exact match for fully qualified username (user@domain) to prevent
Fully Qualified Usernames from being incorrectly parsed as user@host
patterns.
```

modules/pam_access/pam_access.c
```diff
@@ -676,6 +676,10 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
      */
 
     /* Try to split on a pattern (@*[^@]+)(@+.*) */
```
ikerexxe, line 690: The new code and comment should happen before this comment as it breaks its context

modules/pam_access/pam_access.c
```diff
@@ -676,6 +676,10 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
...
+	/* Exact match for fully qualified username (user@domain) */
+    if (strchr(string, '@') != NULL && strcmp(tok, string) == 0)
+        return YES;
```
ikerexxe, line 687: Add debug logging before this line to have some consistency

ikerexxe (commented):
I tested this solution and it seems to work correctly, but there are some things to improve

modules/pam_access/pam_access.c
asakure: > Add debug logging before this line to have some consistency

I tried, please check if I did it correctly. TIA!

modules/pam_access/pam_access.c
```diff
@@ -675,6 +675,17 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
...
 
+    /* Exact match for fully qualified username (user@domain) to prevent
+     * fully qualified usernames from being incorrectly parsed as user@host
+     * patterns. */
+    if (strchr(string, '@') != NULL && strcasecmp(tok, string) == 0)
+    {
+        if (item->debug)
+           pam_syslog (pamh, LOG_DEBUG,
+                       "user_match: tok=%s, item=%s", tok, string);
+        retrun YES;
```
ikerexxe: s/retrun/return

modules/pam_access/pam_access.c
```diff
@@ -675,6 +675,17 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
      * name of the user's primary group.
      */
 
+    /* Exact match for fully qualified username (user@domain) to prevent
+     * fully qualified usernames from being incorrectly parsed as user@host
+     * patterns. */
+    if (strchr(string, '@') != NULL && strcasecmp(tok, string) == 0)
+    {
```
ikerexxe: Inconsisten brace style. Move this to the end of the previous line to match the project style

modules/pam_access/pam_access.c
```diff
@@ -675,6 +675,17 @@ user_match (pam_handle_t *pamh, char *tok, struct login_info *item)
...
+        if (item->debug)
+           pam_syslog (pamh, LOG_DEBUG,
+                       "user_match: tok=%s, item=%s", tok, string);
```
ikerexxe: I'd be a bit more informative here and state that this is an `exact match for fully qualified usernames`

ikerexxe (commented):
Some minor comments inline

ikerexxe (approved):
LGTM!

@ldv-alt do you mind taking a look at this?

ldv-alt (approved):

PR #985, opened 2026-06-02
pam_succeed_if: prevent logging unknown user names in plaintext

### LINKED ISSUE #559: pam_succeed_if logs invalid user names
Hi.

Ubuntu ships Lightdm with the following bit of PAM configuration:
```
auth    sufficient      pam_succeed_if.so user ingroup nopasswdlogin
```
And I get in the logs:
```
lightdm: pam_succeed_if(lightdm:auth): requirement "user ingroup nopasswdlogin" not met by user "XXXXXXXX"
```
But XXXXXXXX was my password, not my login: I typed my password instead of my login, and now thanks to `pam_succeed_if`, my password is in plaintext in the logs.

Typing a password instead of a login is a common mistake, this is why it is a recommended practice to not log a login unless it matches an actual system user. `pam_succeed_if` does that:

```
                                if(!quiet_fail && ret != PAM_USER_UNKNOWN)
                                        pam_syslog(pamh, LOG_INFO,
                                                   "requirement \"%s %s %s\" "
                                                   "not met by user \"%s\"",
                                                   left, qual, right, user);
```

Unfortunately, for the `ingroup` qualifier it misses, because `pam_modutil_user_in_group_common` returns 0 if `pwd` is NULL, not `PAM_USER_UNKNOWN`.

Thanks.

lzwind:
## Summary

- When a user accidentally types their password at the username prompt, `pam_succeed_if` logs the password in plaintext via `pam_syslog`. This happens because `evaluate_ingroup`, `evaluate_notingroup`, `evaluate_innetgr`, and `evaluate_notinnetgr` return `PAM_AUTH_ERR` or `PAM_SUCCESS` for non-existent users instead of `PAM_USER_UNKNOWN`, bypassing the existing log guard (`ret != PAM_USER_UNKNOWN`).
- Instead of changing the return values of the evaluate_* functions (which would break the expected semantics, e.g., `evaluate_notingroup` should return `PAM_SUCCESS` when the user is not in the group), this fix modifies the logging part: check whether the user exists using `pam_modutil_getpwnam()`. If the user does not exist and `audit` is not enabled, print "unknown user" instead of the potentially sensitive user input.

## Changes

| File | Change |
|------|--------|
| `modules/pam_succeed_if/pam_succeed_if.c` | After `evaluate()`, check if user exists via `pam_modutil_getpwnam()`. Use `"unknown user"` in log messages when user does not exist and audit is disabled. |

## Test plan

- [ ] Verify that when a non-existent username (e.g., a mistyped password) is passed to `pam_succeed_if.so user ingroup <group>`, logs show "unknown user" instead of the password
- [ ] Verify that existing valid user checks still work correctly (user name appears in logs as before)
- [ ] Verify that `ingroup`, `notingroup`, `innetgr`, `notinnetgr` qualifiers all behave correctly with both existing and non-existing users
- [ ] Verify that when `audit` flag is set, the actual user input is still logged for debugging purposes

Fixes: #559

### DIFF
--- a/modules/pam_succeed_if/pam_succeed_if.c
+++ b/modules/pam_succeed_if/pam_succeed_if.c
@@ -437,6 +437,28 @@ evaluate(pam_handle_t *pamh, int debug,
 	return PAM_SERVICE_ERR;
 }
 
+static void
+log_requirement(pam_handle_t *pamh,
+		struct passwd **pwd,
+		const char *resolution,
+		const char *left,
+		const char *qual,
+		const char *right,
+		const char *user,
+		int check_user)
+{
+	if (check_user && !*pwd && !(*pwd = pam_modutil_getpwnam(pamh, user))) {
+		pam_syslog(pamh, LOG_INFO,
+			   "requirement \"%s %s %s\" "
+			   "%s by unknown user",
+			   left, qual, right, resolution);
+	} else {
+		pam_syslog(pamh, LOG_INFO,
+			   "requirement \"%s %s %s\" "
+			   "%s by user \"%s\"",
+			   left, qual, right, resolution, user);
+	}
+}
 static int
 pam_succeed_if(pam_handle_t *pamh, int argc, const char **argv)
 {
@@ -539,19 +561,17 @@ pam_succeed_if(pam_handle_t *pamh, int argc, const char **argv)
 					   user);
 			if (ret != PAM_SUCCESS) {
 				if(!quiet_fail && ret != PAM_USER_UNKNOWN)
-					pam_syslog(pamh, LOG_INFO,
-						   "requirement \"%s %s %s\" "
-						   "not met by user \"%s\"",
-						   left, qual, right, user);
+					log_requirement(pamh, &pwd, "not met",
+							left, qual, right, user,
+							!(audit || use_uid));
 				left = qual = right = NULL;
 				break;
 			}
 			else
 				if(!quiet_succ)
-					pam_syslog(pamh, LOG_INFO,
-						   "requirement \"%s %s %s\" "
-						   "was met by user \"%s\"",
-						   left, qual, right, user);
+					log_requirement(pamh, &pwd, "was met",
+							left, qual, right, user,
+							!(audit || use_uid));
 			left = qual = right = NULL;
 			continue;
 		}


modules/pam_succeed_if/pam_succeed_if.c
```diff
@@ -239,10 +246,17 @@ evaluate_notingroup(pam_handle_t *pamh, const char *user, const char *grouplist)
...
 
 	if (group == NULL)
 		return PAM_BUF_ERR;
 
+	pwd = pam_modutil_getpwnam(pamh, user);
+	if (pwd == NULL) {
+		free(group);
+		return PAM_USER_UNKNOWN;
+	}
+
```
ldv-alt: This would break the behavior as `evaluate_notingroup` is expected to return `PAM_SUCCESS` when the user is not in the list.

modules/pam_succeed_if/pam_succeed_if.c
```diff
@@ -279,6 +299,12 @@ static int
 evaluate_notinnetgr(const pam_handle_t* pamh SOMETIMES_UNUSED, const char *host, const char *user, const char *group)
 {
 #ifdef HAVE_INNETGR
+	struct passwd *pwd;
+
+	pwd = pam_modutil_getpwnam((pam_handle_t *)pamh, user);
+	if (pwd == NULL)
+		return PAM_USER_UNKNOWN;
```
ldv-alt: Likewise.

ldv-alt (changes requested):
Looks like the chosen approach is not correct.  Instead, I suggest changing the logging part where the user name is printed:  if `audit` is enabled or the user exists, print the user name as it is printed now, otherwise print an unknown user.

modules/pam_succeed_if/pam_succeed_if.c
lzwind: done

modules/pam_succeed_if/pam_succeed_if.c
lzwind: done

ldv-alt:
I don't see any changes in the new edition of this commit.

lzwind:
Thank you for the suggestion. I've updated the patch to follow your approach: instead of modifying the evaluate_* functions' return values, the fix now only changes the logging part in pam_succeed_if(). After evaluate() returns, if pwd is still NULL, we call pam_modutil_getpwnam() to check whether the user exists. When the user doesn't exist and audit is not enabled, the log message prints 'unknown user' instead of the potentially sensitive user input.

The evaluate_ingroup, evaluate_notingroup, evaluate_innetgr, and evaluate_notinnetgr functions remain unchanged, so their original return value semantics (e.g. evaluate_notingroup returning PAM_SUCCESS when the user is not in the group) are preserved.

lzwind (commented):
Updated per review feedback: moved the user existence check from evaluate_* functions to the logging section. The evaluate functions' return value semantics are preserved.

ldv-alt (commented):
@lzwind: I've changed the patch a bit, please have a look.

lzwind:
Thank you for your assistance; it has been verified that this resolves the issue.

ldv-alt (approved):

PR #994, opened 2026-06-19
pam_namespace: replacing the fork/exec of /bin/rm with in-process fd-relative deletion

rawrmonster17:
## Problem

`cleanup_tmpdirs()` removes polyinstantiated TMPDIR instance directories
when a PAM session is closed.  The existing implementation has a
check-to-use (TOCTOU) race (CWE-367):

```c
/* Step 1 — CHECK: does the instance directory exist? */
faccessat(dfd, pptr->instname, F_OK, AT_SYMLINK_NOFOLLOW);

/* ← TOCTOU WINDOW: attacker replaces instname with a symlink to /etc */

/* Step 2 — USE: delete it (rm follows top-level symlinks) */
pid = fork();
execle("/bin/rm", "/bin/rm", "-rf", pptr->instname, ...);
```

Between the `faccessat()` existence check and the `execle("/bin/rm -rf")`,
a local user can atomically replace their instance directory with a
symlink to any target (e.g. `/etc`, `/var/log`).  Because `rm -rf`
follows symlinks at its top-level argument, the child process — running
as root — deletes the symlink target instead.

- **CWE-367** (TOCTOU), **CWE-59** (improper link resolution)
- **CVSS v3.1**: AV:L/AC:H/PR:L/UI:N/S:U/C:N/I:H/A:H → **6.3 MEDIUM**
- Reported: #972
- Related: CVE-2025-6020 hardened the rest of `pam_namespace.c` with
  `openat(O_NOFOLLOW)` + `unlinkat`; this is the one remaining path
  that was not converted.

## Fix

Replace the `faccessat` / `fork` / `execle` approach with an
**in-process recursive deletion** using only fd-relative syscalls.
No check-to-use window exists because the directory fd is acquired
atomically with `O_NOFOLLOW` and all subsequent operations are
relative to that held fd — no path strings are re-resolved after the
initial open.

Two new static helpers are added:

### `rm_dir_contents(pamh, dir_fd, depth)`

Recursively empties the directory referred to by `dir_fd` (which it
takes ownership of via `fdopendir`).  For every entry:

- `fstatat(AT_SYMLINK_NOFOLLOW)` determines the type without following
  links.
- Subdirectories are opened with `openat(O_NOFOLLOW | O_DIRECTORY)`.
  If the entry was swapped to a symlink between `readdir()` and
  `openat()`, `openat()` returns `ELOOP`/`ENOTDIR`; the code falls
  through to `unlinkat(..., 0)` which removes the symlink itself
  without following it.
- All other entries (regular files, symlinks, specials) are removed
  with `unlinkat(..., 0)` — never following a link.
- `errno` is saved and restored around `pam_syslog()` calls so that
  the post-loop `errno != 0` readdir-error check is not confused by
  the logging path.
- A depth guard (`TMPDIR_CLEANUP_MAX_DEPTH = 64`) prevents fd
  exhaustion on a pathologically deep tree created by a malicious user.

### `rm_instance_dir(pamh, parent_fd, name)`

Opens the instance directory with
`O_RDONLY | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC` relative to the
already-trusted `parent_fd`.  If `name` has been replaced by a
symlink, `openat()` fails with `ELOOP`/`ENOTDIR` and the function
returns an error **without touching anything outside the instance
parent**.  On success, calls `rm_dir_contents()` then
`unlinkat(parent_fd, name, AT_REMOVEDIR)`.

### Modified `cleanup_tmpdirs()`

The `faccessat` + `fork` + `execle` + `waitpid` block is replaced by
a single call to `rm_instance_dir()`.  The now-vestigial
`sigaction(SIGCHLD, ...)` setup/restore (only needed for `waitpid`) is
also removed.

## Security properties

| Property | Before | After |
|---|---|---|
| Symlink follow on top-level arg | Yes (`rm -rf`) | **No** (`O_NOFOLLOW`) |
| Check-to-use window | Yes (faccessat → execle) | **No** (single openat) |
| Symlink follow inside tree | Yes (`rm -rf`) | **No** (`O_NOFOLLOW` + `AT_SYMLINK_NOFOLLOW` at every level) |
| External binary (privilege surface) | Yes (`/bin/rm` as root) | **No** (in-process) |
| fd exhaustion DoS | N/A | Depth-capped at 64 |

The fix uses the identical `openat(O_NOFOLLOW)` + `unlinkat` pattern
that the CVE-2025-6020 hardening already applied to directory creation
and mounting in the same file (see `create_instance()` at line 1802).

## Testing

- Builds clean with `-Wall -Wextra -Wshadow -Wstrict-prototypes
  -Wmissing-prototypes -Wformat=2`: zero new warnings.
- All 78 existing PAM tests pass (`ninja test`).
- Functional tests verified: tree with regular files, symlinks (inner),
  nested subdirectories, and a symlink-replacing-directory attack
  scenario all behave correctly.  `/tmp` (symlink target in attack
  scenario) is untouched; the symlink itself is removed.

Fixes: #972

### DIFF
--- a/modules/pam_namespace/pam_namespace.c
+++ b/modules/pam_namespace/pam_namespace.c
@@ -2089,20 +2089,182 @@ static int cwd_in(char *dir, struct instance_data *idata)
     return retval;
 }
 
+/*
+ * Maximum directory recursion depth for rm_instance_dir().  A polyinstantiated
+ * /tmp instance should never be this deep, but we cap it to avoid exhausting
+ * file descriptors on a pathologically deep tree created by a malicious user.
+ */
+#define TMPDIR_CLEANUP_MAX_DEPTH 64
+
+/*
+ * Recursively remove all entries inside the directory referred to by dir_fd.
+ * Symlinks are unlinked without following.  Regular files and other non-
+ * directory entries are unlinked directly.  Subdirectories are opened with
+ * O_NOFOLLOW before recursing, so a symlink-to-directory swap between
+ * readdir() and openat() is harmless: openat() will fail with ENOTDIR/ELOOP
+ * and we will fall through to unlinkat() which removes the symlink itself.
+ *
+ * Returns 0 on success, -1 if any entry could not be removed (caller should
+ * log; we continue best-effort to remove as much as possible).
+ *
+ * Note: fdopendir() takes ownership of dir_fd; do NOT close(dir_fd) after
+ * a successful fdopendir() call.  closedir() releases the fd.
+ */
+static int
+rm_dir_contents(const pam_handle_t *pamh, int dir_fd, unsigned int depth)
+{
+    DIR *dirp;
+    struct dirent *ent;
+    struct stat st;
+    int child_fd;
+    int ret = 0;
+
+    if (depth > TMPDIR_CLEANUP_MAX_DEPTH) {
+        pam_syslog(pamh, LOG_ERR,
+                   "rm_dir_contents: recursion depth limit reached, "
+                   "instance directory not fully removed");
+        close(dir_fd);
+        return -1;
+    }
+
+    dirp = fdopendir(dir_fd);
+    if (dirp == NULL) {
+        pam_syslog(pamh, LOG_ERR,
+                   "rm_dir_contents: fdopendir failed: %m");
+        close(dir_fd);
+        return -1;
+    }
+    /* dir_fd is now owned by dirp; do not close(dir_fd) separately. */
+
+    errno = 0;
+    while ((ent = readdir(dirp)) != NULL) {
+        if (strcmp(ent->d_name, ".") == 0 || strcmp(ent->d_name, "..") == 0)
+            continue;
+
+        /*
+         * Use AT_SYMLINK_NOFOLLOW so we always operate on the entry itself,
+         * never on what a symlink points to.
+         */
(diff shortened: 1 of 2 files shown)


ldv-alt (changes requested):
Thank you for looking into pam_namespace cleanup paths.

The attack scenario as described does not work, however: rm(1)
does not follow symbolic links.  POSIX specifies that "the rm utility
shall not traverse directories by following symbolic links … but shall
remove the links themselves" (see the rm(1) RATIONALE).  GNU coreutils
rm enforces this with an FTS_PHYSICAL walk.  So even if the race is
won and the instance directory is replaced by a symlink, `rm -rf
instname` will unlink the symlink itself without touching the target.

Additionally, the instance parent directory is verified by
check_inst_parent() to be mode 000 owned by root, so an unprivileged
user cannot create or rename entries in it — the symlink swap cannot
be performed in the first place.

That said, replacing the fork/exec of /bin/rm with in-process fd-relative
deletion could be a worthwhile cleanup: it removes the external binary
dependency, avoids the SIGCHLD handler manipulation, and follows the
openat/O_NOFOLLOW/unlinkat pattern already used elsewhere in
pam_namespace.c.  We would be happy to consider taking a patch along those lines
as a code-quality improvement, but it should not be characterised as
a security fix, and the CVE/CWE claims should be withdrawn.

PR #1001, opened 2026-07-13
libpam: fail when included config file is empty

joantolo:
When a config file referenced by include or substack exists but contains no entries, _pam_parse_conf_file() returned PAM_SUCCESS without adding any handlers. This can happen when system tools or updates generate empty PAM service files, effectively bypassing authentication since no auth modules are loaded and subsequent optional modules succeed trivially. From a security perspective, the safe default for an empty included config is to deny rather than implicitly succeed.

Track the number of config entries seen across all module types. When called from an include/substack context (include_level > 0) and no entries were found, return PAM_SERVICE_ERR so the caller installs a PAM_HT_MUST_FAIL handler, matching the existing behavior for missing config files.

Only truly empty configs trigger the error; configs with entries for other module types (e.g. postlogin with only session entries included for auth) are not affected.


thkukuk:
I'm afraid this can break a lot of configs out there, with the result that even root can no longer login.

joantolo:
And would it be possible to add some option to ensure the substack/include is "not-empty" ?

PR #1004, opened 2026-07-22
pam_time: ignore rules with malformed day specification

### LINKED ISSUE #210: pam_time: Rule with malformed time field behaves like valid blanket rule.
To reproduce:
1. Open a root terminal to avoid locking yourself out..
2. Add pam_time.so to your pam configuration, with no options
3. Add the following line(as an example) to /etc/security/time.conf
```
*; *; *;oogabooga-1234-3423
```
4. try to log in or use sudo or something

You will get access denied. If you added debug to the pam_time options there will be a parsing error in the log.

I've read through the code, and the issue is in the function check_time. If a parsing error happens(not sure why parsing is even happening in this function; parsing the rule and applying it should ideally be decoupled), it just prints an error to syslog and returns FALSE. This doesn't match what happens if the whole rule is malformed. Then it logs an error, skips any further matching and permits the user to log in.

Shouldn't all malformed rules be ignored?

wyu71:
check_time() returned FALSE for a bad or missing day token, causing
pam_time to deny access for an otherwise matching rule.  This was
inconsistent with the handling of other malformed time fields, which
are logged and ignored.  Return TRUE instead so the rule is skipped
as documented.

Resolves: https://github.com/linux-pam/linux-pam/issues/210


modules/pam_time/tst-pam_time-retval.c
```diff
@@ -99,6 +99,38 @@ main(void)
...
+		  fprintf(fp, "#%%PAM-1.0\n"
+			      "auth required %s/" LTDIR "%s.so conffile=%s\n"
+			      "account required %s/" LTDIR "%s.so conffile=%s\n"
+			      "password required %s/" LTDIR "%s.so conffile=%s\n"
+			      "session required %s/" LTDIR "%s.so conffile=%s\n",
+			  cwd, MODULE_NAME, config_file,
+			  cwd, MODULE_NAME, config_file,
+			  cwd, MODULE_NAME, config_file,
+			  cwd, MODULE_NAME, config_file));
+	ASSERT_EQ(0, fclose(fp));
```
ldv-alt: Is this hunk really needed?

modules/pam_time/tst-pam_time-retval.c
```diff
@@ -99,6 +99,38 @@ main(void)
 	ASSERT_EQ(PAM_SUCCESS, pam_end(pamh, 0));
 	pamh = NULL;
 
+	/* malformed time field: rule should be ignored, not deny access */
+	ASSERT_NE(NULL, fp = fopen(config_file, "w"));
+	ASSERT_LT(0, fprintf(fp, "# malformed time field rule\n"
+				 "%s ; * ; root ; oogabooga-1234-3423\n",
+			     service_file));
```
ldv-alt: Consider "%s ; * ; root ; !oogabooga-1234-3423\n",
should this line be ignored as well?

ldv-alt (changes requested):

modules/pam_time/tst-pam_time-retval.c
wyu71: Removed this hunk. The service file already points to config_file, so rewriting it here was unnecessary.

modules/pam_time/tst-pam_time-retval.c
wyu71: Yes, it should be ignored as well. Added a regression case covering the same malformed day specification with the ! prefix.

ldv-alt (changes requested):
The fix is correct - all three check_time() error paths now consistently return FIELD_IGNORE, and logic_field() handles it before applying the negation operator, so both malformed and !malformed time fields are properly ignored.

There is one remaining inconsistency at the logic_field() level: its own garbled-syntax error paths still return FALSE. For the service/tty/user fields this is safe - FALSE means "doesn't match", so good becomes FALSE and the rule can't deny. But for the time field, FALSE means intime = FALSE, which makes the denial condition good && !intime true, so a garbled time expression like Al0000-2400& (trailing operator) would still deny access instead of being ignored. This is the same class of bug, just at the field-expression parsing layer rather than the time-value parsing layer.

One way to fix this would be to return FIELD_IGNORE from the garbled-syntax paths in logic_field() as well, and handle it in check_account().
For the service/tty/user fields, the same treatment would make logic_field() consistent across all callers.

On test coverage: the new test exercises bad day tokens with and without !, which is the key case. It would be good to also cover the other two error paths — a time field with no day prefix (e.g., 0000-2400, triggering marked_day == 0) and a valid day with a missing time range (e.g., Al, triggering the i != 5 check) — to lock in the consistent FIELD_IGNORE behaviour across all three paths.

wyu71:
Addressed the remaining field-expression parsing issue.  logic_field() now returns FIELD_IGNORE for garbled syntax, and check_account() skips rules with ignored fields.  I also added coverage for missing day specifications, missing time ranges, and a garbled logical expression.

ldv-alt (approved):
I've fixed remaining issues and squashed the changes into a single commit.
Please have a look.

PR #1006, opened 2026-07-28
pam_unix: make SHA512 the default

thkukuk:
If no hash is specified as option and ENCRYPT_METHOD is not specified in login.defs, pam_unix.so fails back to DES. Change this to a modern, secure hash algorithm (SHA512).


ikerexxe (commented):
There are still some places where references to weak cryptographic algorithms should be updated:
* `passverify.c` lines 474-488 still defaults to BIGCRYPT. This should be updated to either error out, or force SHA512 as the default.
* `support.h` line 100 still states DES is the default algorithm. This should be updated.

alejandro-colomar:
Thanks!

FWIW, shadow-utils has changed the default to SHA512 too.  This change will be included in the next release, which will likely happen this week.

-  <https://github.com/shadow-maint/shadow/issues/1451>

modules/pam_unix/pam_unix.8.xml
```diff
@@ -87,7 +87,9 @@
       The password component of this module performs the task of updating
       the user's password. The default encryption hash is taken from the
       <emphasis remap="B">ENCRYPT_METHOD</emphasis> variable from
-      <emphasis>/etc/login.defs</emphasis>
+      <emphasis>login.defs</emphasis>. If that variable is absent
+      and no hash algorithm option is passed to the module, SHA512 is
+      used as the built-in default.
```
alejandro-colomar, line 92: The inter-sentence space should be double in source code.  groff(1), which is invoked by man(1), interprets a double space after period as the end of a sentence, while it interprets a single space after period as if it were something like initials (so, not ending the sentence).  This results in different formatting.

In general, you should also use two spaces after period outside of source code, but that's another topic.  In any case, here's a link for reference:
<https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/CONTRIBUTING.d/patches/description#n44>,
which refers to some history about this:
<https://web.archive.org/web/20171217060354/http://www.heracliteanriver.com/?p=324>.

thkukuk:
> * `passverify.c` lines 474-488 still defaults to BIGCRYPT. This should be updated to either error out, or force SHA512 as the default.

The ctrl variable is initialized with SHA512. So in the default case, where no encrypt method is specified, line 472 should match. Line 474-488 should only be reached if somebody specifies bigcrypt as hash algorithm.

One of the next steps could be to remove support for bigcrypt/DES completely, but I would do that with an extra PR.

modules/pam_unix/pam_unix.8.xml
thkukuk: This is xml code. It doesn't matter how many spaces you add there, the resulting code groff see's is always the same and contains only one space.

alejandro-colomar:
> One of the next steps could be to remove support for bigcrypt/DES completely,

Yes, please.

> but I would do that with an extra PR.

And yes, please.  :)

modules/pam_unix/pam_unix.8.xml
alejandro-colomar: If the XML-to-man translator doesn't respect the double spacing, that's a bug in the translator.  Even if the translator is buggy, I think you should do it correctly, so that whenever the translator applies a bugfix you immediately get better manual pages.

alejandro-colomar (approved):
```
Suggested-by: Alejandro Colomar <alx@kernel.org>
```

ikerexxe (approved):
LGTM! Thanks for putting this together.

PR #1013, opened 2026-08-05
pam_selinux and pam_sepermit tweaks

cgzones:
See individual commits for more details.


modules/pam_selinux/pam_selinux.c
```diff
@@ -633,11 +634,13 @@ set_context(pam_handle_t *pamh, const module_data_t *data,
...
+    char *msg;
 
-    pam_sprintf(msg,
-	     _("Security context %s has been assigned."), data->exec_context);
-    send_text(pamh, msg, debug);
+    if ((msg = pam_asprintf(_("Security context %s has been assigned."),
+			    data->exec_context)) != NULL) {
+      send_text(pamh, msg, debug);
+      free(msg);
+    }
```
ldv-alt, line 643: Wouldn't it be better to open-code send_text() instead?

ldv-alt (approved):

modules/pam_sepermit/pam_sepermit.c
```diff
@@ -425,9 +425,11 @@ pam_sepermit(pam_handle_t *pamh, int argc, const char **argv)
 	}
 
 	if (getseuserbyname(user, &seuser, &level) != 0) {
+		pam_syslog(pamh, LOG_ERR, "getseuserbyname failed for user %s: %m", user);
+		free(seuser);
+		free(level);
```
ldv-alt, line 430: The same issue seems to be present in pam_selinux.c: compute_exec_context() leaks seuser/level when getseuserbyname() fails.

modules/pam_selinux/pam_selinux.c
```diff
@@ -748,6 +753,21 @@ pam_sm_setcred(pam_handle_t *pamh UNUSED, int flags UNUSED,
...
+    "nottys", "select_context", "use_current_range", "env_params",
+  };
+  size_t i;
+
+  for (i = 0; i < PAM_ARRAY_SIZE(options); i++)
+    if (strcmp(opt, options[i]) == 0)
+      return 1;
+  return 0;
+}
+
```
ldv-alt, line 770: This creates a second source of truth for option names, diverging from the if/else-if pattern used by all other PAM modules.

ldv-alt (changes requested):

modules/pam_selinux/pam_selinux.c
cgzones: refactored `send_text()`

modules/pam_sepermit/pam_sepermit.c
cgzones: `getseuserbyname(3)` only sets `seuser` and `level` on success (i.e. return value `0`), so dropped the unnecessary cleanup entirely.

modules/pam_selinux/pam_selinux.c
cgzones: refactored into a common `parse_options()` helper

ldv-alt (approved):

PR #1020, opened 2026-09-07
Re-add check for crypt_checksalt

thkukuk:
At some point in time the check for crypt_checksalt() did go lost, re-add it for meson.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -246,7 +246,7 @@ if not libcrypt.found()
   libcrypt = cc.find_library('crypt')
 endif
 crypt_h = cdata.get('HAVE_CRYPT_H') == 1 ? '''#include <crypt.h>''' : '''#include <unistd.h>'''
-foreach f: ['crypt_r']
+foreach f: ['crypt_r', 'crypt_checksalt']
   if cc.has_function(f, prefix: crypt_h,
                      args: '-D_GNU_SOURCE',
                      dependencies: libcrypt)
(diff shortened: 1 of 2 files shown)


modules/pam_unix/passverify.c
```diff
@@ -111,7 +111,7 @@ PAMH_ARG_DECL(int verify_pwd_hash,
 			 * Ok, we don't know the crypt algorithm, but maybe
 			 * libcrypt knows about it? We should try it.
 			 */
-#if defined(CRYPT_CHECKSALT_AVAILABLE) && CRYPT_CHECKSALT_AVAILABLE
+#ifdef HAVE_CRYPT_CHECKSALT
```
ldv-alt, line 114: Given that `CRYPT_CHECKSALT_AVAILABLE` is provided by libxcrypt's crypt.h, ...

meson.build
```diff
@@ -246,7 +246,7 @@ if not libcrypt.found()
   libcrypt = cc.find_library('crypt')
 endif
 crypt_h = cdata.get('HAVE_CRYPT_H') == 1 ? '''#include <crypt.h>''' : '''#include <unistd.h>'''
-foreach f: ['crypt_r']
+foreach f: ['crypt_r', 'crypt_checksalt']
```
ldv-alt, line 249: ... is this really needed?

thkukuk:
Ah, I missed that the define is coming from the header file.

I'm still waiting for feedback on that bug, but meanwhile it looks more like the problem is somewhere else.

thkukuk:
Ok, it's a NixOS problem

PR #1023, opened 2026-09-10
pam_umask: Reject invalid ulimit= values in GECOS field

Sovtouch:
Validate strtol()'s errno and endptr, reject negative values, and bound the value against the "* 512" multiplication before doing it, reporting an error the same way the neighboring "pri=" branch and the setrlimit() failure path already do.


ldv-alt (changes requested):
The patched code doesn't compile, didn't look into the change itself.

Sovtouch:
Sorry, I missed it, I corrected the code block, now it is build-up.

modules/pam_umask/pam_umask.c
```diff
@@ -174,15 +174,30 @@ setup_limits_from_gecos (pam_handle_t *pamh, options_t *options,
...
+	  errno = 0;
+	  limit = strtol (str, &endptr, 10);
+	  if (str == endptr || (*endptr != '\0' && *endptr != ',')
+	      || errno == ERANGE || limit < 0 || limit > LONG_MAX / 512)
 	    {
 	      if (!options->silent || options->debug)
-		pam_error (pamh, "setrlimit failed: %m\n");
-	      pam_syslog (pamh, LOG_ERR, "setrlimit failed: %m");
+		pam_error (pamh, "invalid ulimit value in GECOS field: %s\n", str);
+	      pam_syslog (pamh, LOG_ERR, "invalid ulimit value in GECOS field: %s", str);
```
ldv-alt: These new error messages log the entire GECOS remainder past `ulimit=` rather than just the invalid value token.  For example, GECOS `ulimit=abc,pri=5,umask=022` would log `abc,pri=5,umask=022`.

I suggest fixing this by restructuring the GECOS loop to compute the next comma delimiter once per iteration at the top of the loop body, then reusing it both for loop advancement and for bounding the error message via `%.*s`.  The error message format could also changed from "invalid ulimit value in GECOS field: %s" to "invalid value in GECOS field: %.*s", printing the full field (e.g. "ulimit=abc") so future validation of pri= and umask= can use the same format.

ldv-alt (changes requested):
The PR adds input validation for `ulimit=` in the GECOS field, but two adjacent GECOS parsers in the same function have the same class of bug that remains unfixed.

modules/pam_umask/pam_umask.c
```diff
@@ -152,38 +152,89 @@ setup_limits_from_gecos (pam_handle_t *pamh, options_t *options,
...
-		pam_error (pamh, "setrlimit failed: %m\n");
-	      pam_syslog (pamh, LOG_ERR, "setrlimit failed: %m");
-	    }
+      if (!options->silent || options->debug)
+        pam_error (pamh, "invalid value in GECOS field: %.*s\n",
+                   precision, cp);
+        pam_syslog (pamh, LOG_ERR,
+                    "invalid value in GECOS field: %.*s",
+                    precision, cp);
+        continue;
```
ldv-alt:
```
../modules/pam_umask/pam_umask.c: In function ‘setup_limits_from_gecos’:
../modules/pam_umask/pam_umask.c:197:7: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
  197 |       if (!options->silent || options->debug)
      |       ^~
../modules/pam_umask/pam_umask.c:200:9: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
  200 |         pam_syslog (pamh, LOG_ERR,
      |         ^~~~~~~~~~
```

ldv-alt (changes requested):

modules/pam_umask/pam_umask.c
ldv-alt: This has been detected by CI, please make sure your changes pass CI before submitting.

Sovtouch: Got it, I'm trying to fix it locally and running through some more adjustments.
Done.

ldv-alt (approved):
Please rebase.

Sovtouch:
Done.

ldv-alt:
Thanks!

